PR #30, opened 2016-05-09
fix lists formatting in readme.md

MrMitch:
The ordered lists in `readme.md` were not rendered properly (every item was detected as `1.` by the parser).

image

This is only a formatting issue to make the steps clearer, I have not changed any word or sentence.


Rokt33r:
Thanks for noticing me. I'll check readmd.md and update after merge this PR

PR #71, opened 2016-07-29
Fix delete last note

syossan27:
I was fixed "Can't delete last note in NoteList".


Rokt33r:
Thx! I'll ship it on next update!

Issue #36, opened 2016-05-20
Choose custom hex code when making a folder


MaximilianLloyd:
Having the ability to choose a custom hex code for the color of the folder would be very neat, and quite easy to implement.

Rokt33r:
I'll implement it on next update. please wait a moment. 👍

Rokt33r:
shipped!

PR #77, opened 2016-08-10
Feature: Folder color picker

dotcs:
This PR is based on the feature request in issue #36. I've implemented a color picker dialog instead of a list of predefined colors for the folders. I hope that we did not do work twice, but I just needed that feature for my notes. ;)

The PR makes use of the wonderful [react-color](https://github.com/casesandberg/react-color/) component which is licensed by the [MIT license](https://github.com/casesandberg/react-color/blob/master/LICENSE).

Actually calculating the position is not that easy because the dimensions of the content area of the modal is required. I calculate the dimensions after they are available and hand them as properties to the child components. You might want to have a closer look to the code to see if you're d'accord with this procedure or if this can be done better. Also see the images below.

Additional note:
Since the list of colors is just used for the menu that I've removed it might be a good idea to remove the color constants also. I cannot decide wether you need them for a different purpose, so I let them stay.

## Preview:

screen shot 2016-08-10 at 17 57 14

If there is not enough space for the picker, e.g. for the items at the bottom, it's re-positioned such that the user does not have to scroll.

screen shot 2016-08-10 at 17 57 07

### DIFF
--- a/browser/main/modals/PreferencesModal/StorageItem.js
+++ b/browser/main/modals/PreferencesModal/StorageItem.js
@@ -1,4 +1,5 @@
 import React, { PropTypes } from 'react'
+import ReactDOM from 'react-dom'
 import CSSModules from 'browser/lib/CSSModules'
 import styles from './StorageItem.styl'
 import consts from 'browser/lib/consts'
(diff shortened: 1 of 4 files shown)


Rokt33r:
Cool! I'll merge and apply the next update.

PR #96, opened 2016-08-24
submit the tag when receiving 'tab' key

yosmoc:
In existing implementation, 'enter' key is only allowed to submit the tag. However, if the user finishes to create the tag and make another tag, the user naturally send 'tab' key.
When receiving 'tab' key, the focus will go to the star icon and nothing is submitted, but the text is still remained. This makes confusing to the user.

This commit provide 'tab' key submitting the tag also.

### DIFF
--- a/browser/main/Detail/TagSelect.js
+++ b/browser/main/Detail/TagSelect.js
@@ -22,6 +22,10 @@ class TagSelect extends React.Component {
 
   handleNewTagInputKeyDown (e) {
     switch (e.keyCode) {
+      case 9:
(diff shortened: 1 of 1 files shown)


Rokt33r:
Great! I've recently heard about the tab index of editor. So, I think this kind of feedbacks are quite important. If you have more idea about this, it is also okay to make a new issue instead of making a PR.(It is your own choice)

yosmoc:
Thanks! Yeah, I will make the issue or PR to suggest my idea.

Issue #88, opened 2016-08-22
Finder issue with Multiple Screens


brpaz:
Hello.
I found an issue with finder when using 2 monitors.
It is centered in the middle of the screens making it unusable.

It appears like [this](http://i.imgur.com/2uiC4un.jpg)

I found this issue before with some other electron apps.

I think the correct behavior would be to open finder centered on the current active screen. 
We need to look at the methods of [Electron Screen API](https://github.com/electron/electron/blob/master/docs/api/screen.md)

Probably we need to combine some of these methods the best result. I might try to do some experiments with this if I have some free time.

Btw, This happens on Linux Mint 18. Not tested on other OS, but I guess it should have the same problem.

kazup01:
Is this bug still occurring?

PR #102, opened 2016-08-27
WIP:  Finder Issue with multiple monitors

brpaz:
Hello.
Regarding the issue #88 , i did some experiments and found out a possible solution.

Basically using the  "screen" module of electron, when open finder I can find the current screen based on mouse pointer position and then manually set x, y position of the finder window to be centered on that screen.

I put the code in finder-window.js but thats not the correct place, because it only called the first time.
Each time the user toggles finder I need to run that code to found out the correct position.

I treid the toggleFinder function in ipcServer.js but I put some console.log and that function doesnt seem to be called.
Where should I put this code to execute every time the finder is toggled?

### DIFF
--- a/lib/finder-window.js
+++ b/lib/finder-window.js
@@ -4,6 +4,7 @@ const BrowserWindow = electron.BrowserWindow
 const Menu = electron.Menu
 const MenuItem = electron.MenuItem
 const Tray = electron.Tray
+const Screen = electron.screen
 const path = require('path')
 
 var config = {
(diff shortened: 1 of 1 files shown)


Rokt33r:
Great!! Thank you so much. 

It must be done here. `browser/finder/ipcClient.js`
https://github.com/BoostIO/Boostnote/blob/master/browser/finder/ipcClient.js#L37-L39
It will be fired when finder shows. Also, you can access `finderWindow`.
I think using `finder.setPosition()` method would be enough.
http://electron.atom.io/docs/api/browser-window/#winsetpositionx-y-animate

Rokt33r:
@brpaz Please push it if there are any progress. I'm going to complete this on this week. 😀

sota1235:
@brpaz What is the state of this pull request?
If there is no progress, I will close it and issue this problem!

Thank you for contributing :D

brpaz:
@sota1235, I tried to put the code where @Rokt33r told but it was not working. I havent add time to work on it after.

asmsuechan:
@brpaz How about this PR?

asmsuechan:
I'll close this in under 3 days unless you react.

PR #129, opened 2016-09-24
Fix and improve docs

kg86:
Hi,

I tried to build and found following points to improve the build instruction document.
Could you check it?

- Added an instruction `npm install` which is required to build.
- Added `$` at the heading of each command in order to distinguish between commands and others.
- Fixed typo. "@" to "`"


Rokt33r:
Awesome work! Thank you so much!!

PR #143, opened 2016-10-11
remove oh-my-cdn and support node v0.10.21

osiyuk:
That's it.


Rokt33r:
As I told you we won't support any legacy version of NodeJS. Also, I'm going to focus on supporting only v6(stable and edge).
By the way, I agree with deleting  oh-my-cdn dependencies.

So, I'll cherry-pick the part of your PR instead of merging.

sota1235:
I remove oh-my-cdn partially instead of using npm.
I tried to cherry-pick your commits, but it conflicted. I'm sorry..

Thank you for sending pull request!

PR #157, opened 2016-10-20
Fix syntax error (unexpected token)

tejado:
Got this syntax error the first time compiling/running Boostnote.

### DIFF
--- a/lib/main-window.js
+++ b/lib/main-window.js
@@ -3,7 +3,7 @@ const app = electron.app
 const BrowserWindow = electron.BrowserWindow
 const path = require('path')
 
-var showMenu == process.platform == "Win32" ? false : true;
+var showMenu = process.platform == "Win32" ? false : true;
 
 var mainWindow = new BrowserWindow({
   width: 1080,


mikeres0:
Face palm

tejado:
@Rokt33r 
I'm now confused how this error was not fixed yet. Version 0.7.2 should already include this code, right? But how did you built this version then?
I'm asking because I saw at least one further bugs in the repo version which are not existing in the official built.

Rokt33r:
I'm not sure where should I check. Did you mean the library code of node_modules directory?

Rokt33r:
@tejado It is not included v0.7.2. You might confusing because Github shows all commits by its creation time, not by which branch it committed.

Check this. https://github.com/BoostIO/Boostnote/commits/v0.7.2

screen shot 2016-10-21 at 8 39 33 am

@mikeres0 Sorry for my late reviewing. 
Check this commit. https://github.com/BoostIO/Boostnote/commit/381b7d85f49ceaf0f8ef1071abe03162e6f011c1

- Conditional Operator should not be used when returning Boolean.
- No `==`. Only `obj == null` is allowed to check `null || undefined`.
> https://github.com/feross/standard

tejado:
Thanks a lot for the explanation :) I missed that option.

PR #159, opened 2016-10-20
Improved error messaging

mikeres0:
Instead of a standard `console.log` I've added an alert. Pressing the button and getting nowhere is not good from a user perspective.


Rokt33r:
I forgot to fix it. Thank you so much!

PR #160, opened 2016-10-21
Fix of storage unlink in preference popup

tejado:
The unlink storage in the preference popup isn't working due to an error of calling dispatcher method. With this fix, it should working again.


Rokt33r:
Thanks again! How about to join our slack? We could talk much actively!

Rokt33r:
@tejado Also, is it okay to add you to contributors section of `readme.md`?

tejado:
Yes to both, thanks :) regarding Slack: I already send you a mail. Did you get it?

PR #161, opened 2016-10-24
Updated readme english & reformatting

twhiting:
Updated the look of the readme and fixed some english errors.  Ran into a slight issue w/ how markdown renders html so I had to push a few more changes than expected.


Rokt33r:
It looks great! Thanks for your support.

PR #164, opened 2016-10-27
Minor accelerator updates

bkjohnson:
This updates the accelerator to use the OSX boolean instead of a new function where appropriate.  Also, for accelerators that are identical across platforms *except* for Command/Control, the 'CmdOrCtrl' modifier is used.

Edit:
The dark theme will also be applied to the input boxes and dropdowns in the Config modal.


Rokt33r:
Thanks for contributing! Is it okay to merge now?

bkjohnson:
Yes, go ahead! :)

PR #166, opened 2016-10-28
Adding menu back in temporarily to fix shortcuts.

bkjohnson:
Without the menu some people are unable to access certain functionality.  Until we are able to provide that functionality another way, this should fix the problem.


Rokt33r:
Thanks. @bkjohnson 

## To others

We will bring this back after the counter plan for shortcut without menu is ready.
[atom-keymap](https://github.com/atom/atom-keymap) and [electron-localshortcut](https://github.com/parro-it/electron-localshortcut) would be the valid options.

PR #167, opened 2016-10-28
Make the "Sucessfully applied!" message appear next to whichever button was clicked.

bkjohnson:
Previously, "Successfully applied!" would only appear next to the button for saving the hotkey configuration even if the Save UI Config button was clicked instead.  The code now detects which button the save came from and it will place the message next to the appropriate button.

If there is a better way of accomplishing this please let me know, I am new to both Electron and React.

### DIFF
--- a/browser/main/modals/PreferencesModal/ConfigTab.js
+++ b/browser/main/modals/PreferencesModal/ConfigTab.js
@@ -10,6 +10,9 @@ const ipc = electron.ipcRenderer
 
 const OSX = global.process.platform === 'darwin'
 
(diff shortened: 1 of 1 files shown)


sota1235:
I'm sorry to leaving for this pull request for a long time.
And thank you for request it!

It's so good idea, but the ConfigTab will devide into HotkeyTab and UI Tab in the next version, `v0.8.0`.
So, it is no longer needed.
Thank you for your cooperation!
I am looking forward to pull request again.

Thank you 😄

PR #214, opened 2017-01-07
Fixes the theme of code block by codemirror

asmsuechan:
Hello, I get an error on loading css in preview code block like this picture below.

```
file:///Applications/Boostnote.app/Contents/Resources/app/node_modules/codemirror/theme/default.css Failed to load resource: net::ERR_FILE_NOT_FOUND
```

It means default.css is not available. Thus I tried to fix it. Then I found the default theme `xcode` is no longer available too. Perhaps the theme `elegant` is the most similar to previous one (I attach Screen Shot below).

Please point out if I am wrong.

### DIFF
--- a/browser/components/MarkdownPreview.js
+++ b/browser/components/MarkdownPreview.js
@@ -184,7 +184,7 @@ export default class MarkdownPreview extends React.Component {
   }
 
   setCodeTheme (theme) {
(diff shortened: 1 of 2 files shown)


sota1235:
`elegant` theme is looks good!
Thank you for fixing :)

Issue #209, opened 2017-01-06
Characters are escaped at markdown preview


kasari:
![boostnote 2017-01-06 22-48-45](https://cloud.githubusercontent.com/assets/18321787/21719933/a2fd84f6-d463-11e6-8568-0b3879c4987c.png)
![boostnote 2017-01-06 22-49-03](https://cloud.githubusercontent.com/assets/18321787/21719938/a55824cc-d463-11e6-88d5-a0e71f030c4c.png)

kazup01:
Thank you for your message!
This bag is same as #205 .

Issue #205, opened 2017-01-04
<  > don't show at markdown preview.


kazup01:
![jan-04-0029 23-47-31](https://cloud.githubusercontent.com/assets/8602615/21646346/25638562-d2d9-11e6-8d10-db7539dccac7.gif)

< and > are change to  , &gt; and &lt; at the markdown preview.

asmsuechan:
This bug arises at  https://github.com/BoostIO/Boostnote/blob/master/browser/components/MarkdownPreview.js#L204

`markdown.render(value)` returns correct (like `<?php ?>`) value, however it is assigned to innerHTML as html comment `<!--?php ?-->`.

**Perhaps**, it occurs due to the settings of CodeMirror around auto completion.

igoradamenko:
I think there is an appropriate issue.

If I use JSX in markdown block of code, I get something strange:

![gif](https://cloud.githubusercontent.com/assets/6537798/22907001/297532e2-f258-11e6-9418-6648a60e960a.gif)

Note: app version is 0.8.5.

kazup01:
Hi @igoradamenko san, @asmsuechan fixed this bug.🎉
We will update next version! Thank you for waiting!

PR #216, opened 2017-01-08
Decodes HTML entity in code area

asmsuechan:
Hello, I find a bug in code area on writing HTML entity. They would not be decoded.

And I added encode, it works fine like below.

### DIFF
--- a/browser/components/MarkdownPreview.js
+++ b/browser/components/MarkdownPreview.js
@@ -219,7 +219,7 @@ export default class MarkdownPreview extends React.Component {
       let syntax = CodeMirror.findModeByName(el.className)
       if (syntax == null) syntax = CodeMirror.findModeByName('Plain Text')
       CodeMirror.requireMode(syntax.mode, () => {
-        let content = el.innerHTML
+        let content = decodeHTMLEntities(el.innerHTML)
         el.innerHTML = ''
         el.parentNode.className += ` cm-s-${codeBlockTheme} CodeMirror`
         CodeMirror.runMode(content, syntax.mime, el, {


asmsuechan:
This PR seems to be related with https://github.com/BoostIO/Boostnote/issues/205, however I could not fix it by this modification.

asmsuechan:
Reported by https://github.com/BoostIO/Boostnote/issues/209

sota1235:
I think this is not a bug.
For example, I write the code below. Entity reference should not be decoded.

```javascript
function escape(string) {
  return string
    .replace(/&/g, '&amp;')
    .replace(/</g, '&lt;')
    .replace(/>/g, '&gt;')
    .replace(/"/g, '&quot;')
    .replace(/'/g, '&#039;');
}
```

I think codeblock area is for seeing and reading the code. So it should not be modified by Application.

How about do yu think?

asmsuechan:
I've recognized the reason why we should use entities is for security. In this case, we don't need to care about entities because we use Boostnote in local only.

It's trouble that special characters are displayed as encoded style. I feel it's a big issue.

>  it should not be modified by Application.

Absolutely true.

I also reluctant to be displayed encoded entities in preview. How should I correct?

sota1235:
> I've recognized the reason why we should use entities is for security. In this case, we don't need to care about entities because we use Boostnote in local only and also it's OSS.

I agree.
So, I think all text in code block should not be escaped.
For example, `'&amp;'` must be displayed `'&amp;'` and `>` also must be displayed `>`.

How about you?

asmsuechan:
> For example, `'&amp;'` must be displayed `'&amp;'` and > also must be displayed >.

However in current version > is displayed as `&gt;` at preview code area. I think this is a bug.

According to inspection of [this line](https://github.com/BoostIO/Boostnote/blob/master/browser/components/MarkdownPreview.js#L204), perhaps `<code>` tag shows us `>` as `&gt;`.

refs: https://github.com/BoostIO/Boostnote/issues/205#issuecomment-271130035

sota1235:
I understood...
This problem is so difficult. We will consult how should we do with contributors this weekend.
Wait a moment please :bow:

asmsuechan:
Understood!

asmsuechan:
I strongly feel this code display problem is one of the most important problem at all. Thus we should fix the layout collapsion as soon as possible.

Something went wrong (lines are reduced) like below:
screen shot 2017-02-07 at 07 32 08

screen shot 2017-02-07 at 07 32 05

Issue #146, opened 2016-10-17
vim moving keymap in boostnote


asmsuechan:
Hello,
I'm vimmer also boostnoter. So I'd like to use vim moving keymap in boostnote. Could you develop it?
Thank you!

Rokt33r:
I'd like to do support Vim binging.

But, I am poor at using Vim. Could you help me a little? We have a slack. If you're okay, send a e-mail to me and I'll send an invitation. (fluke8259@gmail.com)

Btw, It doesn't seem to be hard by this. https://codemirror.net/doc/manual.html#vimapi

asmsuechan:
I send you an email just now!

PR #218, opened 2017-01-08
Adds vim keymap setting

asmsuechan:
Hello, finally I add vim-keymap to boostnote.

reference: https://codemirror.net/2/demo/vim.html

### DIFF
--- a/browser/components/CodeEditor.js
+++ b/browser/components/CodeEditor.js
@@ -52,7 +52,7 @@ export default class CodeEditor extends React.Component {
       indentUnit: this.props.indentSize,
       tabSize: this.props.indentSize,
       indentWithTabs: this.props.indentType !== 'space',
-      keyMap: 'sublime',
+      keyMap: this.props.keyMap,
       inputStyle: 'textarea',
       extraKeys: {
         Tab: function (cm) {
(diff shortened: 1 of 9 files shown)


kazup01:
Wow great!
I'm so excited.

asmsuechan:
However, we should reload boostnote after change the setting because the keymap itself isn't loaded only by just change `config.editor.keyMap`. Thus I put a note. Please check the sentence also.

sota1235:
This is so cool!! I'll check it.

sota1235:
@asmsuechan Thank you for your contribution. 
The code changed looks good, but I want you fixing more files.
`browser/finder/NoteDetail.js` and `browser/main/Detail/SnippetNoteDetail.js` also use `CodeEditor` component. So please add `keymap` props to them.

asmsuechan:
@sota1235 Thanks for your review, I try to fix them!

asmsuechan:
I added props to them however I cannot figure out why I should add keyMap to browser/finder/NoteDetail.js. I cannot write note on finder.

sota1235:
That's correct specification that you can't edit the note on finder.
So, that should not be focused when the note clicked. It's other problem.

I like vim, so this pull request is like God. Thank you!

sota1235:
Related #146

Issue #207, opened 2017-01-04
Remember window size


killgrey:
First of all, I would like to say that I really like this app. Although it's not yet polished but I can see it's kickin ass in the future.

I like the feature of remembering the window size when I quit the app. Because everytime I open it, I have to resize it again to the way I want it. 

Thank you!

kazup01:
Hi killgrey san,
Thank you for your message!
I want it too. We will add to the todo list.
How is the new design? Please feedback to us!

killgrey:
Yeah I've checked your todo list and it's quite awesome! I love the UI & UX better than the old one. It's simple and clean, and most of the missing features are already in your todo list. 

One thing I want to see is I can fullscreen the editor mode or write without distraction. That's all you guys are doin great!

sota1235:
asmsuechan implements it. So this will be released at next version.
Wait a moment please!

PR #221, opened 2017-01-08
refs #207 Keeps window size

asmsuechan:
Hello, I dealt with https://github.com/BoostIO/Boostnote/issues/207

### DIFF
--- a/lib/main-window.js
+++ b/lib/main-window.js
@@ -6,10 +6,7 @@ const Config = require('electron-config')
 const config = new Config()
 
 var showMenu = process.platform !== 'win32'
-let windowSize = { width: 1080, height: 720 }
-
-if(config.get('windowsize') === undefined) config.set('windowsize', windowSize)
-windowSize = config.get('windowsize')
+let windowSize = config.get('windowsize') || { width: 1080, height: 720 }
 
 let mainWindow = new BrowserWindow({
   width: windowSize.width,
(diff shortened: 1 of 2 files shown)


lib/main-window.js
```diff
@@ -2,12 +2,18 @@ const electron = require('electron')
...
+const Config = require('electron-config')
+const config = new Config()
 
 var showMenu = process.platform !== 'win32'
+let windowSize = { width: 1080, height: 720 }
 
-var mainWindow = new BrowserWindow({
-  width: 1080,
-  height: 720,
+if(config.get('windowsize') === undefined) config.set('windowsize', windowSize)
```
sota1235: Please add space after `if` statement. We compliant [JavaScript Standard Style](https://github.com/feross/standard)!

asmsuechan: Thanks for your advice! I modify soon.

lib/main-window.js
```diff
@@ -2,12 +2,18 @@ const electron = require('electron')
...
+windowSize = config.get('windowsize')
```
sota1235: You can write the code more short.

```
const windowSize = config.get('windowsize') || { width: 1080, height: 720 }
```

lib/main-window.js
```diff
@@ -2,12 +2,18 @@ const electron = require('electron')
...
+
+let mainWindow = new BrowserWindow({
```
sota1235: Thank you for using `let` instead of `var` :D

asmsuechan: 🙆

lib/main-window.js
asmsuechan: Wow absolutely true...

asmsuechan:
I fixed it :pray:

sota1235:
Thank you for fixing. Looks good to me!

PR #206, opened 2017-01-04
Replace old url.

sota1235:
Related https://github.com/BoostIO/Boostnote/issues/202

# Update

- Replace old url. (`http://b00st.io` -> `http://boostnote.io`)
- Remove unused file.

PR #223, opened 2017-01-09
Modifies invalid url

asmsuechan:
I modified unavailable urls https://boostnote.io to https://b00st.io


asmsuechan:
According to https://github.com/BoostIO/Boostnote/pull/206, I seem to make a mistake 😅

Issue #226, opened 2017-01-10
Make it possible to use Boostnote with the keyboard only


huettenhain:
Briefly put, it would be very nice to have more (configurable) Hotkeys for:

* Focussing the search field (possible default: `Ctrl`+`S`)
* Focussing the tag editor (possible default: `Ctrl`+`T`)
* Selecting the next and previous note (possible default: `Ctrl`+`Up` and `Ctrl`+`Down`)
* Switch between editor and preview (possible default: `Ctrl`+`E`)
switching to editor should move the cursor focus to the editor window.
* Deleting a note (possible default: `Del`)

This list contains the hotkeys that seem necessary to me to make Boostnote usable without the mouse, which I would consider a great boost in usability, no pun intended.

kazup01:
This is great idea. I think these hotkeys can great usability to Boostnote, too.
Thank you for proposal!
I think that it is better for the Hotkeys to be customizable.

asmsuechan:
Hi @huettenhain, do you know actually some shortcuts already exist (however they aren't customizable).

* Delete a note on pressed `d`
* Create a note on pressed `a`
* Edit a note on pressed `e`

I also think the shortcuts should be customizable.

huettenhain:
Hey @asmsuechan, thanks for that info. Unfortunately, these hotkeys do not seem to work application-wide, I have to select a note in the list with the mouse to make them work. Another issue is that once you are in edit mode, there is no way to leave without using the mouse.

asmsuechan:
I'm also bothered by the key map. Yes, I recognize those key maps work only on focused to note area. Perhaps we developers should discuss about the key map.

asmsuechan:
Hi @huettenhain, finally I implemented the shortcuts and be released.

* Focussing the search field (Ctrl+S)
* Selecting the next and previous note (Ctrl+U and Ctrl+J)
* Switch between editor and preview (Ctrl+E)
* Deleting a note (Ctrl+Del)

Next, I try to enable customizable.

asmsuechan:
To be customizable is the another issue, thus I close this issue. Please check https://github.com/BoostIO/Boostnote/issues/302 and comment you opinions if you're interested in customizable. Thank you!

eleijonmarck:
Could you please make it so that you can at least show the shortcuts available. I think that some shortcuts should be default.

Of course it should be able to customize all the shortcuts. That is awesome.

Nice-to-have:
[...]

PR #227, opened 2017-01-10
Multiple key mapping on Preview

asmsuechan:
#Hi,

to handle https://github.com/BoostIO/Boostnote/issues/226, we need to implement multiple key recognition first. And I'd like to discuss based on source code.

On this PR, when I want to delete a note by `Delete` key, I change `config.hotkey.createNote` from `d` to `Delete`.

By the way, I switched `e.keyCode` to` e.key` because `e.keyCode` is a deprecated API. refs: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent

Finally, I'm not sure I wrote a good code thus you can ask anything.

### DIFF
--- a/browser/main/NoteList/index.js
+++ b/browser/main/NoteList/index.js
@@ -39,10 +39,10 @@ class NoteList extends React.Component {
       this.refs.list.focus()
     }
 
+    this.hotkey = props.config.hotkey
+
     this.state = {
     }
-
-    this.shortcutKey = props.config.preview.shortcutKey
   }
 
   componentDidMount () {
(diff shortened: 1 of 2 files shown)


asmsuechan:
This PR is for handling like `Ctrl + E` shortcut. Now we can use only 1 key in shortcut.

asmsuechan:
I'd like to change the default keymap like below:

* delete a note: `d` to `del`

Probably it's intuitive than current keymap. And I'd like to add a keymap which can escape from Edit mode to Preview mode. How do you think it?

asmsuechan:
I close this PR thus I implemented shortcuts to substitute in https://github.com/BoostIO/Boostnote/pull/235.

Issue #239, opened 2017-01-14
after v0.8.1 / Cannot open finder with using HotKey


sota1235:
Cannot open finder with using HotKey.
This bug maybe created when refactoring with ESLint.

Related https://github.com/BoostIO/Boostnote/pull/231#issuecomment-272609449

sota1235:
The bug was fixed for macOS. But the bug on windows is not fixed yet.
I will fix it.

PR #240, opened 2017-01-14
refs #239 Fixes a bug which cannot open finder with using HotKey

asmsuechan:
Hi, I tried to fix however nothing thinking to fix it.


sota1235:
Thank you!!!! I'll check it!!!

sota1235:
I confirm bug is fixed. I'll release this code tonight. Thank you!!

sota1235:
Related #239

asmsuechan:
Oops, should we add lib/finder-app.js to .eslintignore?

sota1235:
I fixed by ignoring the line.
https://github.com/BoostIO/Boostnote/commit/ad4e50d54241637c288ba8eac6cc0fdf2349514b

So, if you'll get same problem, please use ignore the code with inline comments!
http://eslint.org/docs/user-guide/configuring#disabling-rules-with-inline-comments

asmsuechan:
Wow! I didn't know. Thanks for this information!

Issue #208, opened 2017-01-05
Not working inline code syntax hilight.


syossan27:
![2017-01-05 15 48 05](https://cloud.githubusercontent.com/assets/910446/21671373/69d34220-d35e-11e6-8cb5-39c7e91ae069.png)

![2017-01-05 15 48 20](https://cloud.githubusercontent.com/assets/910446/21671377/6c9eb17e-d35e-11e6-8330-3acc66d58a36.png)

Is the syntax highlight function of inline code deleted?

sota1235:
Thank you for reporting.
It is probably the bug. I will check it!

sota1235:
It's released on v0.8.3. Thank you for reporting!

TMorville:
I have this exact issue in `0.11.8` on OSX and `0.11.9` as well. Inline code is not formatted as code.

PR #241, opened 2017-01-14
Fixes syntax hilight of inline code on Preview

asmsuechan:
Hi, I fixed the bug reported by https://github.com/BoostIO/Boostnote/issues/208.

I referenced GitHub for this style.


asmsuechan:
Hmm

sota1235:
That's more easy to read! But I think it can be more easy to read.
For example, color of the text is `CC305F`, it's like below.





I think changing color of the text is easy and best way.
How about you?

asmsuechan:
It looks so good! I'll change the code.

asmsuechan:
How about them?

sota1235:
That's so cool.
Looks good to me!

Issue #243, opened 2017-01-15
A bug occurs on update


asmsuechan:
Hi, I hit a bug on update...

<img width="532" alt="2017-01-16 1 09 57" src="https://cloud.githubusercontent.com/assets/11307908/21964088/fbe77474-db88-11e6-9909-df07d9c618a1.png">

Does anyone arise it?

After I had killed Boostnote process and restarted, it was updated properly (v0.8.2).

kazup01:
We are investigating the cause of this bug.
Downloading the latest version from the official page(http://boostnote.io/#download) will work fine, so I recommend it.
Of course the datas are not goes out, it is possible to take over without problems.

sota1235:
Thank you for reporting.
What is the version you used? Please tell me :bow:

asmsuechan:
It occurs on update from v0.8.1 to v0.8.2.

sota1235:
I see. I try to investigate...!

asmsuechan:
I blamed the line of the bug, I recognize it is my fault :fearful:
https://github.com/BoostIO/Boostnote/blob/master/lib/main-window.js#L52

Perhaps `if (mainWindow != null)` is necessary.

asmsuechan:
electron-gh-releases is used on update (https://github.com/BoostIO/Boostnote/blob/master/lib/main-app.js#L79)

And `install()` method uses `quitAndInstall()`, according to the reference of `quitAndInstall()` ([here](https://github.com/electron/electron/blob/master/docs/api/auto-updater.md#autoupdaterquitandinstall)),
> Note: autoUpdater.quitAndInstall() will close all application windows first and only emit before-quit event on app after that. This is different from the normal quit event sequence.

The bug occurs at `before-quit` event. Thus it's related with the bug.

asmsuechan:
Thus I tried to create `quit` event like below however same error still occur.
```js
  app.on('quit', function (e) {
    config.set('windowsize', mainWindow.getBounds())
  })
```

sota1235:
Great fixing by @asmsuechan !!
It will be released on v0.8.3

kazup01:
Thank you for your help, @asmsuechan !👏

sota1235:
It's fixed on v0.8.3. Thank you for reporting!

PR #244, opened 2017-01-15
Fix a bug due to event handler of before-quit on update

asmsuechan:
For now, I'd like to suggest ignoring any errors at `config.set('windowsize', mainWindow.getBounds())`. I checked this code behaviors as expected.
https://github.com/BoostIO/Boostnote/issues/243


sota1235:
Thank you for your investigation :D

PR #245, opened 2017-01-16
Add export as txt/md

asmsuechan:
Hi, I implemented a function which exports as txt or md. I heard it was said that someone wants this function.

### DIFF
--- a/browser/components/MarkdownPreview.js
+++ b/browser/components/MarkdownPreview.js
@@ -6,7 +6,7 @@ import consts from 'browser/lib/consts'
 import Raphael from 'raphael'
 import flowchart from 'flowchart'
 import SequenceDiagram from 'js-sequence-diagrams'
-import ee from 'browser/main/lib/eventEmitter'
+import eventEmitter from 'browser/main/lib/eventEmitter'
 import fs from 'fs'
 
 function decodeHTMLEntities (text) {
@@ -27,7 +27,7 @@ function decodeHTMLEntities (text) {
 const { remote } = require('electron')
 const { app } = remote
 const path = require('path')
-const dialog = remote.require('electron').dialog
+const dialog = remote.dialog
 
 const markdownStyle = require('!!css!stylus?sourceMap!./markdown.styl')[0][1]
 const appPath = 'file://' + (process.env.NODE_ENV === 'production'
@@ -157,7 +157,9 @@ export default class MarkdownPreview extends React.Component {
     dialog.showSaveDialog(remote.getCurrentWindow(), options,
     (filename) => {
       if (filename) {
-        fs.writeFile(filename, this.props.value, () => {})
+        fs.writeFile(filename, this.props.value, (err) => {
+          if (err) throw err
+        })
       }
     })
   }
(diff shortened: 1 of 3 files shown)


kazup01:
That’s awesome! We'll check it. Cheers👏

browser/components/MarkdownPreview.js
```diff
@@ -6,6 +6,8 @@ import consts from 'browser/lib/consts'
 import Raphael from 'raphael'
 import flowchart from 'flowchart'
 import SequenceDiagram from 'js-sequence-diagrams'
+import ee from 'browser/main/lib/eventEmitter'
```
sota1235: Please do not omit the name of variable. For example, use `eventEmitter`.

browser/components/MarkdownPreview.js
```diff
@@ -25,6 +27,7 @@ function decodeHTMLEntities (text) {
 const { remote } = require('electron')
 const { app } = remote
 const path = require('path')
+const dialog = remote.require('electron').dialog
```
sota1235: `electron` seems to be required twice.
If you import it on the first line, you don't need require twice.

```javascript
import { remote, dialog } from 'electron'
```

browser/components/MarkdownPreview.js
```diff
@@ -134,6 +139,29 @@ export default class MarkdownPreview extends React.Component {
...
+    const options = {
+      filters: [
+        { name: 'Documents', extensions: [fileType]}
+      ],
+      properties: ['openFile', 'createDirectory']
+    }
+    dialog.showSaveDialog(remote.getCurrentWindow(), options,
+    (filename) => {
+      if (filename) {
+        fs.writeFile(filename, this.props.value, () => {})
```
sota1235: Please handle error by using callback function:bow:

lib/main-menu.js
```diff
@@ -63,6 +63,26 @@ var file = {
       type: 'separator'
     },
     {
+      label: 'Export as',
+      submenu: [
+        {
+          label: 'Plain Text (.txt)',
+          click: function () {
```
sota1235: You can write like below. It is more modern way.

```javascript
click () {
}
```

sota1235:
Thank you for your contribution! Please fix some points 🐱

asmsuechan:
@sota1235 Thanks so much for your review :pray: :pray: :pray:, I'll fix them :pray: :pray: :pray:

browser/components/MarkdownPreview.js
asmsuechan: OMG, I don't have to require twice :fearful:. However it cannot operate correctly because I want to use `dialog` in the renderer process. Thus I change it to `const dialog = remote.dialog`
> The Dialog is opened from Electron's main thread. If you want to use the dialog object from a renderer process, remember to access it using the remote:
> ```
>const {dialog} = require('electron').remote
>console.log(dialog)
>```
refs: https://github.com/electron/electron/blob/master/docs/api/dialog.md

asmsuechan:
I fixed them :pray:

sota1235:
Thank you for fixing!
I want you to fix last one point.
If I try to export file when `Snippet` note displayed, it won't happen nothing.
I think that it would be more helpful to have a guide.
For example, show alert message `Markdown note only support`!

asmsuechan:
Thanks for you review so much :pray: I'll handle it!

asmsuechan:
> For example, show alert message Markdown note only support!

Hi, I handled it.

browser/main/NoteList/index.js
```diff
@@ -305,6 +310,20 @@ class NoteList extends React.Component {
     })
   }
 
+  alertIfSnippet() {
+    let { location } = this.props
+    let targetIndex = _.findIndex(this.notes, (note) => {
```
sota1235: Please use `const` instead of `let`!

browser/main/NoteList/index.js
```diff
@@ -305,6 +310,20 @@ class NoteList extends React.Component {
...
+      return note.storage + '-' + note.key === location.query.key
```
sota1235: You can use template literal!

```js
`${note.storage}-${note.key}`
```

sota1235:
Tha's all from me. Please fix, then I'll merge :bow:

asmsuechan:
@sota1235 Thanks for your review! I fixed the code you pointed :pray:

sota1235:
Thank you for fixing. Looks good to me!

Issue #246, opened 2017-01-16
Bug with Text Selection in Finder Window


winkelsdorf:
Hi,

there is a Bug when using the Finder Window with markdown Notes.

**Steps to reproduce**
- Open Finder Window
- Enable Display of markdown Notes
- Select a markdown Note (e.g. `Welcome to Boostnote :)`)
- Try to selected text from the markdown

**Expected**
Text being selected and be able to copy it

**Actual Result**
Finder jumps back to "All Notes"

**Other Information**
Version tested: 0.8.2 (0.8.2)
Environment: macOS Sierra 10.12.2

hth!

Frederik

kazup01:
Hi @winkelsdorf san,
Thank you for your information.
hmm, my Boostnote is not occur this bug.
version 0.8.2
macOSS Sierra 10.12.2

Does anyone occurred same bug?

kazup01:
We'll check it.

winkelsdorf:
Hi @kazup01, 

You are welcome and thank you for checking it!

I made a quick screencast to show the issue, notice the jump from the note to "All Notes":
![boostnote finder selection](https://cloud.githubusercontent.com/assets/1413291/21981269/e4d7331a-dbe6-11e6-95c8-61bd5f074121.gif)

kazup01:
Oh, I see! I was mistaken, and my Boostnote also occurs the same bug.
Since the code of  Finder Window is complicated, so we need to refactoring it, we'll check it together.
Thank you for your reporting!

winkelsdorf:
Good to hear it's reproducible. This has no hurry, I am usually using the default window. At least it's now a documented bug for others stumbling upon this.

Thank you for this useful application and your work!

kazup01:
Thank you for your compassion and enjoying Boostnote!
We'll do our best to better the Boostnote😤❗

winkelsdorf:
You're very welcome! It's already a very helpful tool 👍

kazup01:
Hi @winkelsdorf san, @asmsuechan san has fixed this bug!(It is very early, so I am surprised.
) We will probably improve to this bug the next update!

winkelsdorf:
Hi @kazup01, thank you - that was very fast! Thanks to @asmsuechan for his help, too!

kazup01:
I close this issue. Thank you @winkelsdorf san👏

PR #247, opened 2017-01-16
Fix a Bug when using the Finder Window with markdown Notes

asmsuechan:
Hi, this PR is for handling https://github.com/BoostIO/Boostnote/issues/246. It occurs on blur. Thus I removed blur handler.

### DIFF
--- a/browser/finder/index.js
+++ b/browser/finder/index.js
@@ -48,19 +48,28 @@ class FinderMain extends React.Component {
     }
 
     this.focusHandler = (e) => this.handleWindowFocus(e)
+    this.blurHandler = (e) => this.handleWindowBlur(e)
   }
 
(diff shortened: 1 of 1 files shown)


kazup01:
Wow, Very early!
You are superman! Thank you for your everday supports!!!! We'll check it!

asmsuechan:
Umm, this PR fixes the bug but I feel it seems to decrease usability for now...

asmsuechan:
I recognized an issue that click event becomes invalid when the `MarkdownPreview` area is focused in finder. Perhaps, the cause is that `iframe` in `MarkdownPreview` grabs the events.

sota1235:
Thank you for fixing. LGTM!

Issue #31, opened 2016-05-10
about Zooming UI


Rokt33r:
Putting Zoom select on status bar of GitKraken is quite brilliant design. It can be found easily while not annoying.
I'll try to imitate this feature.
![screen shot 2016-05-11 at 8 01 22](https://cloud.githubusercontent.com/assets/5865853/15165156/9d0d9cf2-174e-11e6-9952-c2daccf4f4ad.png)

Rokt33r:
This feature will be shipped. Check #3

whizark:
This is a really nice feature that many other Electron apps don't have :smiley::+1:
It would be nice if I could select more zoom % because the max zoom % was a bit small on large screen (3840 x 2160 etc.).

e.g. 130% (current max zoom, a bit small), 150% (readable), 175% (comfortable), 200% (super readable)

PR #248, opened 2017-01-17
modify: add more zoom options (140-200%)

whizark:
The current max zoom % (130%) is a bit small on large screen (3840 x 2160, hi-resolution but not so large display etc.).
Probably, it is also related to the default text color (`#939395`) or the font-weight (`normal`) but more zoom % makes text more readable enough.

e.g. 130% (current max zoom, a bit small), 150% (readable), 200% (super readable, a bit large)

See also: https://github.com/BoostIO/Boostnote/issues/31#issuecomment-267843725

### DIFF
--- a/browser/main/StatusBar/index.js
+++ b/browser/main/StatusBar/index.js
@@ -8,7 +8,7 @@ const electron = require('electron')
 const { remote, ipcRenderer } = electron
 const { Menu, MenuItem, dialog } = remote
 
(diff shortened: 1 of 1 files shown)


kazup01:
Thank you for your support! We'll check it!

sota1235:
Thank you for your contribution. LGTM!

whizark:
Thank you for merging the PR! This gives a big boost :rocket: to my productivity! :star2:

Issue #260, opened 2017-01-23
Ctrl + E conflict problem


asmsuechan:
Hi I add `Ctrl + E` shortcut as escape from CodeEditor. However it conflicts a shortcut which moves the cursor to the end of the line on mac.

refs: https://en.wikipedia.org/wiki/Table_of_keyboard_shortcuts

asmsuechan:
I'll try to change the default shortcut `Ctrl + E`. Which key is useful? (ESC is already used in `vim` mode)

asmsuechan:
Also, `Ctrl + U` doesn't work at least `vim` keymap in CodeEdit mode.

asmsuechan:
I think `Ctrl + Q` is the one of choices.

PR #261, opened 2017-01-23
refs #260 Change a shortcut Ctrl-E to Ctrl-W

asmsuechan:
Hi, I just changed the default shortcut keys which escapes from CodeEditor to MarkdownPreview.

refs: https://github.com/BoostIO/Boostnote/issues/260


sota1235:
Thank you for fixing! Good work!

PR #262, opened 2017-01-24
Changed the order of English and Japanese.

kazup01:


contributing.md
```diff
@@ -1,23 +1,21 @@
-# Contributing to Boostnote
+# Contributing to Boostnote(English)
```
sota1235: I think `(English)` is not necessary, because it is obvious that it is English.

sota1235:
LGTM!

PR #267, opened 2017-01-26
Fix Ctrl + E shortcut on escape from CodeEditor

asmsuechan:
Hi, I fixed a reported bug which is escaped on pressed any key in CodeEditor when you focus for editing the note by click.

related: https://github.com/BoostIO/Boostnote/issues/268


sota1235:
Looks good to me. Thanks a lot!

PR #278, opened 2017-02-09
refs #239 [Fix] cannot open finder with using hotkey on windows

EmEpsilon:
I fix the bug #239 on windows


kazup01:
Wow, great👏 Thank you for your cooperation! We will check it!

kohei-takata:
LGTM
Thanks @EmEpsilon :)

PR #301, opened 2017-02-21
Add a shortcut for supporting to input bold

asmsuechan:
I added a shortcut for inputting `****` and it brings the caret to the center of `****`.

b397b162808bb03d54c99e5792c782fd


asmsuechan:
I also added to enable `Ctrl+:` in a word to make it bold. And I'd like to ask the shortcut key. Actually `Ctrl+B` is duplicated.

a1a874d8f4a8f6205b8ec0fb8ef805c8

PR #317, opened 2017-03-10
Add prefer-const rule and fix code

sota1235:


asmsuechan:
Hi @sota1235, it does not work at least in my environment. The way how to test:

1. Add line `let a = 1`
2. Run `npm run lint`

However, nothing happens.

sota1235:
@asmsuechan I fixed npm script bug. Then more then 500 errors occurs :sob:
So firstly, add `prefer-const` rule as `warning` and fix little by little.
Plz review again :bow:

sota1235:
Thanks a lot for reviewing.

Issue #302, opened 2017-02-22
~/.boostnoterc


asmsuechan:
Hi, I want `~/.boostnoterc` for customizable. First, do you need this? If needed, we have to consider about the styles and what settings to be customizable.

asmsuechan:
## Settings
* whole settings in MENU
* custom theme
* shortcuts
* any a bit programs (by javascript?)

asmsuechan:
The style of ~/.boostnoterc I suggest:
```sh
set editor: {
  theme: 'molokai'
}

exec ~/boostnote.js
before_save before_save.js
after_save after_save.js
plugin asmsuchan/boostnote-plugin-github-integration
```

eleijonmarck:
Please could you enable to see what shortcuts are available for within the editor as well

asmsuechan:
Hi, @huettenhain. Alright, I try to write them down.

|Keys|Behaviour|
|:--|--:|
|Ctrl + N|Add new note|
|Ctrl + E|Switch to Editor mode|
|Ctrl + W|Switch to Preview mode|
|Ctrl + J|Move to the note below|
|Ctrl + U|Move to the note above|
|Ctrl + S|Focus on the search input|
|Ctrl + Del|Delete the note|
|Ctrl + B|Add bold markdown(****) (only Editor mode) (DELETION SCHEDULE)|
|Ctrl + :|Change the word to bold|
|Ctrl(Cmd) + Alt + L|Toggle main (customizable in `MENU`)|
|Ctrl(Cmd) + Alt + S|Toggle Finder(popup)  (customizable in `MENU`)|

If you don't understand what I say, you can point it out :pray:

Flexo013:
This is a nice suggestion. Is there still a lot of interest in a feature like this?

PR #318, opened 2017-03-10
Enable to set configs in ~/.boostnoterc

asmsuechan:
Hi, I added a function to enable to set configs in `~/.boostnoterc`. The priority of reading order is `LocalStorage` first, `~/.boostnoterc` second. And the reason why I created `RcParser` is because I'd like to add a function which enables to use `plugin`.

### refs
https://github.com/BoostIO/Boostnote/issues/302

### DIFF
--- a/browser/main/lib/ConfigManager.js
+++ b/browser/main/lib/ConfigManager.js
@@ -67,15 +67,7 @@ function get () {
     config = Object.assign({}, DEFAULT_CONFIG, JSON.parse(config))
 
     config = Object.assign({}, DEFAULT_CONFIG, boostnotercConfig)
-    config.hotkey = Object.assign({}, DEFAULT_CONFIG.hotkey, boostnotercConfig.hotkey)
-    config.ui = Object.assign({}, DEFAULT_CONFIG.ui, boostnotercConfig.ui)
-    config.editor = Object.assign({}, DEFAULT_CONFIG.editor, boostnotercConfig.editor)
-    config.preview = Object.assign({}, DEFAULT_CONFIG.preview, boostnotercConfig.preview)
-
-    config.hotkey = Object.assign({}, DEFAULT_CONFIG.hotkey, config.hotkey)
-    config.ui = Object.assign({}, DEFAULT_CONFIG.ui, config.ui)
-    config.editor = Object.assign({}, DEFAULT_CONFIG.editor, config.editor)
-    config.preview = Object.assign({}, DEFAULT_CONFIG.preview, config.preview)
+    config = assignConfigValues(config, boostnotercConfig, config)
 
     if (!validate(config)) throw new Error('INVALID CONFIG')
   } catch (err) {
@@ -136,6 +128,14 @@ function set (updates) {
   ipcRenderer.send('config-renew', {
     config: get()
   })
+}
+
+function assignConfigValues (config, rcConfig, originalConfig) {
+  config.hotkey = Object.assign({}, DEFAULT_CONFIG.hotkey, rcConfig, originalConfig.hotkey)
+  config.ui = Object.assign({}, DEFAULT_CONFIG.ui, rcConfig, originalConfig.ui)
+  config.editor = Object.assign({}, DEFAULT_CONFIG.editor, rcConfig, originalConfig.editor)
+  config.preview = Object.assign({}, DEFAULT_CONFIG.preview, rcConfig, originalConfig.preview)
+  return config
 }
 
 export default {
--- a/browser/main/lib/ConfigManager.js
+++ b/browser/main/lib/ConfigManager.js
@@ -68,13 +68,11 @@ function get () {
 
     config = Object.assign({}, DEFAULT_CONFIG, JSON.parse(config))
 
-    if (boostnotercConfig !== undefined) {
-      config = Object.assign({}, DEFAULT_CONFIG, boostnotercConfig)
-      config.hotkey = Object.assign({}, DEFAULT_CONFIG.hotkey, boostnotercConfig.hotkey)
-      config.ui = Object.assign({}, DEFAULT_CONFIG.ui, boostnotercConfig.ui)
-      config.editor = Object.assign({}, DEFAULT_CONFIG.editor, boostnotercConfig.editor)
-      config.preview = Object.assign({}, DEFAULT_CONFIG.preview, boostnotercConfig.preview)
-    }
+    config = Object.assign({}, DEFAULT_CONFIG, boostnotercConfig)
+    config.hotkey = Object.assign({}, DEFAULT_CONFIG.hotkey, boostnotercConfig.hotkey)
+    config.ui = Object.assign({}, DEFAULT_CONFIG.ui, boostnotercConfig.ui)
+    config.editor = Object.assign({}, DEFAULT_CONFIG.editor, boostnotercConfig.editor)
+    config.preview = Object.assign({}, DEFAULT_CONFIG.preview, boostnotercConfig.preview)
 
     config.hotkey = Object.assign({}, DEFAULT_CONFIG.hotkey, config.hotkey)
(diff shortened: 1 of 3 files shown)


asmsuechan:
An example of `~/.boostnoterc`

```
{
  "editor": {
    "keyMap": "vim",
    "theme": "railscasts"
  },
  "execs": [
    "CodeMirror.Vim.map('jj', '<Esc>', 'insert')"
  ]
}

```

browser/main/lib/RcParser.js
```diff
@@ -0,0 +1,23 @@
+import _ from 'lodash'
+
+const path = require('path')
+const sander = require('sander')
+
+function parse (boostnotercPath) {
```
sota1235: Is it necessary that receiving `boostnotercPath` from outside?
I think the path of `.boostnoterc` is not multiple. So declare the path on this file is better.

browser/main/lib/RcParser.js
```diff
@@ -0,0 +1,23 @@
...
+  if (!sander.existsSync(boostnotercPath)) return
+  let config = JSON.parse(sander.readFileSync(boostnotercPath).toString())
+  return config
+}
+
+function exec (boostnotercPath) {
+  const config = this.parse(boostnotercPath)
+  if (config.execs === undefined) return
+  _.forEach(config.execs, (exec) => {
+    eval(exec)
```
sota1235: Why do you use `eval()`? `.boostnoterc` will be like below?

```
{
  "execs": [
    "alert('hoge');"
  ]
}
```

sota1235: If you want to use plugin function, decide interface of the plugin and require it is better.
For example, ESLint plugin is so.

http://eslint.org/docs/developer-guide/working-with-plugins#processors-in-plugins

Please consider about it :)

browser/main/lib/RcParser.js
```diff
@@ -0,0 +1,23 @@
+import _ from 'lodash'
+
+const path = require('path')
+const sander = require('sander')
```
sota1235: Why don't you use `import`? Please unify the coding style, `require` or `import`

browser/main/lib/RcParser.js
```diff
@@ -0,0 +1,23 @@
...
+
+function parse (boostnotercPath) {
+  if (!sander.existsSync(boostnotercPath)) return
```
sota1235: Maybe `return {};` is better because return value is used for `Object.prototype.assign`.

browser/main/lib/RcParser.js
```diff
@@ -0,0 +1,23 @@
...
+  let config = JSON.parse(sander.readFileSync(boostnotercPath).toString())
```
sota1235: You don't need to declare `config`. Return right now!

browser/main/lib/RcParser.js
sota1235: This line failed when return void.
https://github.com/BoostIO/Boostnote/pull/318/files#diff-4bdfd7ba1c9cd532de6563d2e3cb4934R14

browser/main/lib/RcParser.js
asmsuechan: I thought they're possible:

```
$HOME/.${appname}rc
$HOME/.${appname}/config
$HOME/.config/${appname}
$HOME/.config/${appname}/config
/etc/${appname}rc
/etc/${appname}/config
```
refs: https://github.com/dominictarr/rc#standards

But if so, the name `boostnotercPath` is something wrong.

browser/main/lib/RcParser.js
asmsuechan: Oh.... I made mistakes....

browser/main/lib/RcParser.js
asmsuechan: Alright :pray:

browser/main/lib/RcParser.js
asmsuechan: Oh... I made a mistake....

browser/main/lib/RcParser.js
asmsuechan: It's not `plugin`. Just I'd like to run some codes on starting Boostnote by `execs`.

browser/main/lib/ConfigManager.js
```diff
@@ -58,11 +62,25 @@ function get () {
...
+
     config = Object.assign({}, DEFAULT_CONFIG, JSON.parse(config))
+
+    if (boostnotercConfig !== undefined) {
+      config = Object.assign({}, DEFAULT_CONFIG, boostnotercConfig)
+      config.hotkey = Object.assign({}, DEFAULT_CONFIG.hotkey, boostnotercConfig.hotkey)
+      config.ui = Object.assign({}, DEFAULT_CONFIG.ui, boostnotercConfig.ui)
+      config.editor = Object.assign({}, DEFAULT_CONFIG.editor, boostnotercConfig.editor)
+      config.preview = Object.assign({}, DEFAULT_CONFIG.preview, boostnotercConfig.preview)
+    }
```
sota1235: If `RcParser.parse()` returns `{}`, you don't need to write this `if` statement.

browser/main/lib/RcParser.js
sota1235: mm...Do you want to use multiple config files? I have no idea for its situation.

asmsuechan: Actually no. For now, only to use `~/.boostnoterc` is the best I think.

asmsuechan:
Finally, I deleted `RcParser.exec` for now. Currently, assumed `.boostnoterc` is:

```
{
  "editor": {
    "keyMap": "vim",
    "theme": "railscasts"
  }
}
```

It just override settings int `Menu`.

browser/main/lib/ConfigManager.js
```diff
@@ -58,11 +62,21 @@ function get () {
...
+
     config = Object.assign({}, DEFAULT_CONFIG, JSON.parse(config))
+
+    config = Object.assign({}, DEFAULT_CONFIG, boostnotercConfig)
+    config.hotkey = Object.assign({}, DEFAULT_CONFIG.hotkey, boostnotercConfig.hotkey)
+    config.ui = Object.assign({}, DEFAULT_CONFIG.ui, boostnotercConfig.ui)
+    config.editor = Object.assign({}, DEFAULT_CONFIG.editor, boostnotercConfig.editor)
+    config.preview = Object.assign({}, DEFAULT_CONFIG.preview, boostnotercConfig.preview)
+
     config.hotkey = Object.assign({}, DEFAULT_CONFIG.hotkey, config.hotkey)
```
sota1235: Can't you write like below?

```javascript
config.hotkey = Object.assign({}, DEFAULT_CONFIG.hotkey, boostnotercConfig.hotkey, config.hotkey);
```

And I think this code doesn't implement with DRY principle. So maybe, create function to bind config and using it best!(You don't need to do that at this PR. Let's challenge next time :) )

browser/main/lib/RcParser.js
```diff
@@ -0,0 +1,15 @@
...
+  const homePath = global.process.env.HOME || global.process.env.USERPROFILE
+  const boostnotercPath = path.join(homePath, BOOSTNOTERC)
+
+  if (!sander.existsSync(boostnotercPath)) return {}
+  return JSON.parse(sander.readFileSync(boostnotercPath).toString())
+}
+
+export default {
+  parse
+}
```
sota1235: Good module!
If you have a time, I want unit test. (If you don't, you don't need to make it.)

asmsuechan: This module is hard to test. Because `~/.boostnoterc` is fixed value.

browser/main/lib/ConfigManager.js
asmsuechan: Sure

asmsuechan:
Close once

Issue #285, opened 2017-02-11
Change icon color in macOS menu bar


igoradamenko:
Hey guys!

Thanks for Boostnote, it's great. Is it possible (or is it in a roadmap) to change color of app icon in macOS menu bar? It's hellishly bright. Check this out:

<img width="479" alt="screen shot 2017-02-11 at 15 42 19" src="https://cloud.githubusercontent.com/assets/6537798/22853819/b78ae9da-f070-11e6-9517-129c51de67fb.png">

It would be great to get a monochrome icon!

asmsuechan:
Hi @igoradamenko, Thanks for your feedback! We'll try to confirm the icon.

PR #321, opened 2017-03-14
Change the tray icons.

kazup01:
Requested at #285.


brpaz:
I think it would be nice to have an icon for dark themes also. The new icon wont look good on my black menubar. Maybe make possible to select if we want a light or dark icon in the settings?

peterpetre:
Is it possible the icon in menu bar to be monochrome (black/white)? I think its already suggested but  it became brown :)

boostnote

PR #331, opened 2017-03-18
Fix the behavior of a feature which locks the editor

asmsuechan:
A feature what lock the editor behaves incorrectly.

f69baab1241f83c80ff89ea3412813f0

Sometimes, even the lock is opened, the editor keeps its mode when I move another item. Thus I fixed it.

### DIFF
--- a/browser/main/Detail/MarkdownNoteDetail.js
+++ b/browser/main/Detail/MarkdownNoteDetail.js
@@ -281,7 +281,7 @@ class MarkdownNoteDetail extends React.Component {
                   onFocus={(e) => this.handleFocus(e)}
                   onMouseDown={(e) => this.handleLockButtonMouseDown(e)}
                 >
-                  <i className={faClassName}/>
+                  <i className={faClassName} styleName='lock-button'/>
                   <span styleName='control-lockButton-tooltip'>
                   {this.state.isLocked ? 'Unlock' : 'Lock'}
                   </span>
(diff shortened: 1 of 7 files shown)


asmsuechan:
229906e4bf2f86d41ec1b01e37d87a33
Also, I added a tooltip for the LockButton.

browser/main/Detail/MarkdownNoteDetail.js
```diff
@@ -219,6 +219,7 @@ class MarkdownNoteDetail extends React.Component {
     e.preventDefault()
     ee.emit('editor:lock')
     this.setState({ isLocked: !this.state.isLocked })
+    if (this.state.isLocked) this.focus()
```
asmsuechan: This is a bug fix for a problem which not to return Preview mode.

browser/main/Detail/MarkdownNoteDetail.js
```diff
@@ -229,8 +230,12 @@ class MarkdownNoteDetail extends React.Component {
     if (e.keyCode === 27) this.handleDeleteCancelButtonClick(e)
   }
 
-  handleShowLockButton () {
-    this.setState({editorStatus: this.refs.content.state.status})
+  handleToggleLockButton () {
+    if (this.props.config.editor.switchPreview === 'BLUR' && this.refs.content.state.status === 'CODE') {
```
asmsuechan: This is a bug fix for a bug which handles with switching preview by right click.

sota1235:
At dark theme, lock icon's color seems to be strange. Please fix it :pray:

image

asmsuechan:
I suppose the color of the lock is not wrong. The color of the trash can is wrong. Because the `color of buttons` in dark theme is set at https://github.com/BoostIO/Boostnote/blob/master/browser/styles/index.styl#L158:L166. It means the `+` button for creating a new note and the lock button have the same color. On the other hand, the icon of the trash can is made of svg. Totally different.

sota1235:
Please fix the style of trash icon on snippet note!

PR #336, opened 2017-03-19
Add a feature which enables to link between notes

asmsuechan:
Hi. I implemented a feature which is link between notes.

c83f3d8be4c391c20f72fbcf631b6c9f

### DIFF
--- a/browser/components/MarkdownPreview.js
+++ b/browser/components/MarkdownPreview.js
@@ -244,7 +244,6 @@ export default class MarkdownPreview extends React.Component {
       el.removeEventListener('click', this.linkClickHandler)
     })
 
-
     let { value, theme, indentSize, codeBlockTheme } = this.props
 
     this.refs.root.contentWindow.document.body.setAttribute('data-theme', theme)
--- a/browser/components/MarkdownPreview.js
+++ b/browser/components/MarkdownPreview.js
@@ -97,6 +97,8 @@ export default class MarkdownPreview extends React.Component {
     this.checkboxClickHandler = (e) => this.handleCheckboxClick(e)
     this.saveAsTextHandler = () => this.handleSaveAsText()
     this.saveAsMdHandler = () => this.handleSaveAsMd()
+
+    this.linkClickHandler = this.handlelinkClick.bind(this)
   }
 
   handlePreviewAnchorClick (e) {
@@ -238,6 +240,11 @@ export default class MarkdownPreview extends React.Component {
       el.removeEventListener('click', this.checkboxClickHandler)
     })
 
+    _.forEach(this.refs.root.contentWindow.document.querySelectorAll('a'), (el) => {
+      el.removeEventListener('click', this.linkClickHandler)
+    })
+
+
     let { value, theme, indentSize, codeBlockTheme } = this.props
 
     this.refs.root.contentWindow.document.body.setAttribute('data-theme', theme)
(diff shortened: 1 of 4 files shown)


browser/main/NoteList/index.js
```diff
@@ -163,6 +166,32 @@ class NoteList extends React.Component {
     ee.emit('list:moved')
   }
 
+  jumpNoteByHashHandler (event, noteHash) {
```
asmsuechan: I'd like to use this function for DRY.

asmsuechan:
This is bad layout for snippet.
screen shot 2017-04-07 at 13 48 53

asmsuechan:
Where should I put the link of the note to avoid that layout corruption in the snippet? I write my ideas. Any ideas?

## Slide the hash in snippet
Just I slide the `div` 240px. But actually, it's a little complicated because of this override https://github.com/BoostIO/Boostnote/blob/master/browser/main/Detail/SnippetNoteDetail.js#L583:L605.

screen shot 2017-04-07 at 14 13 11

## Share button
This button has a lot of scalability e.g. `Share in Evernote`, `Share in Slack`, etc.

screen shot 2017-04-07 at 14 10 14

I want not to implement it in this PR. Because I just want to add the feature of `link` for jumping note to note.

browser/components/MarkdownPreview.js
```diff
@@ -255,6 +261,10 @@ export default class MarkdownPreview extends React.Component {
       el.addEventListener('click', this.checkboxClickHandler)
     })
 
+    _.forEach(this.refs.root.contentWindow.document.querySelectorAll('a'), (el) => {
+      el.addEventListener('click', this.linkClickHandler)
```
sota1235: You can write this line at [this line](https://github.com/BoostIO/Boostnote/pull/336/files#diff-a54dec67725b3e604a34d39a7fcb72f6R257).

browser/components/MarkdownPreview.js
```diff
@@ -338,6 +348,14 @@ export default class MarkdownPreview extends React.Component {
     e.stopPropagation()
   }
 
+  handlelinkClick (e) {
+    const noteHash = e.target.hash
+    const regexIsNoteLink = /^#(.{20})-(.{20})$/
```
sota1235: Can you cut out this logic into the module?
If you do it, you can write unit test for the logic. (And I think this is business logic.)

browser/components/MarkdownPreview.js
```diff
@@ -338,6 +348,14 @@ export default class MarkdownPreview extends React.Component {
...
+    if (regexIsNoteLink.test(noteHash)) {
+      eventEmitter.emit('list:jump', noteHash.replace(/^#/, ''))
```
sota1235: [Question]Don't you need to exec `e.preventDefault()` after emit an event?

browser/main/NoteList/index.js
sota1235: You can do it by cut out this logic to the module.
Like this.

```javascript
export function jumpToNote(event, noteHash, router, location) {
  // brabra
}
```

You can use it on any other components.

asmsuechan:
Close this PR because one of the specifications is changed. I'm going to create a `share` button mentioned in https://github.com/BoostIO/Boostnote/pull/336#issuecomment-292655545.

Issue #281, opened 2017-02-11
Add ability to toggle the menu bar


delk0:
The expected behaviour on Windows (and linux, I guess) is that when you press the alt key it allows you to toggle the menu bar.
Or, also would be cool to have an option like some others apps do such as Atom for example where there's an option to hide it on the "View" section which this app also has. I think it would make Boostnote look a lot more cleaner.

kazup01:
Thank you for your feedback! Along with improving the current design, I will consider it.

klauscfhq:
Hi! 
I have created the #339 PR probably solving this issue as well as issue #169 .
I hope that it can be actually of some help!
Love the project 🎉

kazup01:
Thank you for your cooperation @klauscfhq san! Your work is great🎉 We will update it sometime soon!

Right2Drive:
What is the status of this? It doesn't look like the PR was ever merged?

KuangLei:
@Right2Drive 
>  It doesn't look like the PR was ever merged?

Yes, it's not merged now, but the Boostnote developers will add the feature sooner or later.  

If you want to hide the menu bar, 

>  You can add the feature to your local build just by switching the `autoHideMenuBar: showMenu` property to `autoHideMenuBar: true`, which is located inside the `resources/app/lib/main-window.js` file.

Thanks.

Issue #169, opened 2016-10-31
Request: Hide the toolbar on linux, unless ALT is pressed.


madc:
I'm using Boostnote on Linux/Gnome. It is common practice for apps on Gnome to hide the toolbar unless the ALT-key is pressed.

It would be great, if Boostnote would also handle it like this.

twhiting:
+1 For windows (ALT to toggle menu)

IssuehuntBot:
@zerox-dg has started working. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/169)

IssuehuntBot:
@zerox-dg has submitted [output](https://github.com/BoostIO/Boostnote/pull/1809). [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/169)

PR #150, opened 2016-10-17
Various updates

mikeres0:
See issue #139

PR #339, opened 2017-03-19
ALT key toggles the menu bar

klauscfhq:
Hi!
Now it's possible to have the menu bar hidden when launching the app, and only toggling it whenever you need it, just by pressing the `ALT` key!

🚀Tested both on **Windows 10 Pro** and **Linux Ubuntu 16.04**!

🔧It probably covers issues #281 & #169 

📷Here is a screenshot and an in action .gif running the windows build!

untitled-1

boostnote


asmsuechan:
Hi @klauscfhq! Thanks for your PR. Could you remove the variable you deleted? https://travis-ci.org/BoostIO/Boostnote/jobs/212747642
Travis says `error  'showMenu' is assigned a value but never used  no-unused-vars`.

asmsuechan:
Actually, I don't make sense this code @klauscfhq removed: `var showMenu = process.platform !== 'win32'`.
And I dug commits, I found this PR: https://github.com/BoostIO/Boostnote/pull/150. However, I don't have any ideas why he added this because it behaves only on mac thus it's useless. From these, I want to make a decision that it's his mistake to use ternary operator.

klauscfhq:
Sorry for that 😄 Of course! Happy to contribute to the app! 🎉

sota1235:
Thanks for contributing :)
We'll consider about this feature. Because the points like below.

- It is more smart design 🎨 
- But if the user don't know `alt` key feature, use can't never use menu bar features(export the note, confirm some information and etc...)

gotenyama:
Hi any update on this ? I've asked a similar question on #537 seems like it is waiting for PR to be merged by collaborators.

jasondavis:
This is a common feature among Windows apps to use the `ALT` key to toggle toolbar so I would say it's a good PR

klauscfhq:
@gotenyama @jasondavis You can add the feature to your local build just by switching the `autoHideMenuBar: showMenu`  property to `autoHideMenuBar: true`, which is located inside the `resources/app/lib/main-window.js` file.
It can be a temporary fix, until a better implementation is achieved, though you would still have to repeat it after an app update! 
Hope it helps 😄

PR #345, opened 2017-03-20
Fix the behaviour of Ctrl+E and Ctrl+W when switching preview by RIGHTCLICK

asmsuechan:
## before
Nothing happens when `Ctrl+E` is pressed on RIGHTCLICK mode. Also `Ctrl+W` doesn't work.

## after
They work well.


browser/main/Detail/index.js
```diff
@@ -14,7 +14,12 @@ class Detail extends React.Component {
     super(props)
 
     this.focusHandler = () => {
+      if (this.refs.root.state.status === 'CODE') return
       this.refs.root != null && this.refs.root.focus()
```
sota1235: [NOT MUST] It seems to be not necessary. Can you remove this line?

sota1235: Sorry, we can't remove this line.
Maybe fixing like below is better.

```javascript
if (this.refs.root != null) {
  this.refs.root.focus()
}
```

browser/main/Detail/index.js
```diff
@@ -14,7 +14,12 @@ class Detail extends React.Component {
...
+      let { config } = this.props
+      if (config.editor.switchPreview === 'RIGHTCLICK') {
+        this.refs.root.setState({status: ''})
```
sota1235: [NOT MUST] It is not good way to use React component.
Can you keep `status` state on this component and pass it as props to `this.refs.root`?

asmsuechan:
close this once.

PR #349, opened 2017-03-22
Fix the screen become black on closing Boostnote when a state of Boostnote is fullscreen on mac

asmsuechan:
I fixed it. But I don't have any ways to check the behavior of windows or Linux. Someone, please help me.

2b93d78c673a56aa189ded5cfc05f316

https://github.com/BoostIO/Boostnote/issues/332#issuecomment-288311835


lib/main-window.js
```diff
@@ -40,12 +40,19 @@ mainWindow.webContents.sendInputEvent({
 
 if (process.platform !== 'linux' || process.env.DESKTOP_SESSION === 'cinnamon') {
   mainWindow.on('close', function (e) {
+    e.preventDefault()
```
sota1235: 👍

Issue #288, opened 2017-02-14
Snippet tabs overwriting other tabs when closed


ericsolomon:
When you create a snippet note with multiple tabs, upon closing the tab, the editor of the tab you closed overwrites the contents of the tab to the right, effectively losing all of the content in the right tab. I will try to post a gif of this in action.

Steps to re-create:

1. Create snippet note
2. Add text into editor (e.g. "foo 123")
3. Hit '+' to create new tab
4. Add text into editor of new tab (e.g. "foo 234")
5. Navigate back to original tab
6. Click to 'x' in the original tab to close it
7. Tab that contained "foo 234" now contains original tab content "foo 123"

This only seems to occur when you close a tab that has a tab to its right. Closing a right tab does not copy the contents over to the left tab.

![snippet_bug](https://cloud.githubusercontent.com/assets/3520251/22924544/177353f4-f26b-11e6-8c29-a128f1333568.gif)

ericsolomon:
For the record, I have been looking into fixing this. I am not too familiar with developing an electron app though so I have had some issues into debugging the process. Is there an IRC/slack that I could discuss this?

asmsuechan:
Hi @ericsolomon, you can ask us anything here. Actually, the maintenance of any IRC is not in place so far.

asmsuechan:
And this snippet issue is a quite complicated. Thus I'm pleased you are attempting to fix it.

redcom:
One should not change directly the  state of the component using assignation or . 

this.state.*=* can lead to undesired behavior.


Try altering the code
'''
  deleteSnippetByIndex (index) {
    let snippets = this.state.note.snippets.slice()
    snippets.splice(index, 1)
    this.state.note.snippets = snippets
    let snippetIndex = this.state.snippetIndex >= snippets.length
      ? snippets.length - 1
      : this.state.snippetIndex
    this.setState({
      note: this.state.note,
      snippetIndex
    }, () => {
      this.save()
    })
  }
'''

this.state.note.snippets = snippets

this.setState accepts also a function so you can scope the note.snippets and work inside that closure.

redcom:
Sorry for bad formatting. I am on the road

asmsuechan:
Hi, @redcom. Why don't you send a PR?

redcom:
I was trying to actually run this application with npm run hot or npm run start or grunt and npm start.
I was not able to start it and there is no error in the electron javascript console
I will submit a patch.

[...]

PR #351, opened 2017-03-22
fix: #288 Fixed Snippet tabs overwriting other tabs when closed

redcom:
I also replaced the let with const because that is best practices. If the variable shape does not change it should be a const[ant]

### DIFF
--- a/browser/main/Detail/SnippetNoteDetail.js
+++ b/browser/main/Detail/SnippetNoteDetail.js
@@ -250,16 +250,20 @@ class SnippetNoteDetail extends React.Component {
   }
 
   deleteSnippetByIndex (index) {
-    let snippets = this.state.note.snippets.slice()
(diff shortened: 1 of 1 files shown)


asmsuechan:
According to this comment https://github.com/BoostIO/Boostnote/issues/288#issuecomment-288373268, do you still fail to run the electron app?

asmsuechan:
screen shot 2017-03-22 at 09 24 53

redcom:
How does it look now?

redcom:
screen shot 2017-03-22 at 17 49 24
my issue looks more like the above

asmsuechan:
Did you run `npm install`?

redcom:
Yes. I made another pull request. One has to run the webpack and npm start to get the application running. For the future you can use parallelshell to run 2 command lines at the same time.

PR #357, opened 2017-03-27
Add concurrently for running local environment more easily

asmsuechan:
Hi. There is a request to run npm tasks parallel. So I added `concurrently` and `npm run dev-start` task.


asmsuechan:
The reason for using `concurrently` instead of `npm-run-all` here is because it is easy to remove and I don't want to add such commands below for only 1 command: `dev:webpack` and `dev:hot`.

sota1235:
LGTM

PR #364, opened 2017-03-28
Uniform the accelerator notation

Hastegan:
I think it is more consistent to have the same notation everywhere.


asmsuechan:
@Hastegan Sorry for my late review. I checked this PR works fine. Thanks for your refactoring!

PR #368, opened 2017-03-31
Create ISSUE_TEMPLATE.md

VoLuong:


kazup01:
Thank you for your PR! But I think template is unnecessary. Because I think that there should be a free place.

Issue #369, opened 2017-03-31
Tags menu width


Myrga:
Hello,

We have found an issue with the tag's component. It has a fixed width (300px) : 
![tag_menu_before](https://cloud.githubusercontent.com/assets/15054308/24541792/3828aa66-15f9-11e7-8c1e-8d7e3092fa57.PNG)

We think it would be better if its width could take all the available space : 
![tag_menu_after](https://cloud.githubusercontent.com/assets/15054308/24541791/38254894-15f9-11e7-9b35-357455ac4f0c.PNG)

To get this result, we have taken out the width's property `width: 300px` (enable in the screenshot)
![developper_tools_screen](https://cloud.githubusercontent.com/assets/15054308/24541790/3823dbb2-15f9-11e7-948b-8a7e6cbdf601.PNG)

Cheers,
@Myrga and @Goulpe

Myrga:
Issue corrected in #370

PR #370, opened 2017-03-31
Update TagSelect.styl

Myrga:
See issue #369


kazup01:
Thank you for your support @Myrga san!

PR #386, opened 2017-04-05
Fix for 'npm prune --production' deleting needed dependencies

clone1612:
Encountered this little bug when compiling source code for '_production_' with the cause for this being the fact that two dependencies are also present in the devDependencies list.

**Problem**
`npm prune --production` is also deleting actual needed dependencies

**Cause**
Both `react` and `react-dom` are mentioned in both dependencies and devDependencies

**Solution**
Remove both packages from the devDependencies list


asmsuechan:
@clone1612 Thanks!

Issue #42, opened 2016-06-02
Auto-insert bullet in list


JasonBoccuti:
When writing an unordered list in markdown, Boostnote inserts a new asterisk as you hit return in the list.  But, if you hit return twice (to start writing new content outside the list) that last asterisk should delete itself.

This would really help with productivity while writing!

Rokt33r:
Great idea. 
The only problem is that I don't know how to modify the behaviour of Ace editor. Anyway, I'll give a shot.
If you know some kind of hint to embed it, please teach me.

Rokt33r:
https://github.com/ajaxorg/ace/blob/master/lib/ace/mode/markdown.js#L59-L71
It occurs here. But, it doesn't seem to be possible to do that by just modifying method, `getNextLineIndent`. I'll try to find another approach. :)

Rokt33r:
Closed! You can track the progress on TodoList #68

PR #425, opened 2017-04-14
Fix new line problem in list and task list

AWEEKJ:
Hello. I changed the code to use `newlineAndIndentContinueMarkdownList` which is defined in `node_module/codemirror/addon/edit/continuelist.js`of library named`codemirror`. It'll fix the new line issues in list and task list. 

This will cover these issues in [List of the requested features](https://github.com/BoostIO/Boostnote/wiki/List-of-the-requested-features#markdown-note)
- [ ] Remove last list item when Enter pressed on empty list item. [#42](https://github.com/BoostIO/Boostnote/issues/42)
- [ ] Task list should append - [ ] instead of - when Enter pressed.

### Before
before
### After
after

### DIFF
--- a/browser/components/CodeEditor.js
+++ b/browser/components/CodeEditor.js
@@ -83,37 +83,7 @@ export default class CodeEditor extends React.Component {
         'Cmd-T': function (cm) {
           // Do nothing
         },
-        Enter: (cm) => {
-          const cursor = cm.getCursor()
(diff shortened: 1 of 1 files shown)


kazup01:
Wow, great!! Thank you for your cooperation @AWEEKJ san 🎉 We'll check it!

Ilyes-Hammadi:
Great work 👍  🥇

PR #453, opened 2017-04-19
double click on tab to rename

bubuzzz:
Double click is more convenient i guess. When opening a new snippet, the first thing i have to do is rename the first tab, thus rename is always the thing need to be done first and with the current flow, i have to right click, select rename, which is slow


kazup01:
Thank you @bubuzzz san! Great! We will check it🎉

PR #457, opened 2017-04-20
Updated initial pieces of text to read more clearly in English

SalvatoreTosti:
Made a few small tweaks so the wording makes more sense to English speakers.


kazup01:
Thank you for your cooperation! We will check it!

asmsuechan:
Thank you!

PR #465, opened 2017-04-21
Fix the loading layout

kazup01:
loading


kazup01:
Thanks <3 @sosukesuzuki

Issue #151, opened 2016-10-18
Presentation mode


asmsuechan:
Hi,
I'd like to use boostnote as a presentaton tool.

So I want 3 function below
- fullscreen mode
- pagination
- pasting images

Perhaps, are there in it already?
Thanks!

Rokt33r:
Pasting images and Fullscreen mode is currently developing.

But, I can not understand *pagination* function.
Could you explain more about this? Did you mean [Marp](https://yhatt.github.io/marp/)?

mikeres0:
I believe he means pagination within the notes, sections as such.

Rokt33r:
I need more explanation how should it be realized.
Does it mean we should make a new format of note, like Wiki or Gitbook?

asmsuechan:
Sorry for my absent.
I mean it's pagination like Marp, not Wiki or Gitbook.
Perhaps, it's nice to implement a function which exports to Marp.

May I ask if you understand what I mean?

Rokt33r:
@mikeres0
I and @asmsuechan talked at slack. He want pagination function only for presentation mode. 
We might be able to use [gnab/remark](https://github.com/gnab/remark). But, I think we should mimic this library because it uses another markdown renderer.

Flexo013:
@asmsuechan Is this feature still desired by you?

arcturus140:
2 out of 3 functions are already supported

Flexo013:
As @arcturus140 pointed, the only thing that isn't present currently is pagination.

@asmsuechan If you would really like this feature then feel free to open up a new issue with the feature request (preferably mentioning this issues number 151).

Issue #94, opened 2016-08-24
Other Features


rednuz:
First of all, congratulations. This project is fantastic.

I have been looking exactly for this for over a week; A blend between Atom, Quiver and Evernote. Simply perfect. Seriously, I am very, very impressed and excited for Boostnote.

I thought I would add some keypoints that would make this little app the ultimate writing tool:
- Menubar quick note scratchpad (see MiniNote Pro [menubar icon needs to be black and white btw]).
- Word and character counter.
- Extensive theme styling and customization (I am using OS X and its very limited).
- Multiple note selection for deletion (I would personally get rid of the blue border and just add a bit of a darker background for selection: http://i.share.pho.to/4bca6da7_o.png).
- Proper dropbox integration with encryption (see: https://laverna.cc/).
- Each storage saves to separate folder. Each snippet saves in a subfolder with the according file syntax extension e.g.: MyStorage>Default(4bc6eb15d8b8)>MySnippet>TabOneFilename.html
- Import folders to left sidebar (dropdown tree structure).
- Distraction free and dual panel live preview modes.

That's all I can come up with for now. I would love to donate for this project too if you have a link you should add it to the website.

Rokt33r:
I'm very very excited your feedback.

Btw, these are my answer to each item.

> Menubar quick note scratchpad

Check #85. There are some comment about this. We can talk more about this issue.
And I agree with that the icon should be black & white.

> Word and character counter

Should I count all of characters(including markdown syntax) or only rendered text? 

> Extensive theme styling and customization

I was going to implement it on the last update, v0.6.3. I'd tried to set the colors to inline style instead of CSS. But, I noticed that, like `:hover`, some selector isn't available on inline style.
So, this is currently pending now. 
But, It doesn't mean I gave up. I think it can be done by re-compiling customized style and saving it to some kind of cache(localStorage would be fine). So, It would be a matter of time.

> Multiple note selection for deletion

It should be. Btw, I want to ask something. Is there any other function for multiple note selection?
I think moving notes to other storage or folders might be a valid choice.

FYI, OS X Notes only provides these options.

<img width="123" alt="screen shot 2016-08-25 at 10 12 01 am" src="https://cloud.githubusercontent.com/assets/5865853/17953212/756219b0-6aac-11e6-8910-212ec6eef8dc.png">


> (I would personally get rid of the blue border and just add a bit of a darker background for selection: http://i.share.pho.to/4bca6da7_o.png)

[...]

PR #491, opened 2017-04-23
Feature: Toggle Fullscreen mode

klauscfhq:
Hello!
Now it's possible to toggle the app in and out of the fullscreen mode just by pressing the <kbd>F11</kbd> key on Windows/Linux and <kbd>Command</kbd> + <kbd>F</kbd> on MacOS! Feel free to modify the key combinations if needed!

fullscreen-feature-screen

🚀 Tested on Windows 10 Pro, Linux Ubuntu 16.04 & Mac OS Sierra!

🔧 Solves partially issues #151 & #94 

📷 Here is an in action gif running the windows build!

Start taking notes distraction free! 🎉

### DIFF
--- a/lib/main-menu.js
+++ b/lib/main-menu.js
@@ -155,15 +155,15 @@ var view = {
     {
       label: 'Toggle Full Screen',
       accelerator: (() => {
-        if (process.platform === 'darwin') return 'Ctrl+Command+F';
-        return 'F11';
+        if (process.platform === 'darwin') return 'Ctrl+Command+F'
+        return 'F11'
       })(),
       click: (item, focusedWindow) => {
         if (focusedWindow) {
-          focusedWindow.setFullScreen(!focusedWindow.isFullScreen());
-          focusedWindow.send('window:fullscreen', { state: focusedWindow.isFullScreen() });
+          focusedWindow.setFullScreen(!focusedWindow.isFullScreen())
+          focusedWindow.send('window:fullscreen', { state: focusedWindow.isFullScreen() })
         }
-      },
+      }
     },
     {
       label: 'Reload',
--- a/lib/main-menu.js
+++ b/lib/main-menu.js
@@ -153,6 +153,19 @@ var view = {
   label: 'View',
   submenu: [
     {
+      label: 'Toggle Full Screen',
+      accelerator: (() => {
+        if (process.platform === 'darwin') return 'Ctrl+Command+F';
+        return 'F11';
+      })(),
+      click: (item, focusedWindow) => {
+        if (focusedWindow) {
+          focusedWindow.setFullScreen(!focusedWindow.isFullScreen());
+          focusedWindow.send('window:fullscreen', { state: focusedWindow.isFullScreen() });
+        }
+      },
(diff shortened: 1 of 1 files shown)


lib/main-menu.js
```diff
@@ -153,6 +153,19 @@ var view = {
   label: 'View',
   submenu: [
     {
+      label: 'Toggle Full Screen',
+      accelerator: (() => {
+        if (process.platform === 'darwin') return 'Ctrl+Command+F'
+        return 'F11'
```
asmsuechan: Could you change like below?

```
accelerator: OSX ? 'Control + Command + F' : 'F11',
```

And I'm not sure F11 is available on Linux or not. Do you have any ideas?

asmsuechan:
Thank you @klauscfhq for opening a PR!

lib/main-menu.js
```diff
@@ -153,6 +153,19 @@ var view = {
...
+      })(),
+      click: (item, focusedWindow) => {
```
asmsuechan: What does `item` mean?

lib/main-menu.js
klauscfhq: Sure, switching to conditional expression 
```js
accelerator: macOS ? 'Control+Command+F' : 'F11',
```
Yes, F11 is available in Linux :)

lib/main-menu.js
klauscfhq: It's one the optional variables defined when calling the `click` function and it corresponds to the MenuItem. You can get a better idea about it [here](http://electron.atom.io/docs/api/menu-item/#new-menuitemoptions), the official electron API docs :)

asmsuechan: Got it :smile: So could you add a comment like `// item is not used` under L161?

klauscfhq: Actually it is used and also needed, because when building the menu-item `Toggle Full Screen`, the `click` property is expected to be a callback with two parameters: 
- `item`, represents the `Toggle Full Screen` object inside the menu, and clicking it triggers the execution of the below of it defined code 
- and `focusedWindow`, which works as the bridge between the menu and the main app process allowing us to pass through the message inside of the [send() function](https://github.com/BoostIO/Boostnote/pull/491/files#diff-c4d0b50cf156e17e35d992365a6fb80aR161)

I know that it is a bit confusing, but this is how electron operates, and without defining it inside of the `click` property the code will result in an error. It is displayed more on electon's official [menu-example](https://github.com/electron/electron-docs/blob/master/test/fixtures/menu.md#menu), where you can spot it after scrolling down a bit : )

I hope my explanation was somehow helpful! 😄

asmsuechan: @klauscfhq Thank you for this good information :smile: Year, my opinion is: `item` is used *inside* the menu, in other words, `item` is *not* used in our code. This is why I asked you to add a comment. Is this different from your opinion?

klauscfhq: I totally agree in this case 😄

asmsuechan:
Sorry for long absent, but we decided not to merge this PR.

PR #502, opened 2017-04-25
Add spectron

asmsuechan:
I added spectron as an e2e test.

## how to run
You need to run `$ grunt pre-build` before this e2e test. Because I could not run `$ electron-prebuild ./index.js` due to relative paths around [here](https://github.com/BoostIO/Boostnote/blob/master/lib/main.html#L56).

```
$ grunt pre-build
$ yarn run test:e2e
```


.travis.yml
```diff
@@ -2,5 +2,15 @@ language: node_js
...
-script: npm run lint && npm run test
+before_script:
+  - 'npm install grunt-cli -g'
+  - 'export DISPLAY=:99.0'
+  - 'sh -e /etc/init.d/xvfb start'
+  - 'sleep 3'
+script:
+  - 'npm run lint'
+  - 'npm run test'
+  - 'if [ ${TRAVIS_NODE_VERSION} = "stable" ]; then
```
asmsuechan: I make it that it runs only `stable` because there is no support of ES7 syntax in `lts/*`.

asmsuechan: I assume cutting out into `runner.sh` or something is also better but I prefer to write this procedure here rather than that way. Because I'm not sure where I should put the file: I thought `tests/e2e/runner.sh` and `scripts/e2e-runner.sh` but both is not the best way because I thought this procedure is too short to cut out.

Issue #509, opened 2017-04-26
Copy note in viewer mode--looses EOLs


Alexey-T:
copied all note from MD note in viewer mode. Pasted to other app.
All EOLs lost.
If copy in editor mode-- EOLs are ok.

Linux Ubuntu 

![note--](https://cloud.githubusercontent.com/assets/4154978/25420585/726582fe-2a61-11e7-8884-6b4970fc0bd5.png)

LeoLamCY:
This can be fixed by passing `breaks: true` to markdown-it so line breaks will be rendered as <br>, but I'm not sure if it has been left as default (false) deliberately by @kazup01.

kazup01:
Thanks for your support @LeoLamCY san!

PR #512, opened 2017-04-26
Fix md viewer line breaks when copying

LeoLamCY:
#509
-set `white-space` to normal for `p` to remove extra space between lines


kazup01:
merged! Thank you @LeoLamCY san🎉

PR #522, opened 2017-04-28
fix: modify the star layout at NoteItem

kazup01:
apr-29-0029 05-58-00


kazup01:
Thanx @sosukesuzuki !

PR #527, opened 2017-04-29
Add description for package.json

sota1235:


package.json
```diff
@@ -1,7 +1,8 @@
 {
-  "name": "Boostnote",
+  "name": "boost",
```
sota1235: Its for backward compability.

Issue #535, opened 2017-04-30
[BUG] Two scrollbars appears on markdown note (Windows 10)


LetItRock:
Hello,
I've found this bug when I was creating new markdown note:

![przechwytywanie](https://cloud.githubusercontent.com/assets/2607232/25565705/b12bdcea-2dcc-11e7-996b-af3e07ada173.png)

As you can see from image above - two scrollbars appeared, one was from "Edit mode" and another from "Preview mode". When I switched to "Edit mode" the latter one has disappeared.

P.S. Really appreciate your work and love this app :)

sota1235:
Thank for your feedback!
Can you paste the text of its note?

NonlinearFruit:
This happens on Linux (Ubuntu 16.04) as well.

kazup01:
Fixed at current version. Thanks for reporting!

PR #536, opened 2017-04-30
fix viewer showing 2 scrollbars

LeoLamCY:
fix scroll bar for codemirror showing through #535


kazup01:
Thank you for your cooperation @LeoLamCY san! We'll check it!

PR #545, opened 2017-05-03
Revamp handling of multiple key press

### LINKED ISSUE #540: Shift / ctrl key introducing asterisks
I hope this can be re-produced on your end.

When using markdown editor, try typing:

"hello:"

then pressing "ctrl+n" (for e.g. ) or just ctrl... in my case I get asterisks inserted (not always)... I use (ctrl + n, ctrl + p, ctrl + a, ctrl + e) from muscle memory to go to the next line, go to the previous line, go to the beginning of a line, and go to the end of a line.

asterisks-issue

LeoLamCY:
- use keyCodes instead of strings to detect keys
- delete key from keyPressed array on key up
- fix #540

### DIFF
--- a/browser/components/MarkdownEditor.js
+++ b/browser/components/MarkdownEditor.js
@@ -9,10 +9,10 @@ class MarkdownEditor extends React.Component {
   constructor (props) {
     super(props)
 
-    //char codes for ctrl + w
+    // char codes for ctrl + w
     this.escapeFromEditor = [17, 87]
 
-    //ctrl + shift + ;
+    // ctrl + shift + ;
     this.supportMdSelectionBold = [16, 17, 186]
 
     this.state = {
@@ -89,7 +89,7 @@ class MarkdownEditor extends React.Component {
 
   handleBlur (e) {
     if (this.state.isLocked) return
-    this.setState({ keyPressed: new Set()})
+    this.setState({ keyPressed: new Set() })
     let { config } = this.props
     if (config.editor.switchPreview === 'BLUR') {
       let cursorPosition = this.refs.code.editor.getCursor()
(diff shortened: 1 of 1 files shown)


asmsuechan:
Hi, @LeoLamCY! Thank you for this PR! I'd like to ask you a question. Why do you think numbers is better than string in keyCodes? I think a string is readable.

LeoLamCY:
please refer to #540 
If key codes are used, that bug will not occur when user try to input something combining the `shift` key because for example in key codes, the key `:` is handled as `shift + ;` (16, 186) instead of being treated as `:`. So when user releases `shift` before `;`, both key codes are removed from `keyPressed` properly. 
While previously when Strings are used, `shift + ;` is pushed into `keyPressed` as `{Shift: true, ;: true, :: true}` so when user releases `shift` before releasing `;`, then `shift` and `;` would become `false` in `keyPressed` but `:` would remain `true`, resulting the bug in #540

asmsuechan:
Oh, I understand. The keyboard difference between US and others makes this issue difficult. I'm using JP keyboard. But no worry, it's my mistake. We should implement shortcuts in US keyboard.

Thank you so much for informing me!

BTW, do you use `Ctrl + :` shortcut? Actually, I've never used it. So I'm contemplating to delete it.

LeoLamCY:
I don't use it personally but I think it doesn't hurt to have a few keyboard shortcuts in the program. Some people might use them, you never know :)
I would rather the shortcut be `ctrl + b` for bold though, don't even need to hit `shift` then.

asmsuechan:
Yes, exactly true...! Umm, `Ctrl + B` has a conflict problem with vim key-bind. I know it's nonsense from who doesn't use vim key-bind.

LeoLamCY:
By the way, is it really caused by the difference between US and JP keyboards?
I believe it is caused by the `shift` key being released early causing `keyPressed` to have incorrect values.

- hit `shift + ;` will result in `{Shift: true, ::true}`
- release `shift` first while still holding `;` will result in `{Shift: false, ::true, ;:true}`
- release `;` now will result in `{Shift: false, ::true, ;:false}`
- `:` remains `true` in `keyPressed` even though all keys are now released, which causes the bug

asmsuechan:
In JP keyboard, `:` doesn't use `Shift`. Just `{ 'Ctrl': true, ':' true}`.

LeoLamCY:
Oh ok, good to know :)

justin-calleja:
@LeoLamCY @asmsuechan ... ye please don't use `ctrl + b` it's not just for vim mode. That key combo goes back (like hitting arrow right) in normal mode (it's an Emacs keybinding... or whatever they're called there - and it's probably the default in your terminal program too).

As a side note, in vim ports across editors, being in insert mode also allows the use of these emacs shortcuts allowing one to stay in insert mode to do minor movements while still keeping your hands in home position (and without going back to normal mode). i.e. vim mode or not... these shortcuts are used across many editors and are likely to be ingrained in some ppl's muscle memory.

For the record, I don't personally use `ctrl + :`.

Also, @asmsuechan - you had an issue open about a "dot config" file for Boostnote (which I can't seem to find now). That's a great idea and would probably be the ideal place to map these keyboard bindings - so users can change according to personal prefs.

asmsuechan:
Hi, @justin-calleja! I got it. `.boostnoterc` will support to change key-bind in our future :star2: I try to implement it as soon as possible 🙇

PR #548, opened 2017-05-04
Add a module to find the title

asmsuechan:
I added a module to find a title of a note.

### DIFF
--- a/browser/lib/findNoteTitle.js
+++ b/browser/lib/findNoteTitle.js
@@ -1,0 +1,32 @@
+export function find (value) {
+  let splitted = value.split('\n')
+  let title = null
+  let isMarkdownInCode = false
+
+  for (let i = 0; i < splitted.length; i++) {
+    let trimmedLine = splitted[i].trim()
+    if (trimmedLine.match('```')) {
+      isMarkdownInCode = !isMarkdownInCode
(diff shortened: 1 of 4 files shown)


browser/lib/findNoteTitle.js
```diff
@@ -0,0 +1,32 @@
+export function find (value) {
```
sota1235: [Question] Why didn't you name this method like `findNoteTitle`?
I think `find` is a little ambiguous.

asmsuechan: Because I don't like 
repetition like `findNoteTitle.findNoteTitle()`. What do you think `noteUtils.findNoteTitle()`?

The reason why I didn't choose that name is because I add only 1 method this time. I thought don't need to change the name to `noteUtils` unless need to add a lotta methods.

sota1235: Is the way like below is better for you?

```javascript
export function findNoteTitle {
  //
}

export default {
  findNoteTitle,
}
```

```javascript
import { findNoteTitle } from 'browser/lib/findNoteTitle'
```

asmsuechan: Oh, yes! It's the better way :smile:

asmsuechan:
@sota1235 I fixed the name :pray:

PR #583, opened 2017-05-18
Fix the position of a tooltip of new note

asmsuechan:
### before
screen shot 2017-05-18 at 14 26 20

### after
screen shot 2017-05-18 at 14 26 08


kazup01:
I think this tooltip should not put in specific position.

asmsuechan:
This is a stupid...

Issue #595, opened 2017-05-24
Resize an image with markdown


darkomen:
Hí!
I'm trying to put an image in oe note, but when I have tried to change the size with the follow code it doesn't work.
`![Alternate text](path/to/image =150x50)`
How I could to change the size of an image?

kazup01:
Hi @darkomen san!
It can not do it at current version, but I also want it. If possible, could you develop it and send us a Pull request?

darkomen:
I haven't programmed in JavaScript 😓 But I will try to do the implementation. If I get something I'll make a pull request

kazup01:
Thank you for your suggestion @darkomen , and thank you for your cooperation @dangvanthanh !
Released on v0.8.10! Enjoy 🙌

PR #598, opened 2017-05-26
#595 - Resize an image with markdown

dangvanthanh:
- Update resize an image with markdown #595
- Use

```markdown
![](/path/to/image.png =100x200)
```


kazup01:
Wow @dangvanthanh san, thank you for your cooperation everyday! We will check it!

kazup01:
Cool!

darkomen:
great!!! Thanks you very much!

joelcwe:
@aquCHAOS i think you can leave the width out and it should scale correctly.
`` image.png =100x  ``

Issue #619, opened 2017-06-06
[BUG] Code Block Format Error！


lyrl:
### Sample Code Block 

```ini
http {
                upstream my_nodejs_upstream {
                        server 127.0.0.1:3000;
                        keepalive 64;
                }

        server {
            listen 80;
            server_name example.com;
            #root /home/www/project_root;
            #return 301 $scheme://example.com$request_uri;
            
            if ($host !~ '^[a-zA-Z0-9]*.{0,}example.com$') {
                return 403;
                #return 301 $scheme://example.com$request_uri;
                }

            location / {
                proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
                proxy_set_header Host $http_host;
                proxy_set_header X-NginX-Proxy true;
                proxy_http_version 1.1;
                proxy_set_header Upgrade $http_upgrade;
                proxy_set_header Connection "upgrade";
                proxy_max_temp_file_size 0;
                proxy_pass http://my_nodejs_upstream/;
                proxy_redirect off;
                proxy_read_timeout 240s;
            }
        }

}

```

### Rendered As Below

![](http://ongd3mgn7.bkt.clouddn.com/17-6-6/38468450.jpg)

asmsuechan:
Hi, @lyrl. Thank you for reporting. I'll fix it.

kazup01:
Released on v0.8.10! Enjoy!

PR #633, opened 2017-06-12
Fix a rendering bug

asmsuechan:
As you see, I added an entity (`$`) to the methods to handle the issue below.

refs: https://github.com/BoostIO/Boostnote/issues/619


asmsuechan:
screen shot 2017-06-12 at 18 00 04

PR #647, opened 2017-06-17
Fix dependency type of Spectron module

clone1612:
Spectron is a devDependency so I moved the package there where it belongs. This helps when building from source to reducere the size of the Boostnote package since we do not need Spectron in a production environment.


asmsuechan:
Thank you!

Issue #646, opened 2017-06-17
Can't copy text in Vim binding mode


IzumiSy:
I can't copy any texts in Vim mode, because `Ctrl + C` is used in Vim binding to change edit mode. However, yank also does not share its data with system clipboard, so I cannot copy any texts and paste it outside of Boostnote. Any idea to solve this issue?

IzumiSy:
In my case, always when I want to copy texts on Boostnote, I change keybindings from Vim to default. This is too bothering, but no way out.

kazup01:
This is a good point. I'm not using Vim mode, but I want to improve it.

Just idea:
1, Allow right click and make copy / paste possible.
2, Allow key setting for copy / paste.

---

BTW, how about Finder or Copy at codeblock?

- Finder
![finder](https://user-images.githubusercontent.com/8602615/27253410-1907d492-53af-11e7-8873-ba7914cb0e6e.gif)

- Copy at codeblock
![copy](https://user-images.githubusercontent.com/8602615/27253412-2b2b6be8-53af-11e7-8058-a6f85b4eb8dc.gif)

kazup01:
Hi @IzumiSy ,
This is a good point. I'm not using Vim mode, but I want to improve it.

Just idea:
1, Allow right click and make copy / paste possible.
2, Allow key setting for copy / paste.

---

BTW, how about Finder or Copy at codeblock?

- Finder
![finder](https://user-images.githubusercontent.com/8602615/27253410-1907d492-53af-11e7-8873-ba7914cb0e6e.gif)

- Copy at codeblock
![copy](https://user-images.githubusercontent.com/8602615/27253412-2b2b6be8-53af-11e7-8058-a6f85b4eb8dc.gif)

IzumiSy:
Thanks for replying.

> 1. Allow right click in editor and make copy / paste possible
> 2. Allow key setting for copy / paste.

Cool if one of those two is possible, but how do I achieve that from settings? (Or would it be a future implmentation? nice too!)

> how about Finder

My machine is Ubuntu 16.04 with Unity. I am not sure if my environment affects in some bad ways or not, but I can't open Finder. This pic is my hotkey preference which does not work. Is this a bug?

![selection_027](https://user-images.githubusercontent.com/982850/27253648-70eae376-53b3-11e7-8514-a502b59b899d.png)

For further information, it looks I am getting no error.

[...]

PR #649, opened 2017-06-18
Make Ctrl-C copy

asmsuechan:
I added copy feature into the existing shortcut (C-c). It doesn't prevent the ordinal feature.

ref: https://github.com/BoostIO/Boostnote/issues/646


kazup01:
q4c7g8c - imgur

Issue #652, opened 2017-06-20
Dragging and dropping images overwrites existing images with no warning


vvs:
When you drag'n'drop the image into the code editor, the image is being copied and overwrites previously existing image in the STORAGE/images folder.

It would be nice to warn about this, and provide some means for user to resolve the issue. Or, maybe, just append number at the end of the file name to make the new filename unique.

asmsuechan:
Hi, @vvs. I believe it's good to solve this by naming a random name to the file is good.

asmsuechan:
Fixed.

yashha:
How do you drop something?
When I want to drop an image I'm back in Preview mode.

PR #667, opened 2017-06-28
Randomize the image name when it's dropped

asmsuechan:
screen shot 2017-06-28 at 13 07 34

I changed an image name to a random string to avoid duplication when it's dropped. And remove `encodeURI()`.

```
## before
![file%20name.jpg](/Users/asmsuechan/Boostnote/images/filename.jpg)

## after
![file name.jpg](/Users/asmsuechan/Boostnote/images/skgvvqw0ozbprpb9.jpg)
```

`%20` changes to ` ` and `filename` changes to `skgvvqw0ozbprpb9`.

ref: https://github.com/BoostIO/Boostnote/issues/652

### DIFF
--- a/browser/components/CodeEditor.js
+++ b/browser/components/CodeEditor.js
@@ -193,7 +193,7 @@ export default class CodeEditor extends React.Component {
     const filename = path.basename(imagePath)
 
     copyImage(imagePath, this.props.storageKey).then((imagePathInTheStorage) => {
-      const imageMd = `![${encodeURI(filename)}](${imagePathInTheStorage})`
+      const imageMd = `![${filename}](${imagePathInTheStorage})`
       this.insertImageMd(imageMd)
     })
   }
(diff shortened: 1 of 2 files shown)


samontea:
What happens if there's a naming collision? Will this clobber (i.e. overwrite) the existing image?

asmsuechan:
Yes, it'll override for now. But it's very rare. So you should be delighted when it happens :laughing:

thomas-lange-smf:
1.) I think using a hash function (even md5) and do not overwirte on collisions would be better
Obv. sha512 would be less prone to collisions. but more expensive.

2) This will not work on linux (ubuntu xenial / linux mint): it resolves the path to (example)
file:///usr/share/boostnote/resources/app/lib/images9vdwtwps2e6irudi.jpg

Not shure why. But I am on Boostnote 0.8.12

thomas-lange-smf:
Quick and Dirty fix for single user:

`sudo ln -s /home/<username-goes-here>/<note-folder-goes-here>/images images`
so mine was:

`sudo ln -s /home/letigre/Dropbox/Notes/images images`

PR #672, opened 2017-07-01
Componentnize TrashButton

asmsuechan:
I change it into a component.


kazup01:
Great! Sorry, I forgot to send you a trash can UI. How about like this?

trash_can_ui

asmsuechan:
@kazup01 No worry. Thanks! But actually, it's a refactoring. So the icon is not related for now :cry:

PR #674, opened 2017-07-02
Add LockButton component

asmsuechan:
I change the LockButton into a component.


asmsuechan:
Sadly, it doesn't work fine so I close this PR.

PR #686, opened 2017-07-07
Add search-box

asmsuechan:
Because of its high demand, I finally implemented it.
screen shot 2017-07-07 at 17 14 38

PR #696, opened 2017-07-09
Replace cm searchbox to official search addon

asmsuechan:
I want to replace `cm-searchbox` to the official search addon because it's very hard to customize.

### Pros
* Easy to maintain
* Easy to customize
* Officially supported

### Cons
* Unintuitive behavior

## ScreenShots
screen shot 2017-07-09 at 22 49 18

And vim-search is available (Look at the bottom of CodeEditor).
screen shot 2017-07-09 at 22 49 26

### DIFF
--- a/browser/components/CodeEditor.js
+++ b/browser/components/CodeEditor.js
@@ -57,7 +57,6 @@ export default class CodeEditor extends React.Component {
       keyMap: this.props.keyMap,
       inputStyle: 'textarea',
       dragDrop: false,
-      searchbox: true,
       extraKeys: {
         Tab: function (cm) {
           const cursor = cm.getCursor()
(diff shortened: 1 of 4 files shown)


asmsuechan:
ref: https://github.com/BoostIO/Boostnote/pull/686

BobTB:
Is this included / released in the latest version? Without full text search the thing is almost useless.

asmsuechan:
@BobTB Yes, it is. https://medium.com/boostnote/release-information-7-29-2017-v0-8-12-5ec72bb88ec

BobTB:
OK, it took me a while ti find it! It is only for searching the currently selected note. A really good search would be a single one, global for the whole app, which would filter the notes containing the searched string and at the same time showing it on the "edit" panel highlighted for the currently selected note. Search as you type thing is also good, not that I have to enter the search string and then hit enter for it to show the results. 

Check out resophnotes ( http://resoph.com/ResophNotes/Welcome.html ), a 7 years old app, it has all of this and works just beautifully, this is how the search functionality should be implemented.

vgrange:
+1 for having a global full-text search on snippet content as described by @BobTB

PR #715, opened 2017-07-22
Add checkbox for ama

asmsuechan:
screen shot 2017-07-22 at 08 58 57

ref: https://github.com/BoostIO/Boostnote/pull/620

### DIFF
--- a/browser/main/lib/AwsMobileAnalyticsConfig.js
+++ b/browser/main/lib/AwsMobileAnalyticsConfig.js
@@ -3,7 +3,7 @@ const AMA = require('aws-sdk-mobile-analytics')
 const ConfigManager = require('browser/main/lib/ConfigManager')
 
 AWS.config.region = 'us-east-1'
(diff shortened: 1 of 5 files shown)


asmsuechan:
image

browser/main/lib/AwsMobileAnalyticsConfig.js
```diff
@@ -3,7 +3,7 @@ const AMA = require('aws-sdk-mobile-analytics')
 const ConfigManager = require('browser/main/lib/ConfigManager')
 
 AWS.config.region = 'us-east-1'
-if (process.env.NODE_ENV === 'production') {
+if (process.env.NODE_ENV === 'production' && ConfigManager.default.get().amaEnabled) {
```
asmsuechan: This condition is necessary because unless it exists, the user, who just disable ama,  cannot disable sending data without reloading.

PR #717, opened 2017-07-22
Fix a bug on click

asmsuechan:
# context
An error as follows happens when I clicked `cancel` button on `delete dialog box`.

```
Uncaught TypeError: e.target.className.includes is not a function
```

Although it causes a bug, the application works fine.

# before
image

# after
The application works fine without the error.


asmsuechan:
ref:
(Add info button)https://github.com/BoostIO/Boostnote/pull/692
(Fix an error on click event)https://github.com/BoostIO/Boostnote/pull/695

PR #722, opened 2017-07-24
Fix layout

kazup01:
I fixed following layout.

- InfoTab
- Export icon at InfoPanel
- Overall button

PR #723, opened 2017-07-25
Fix typo in InfoTab

asmsuechan:
# context
There's a typo in https://github.com/BoostIO/Boostnote/pull/722.

# before
The part of a sentence was `We collect only the number of DAU for Boostnote and DO NOT COLLECTED any detail information`.

# after
`We collect only the number of DAU for Boostnote and DO NOT collect any detail information`


kazup01:
LGTM

PR #736, opened 2017-07-29
Fix layout

kazup01:


kazup01:
- Before
screen shot 0029-07-29 at 1 50 16 pm

- After
screen shot 0029-07-29 at 1 50 20 pm

PR #754, opened 2017-08-01
Add Print

asmsuechan:
# context
It's easy to export as pdf.

# before
There was no print feature.

# after
c17f20c6d07a1f126d35d9887b6b4a76

# note
You can export as pdf through this.


asmsuechan:
The saved pdf is this.
image

kazup01:
Great🎉✨

PreeyaDahya:
When I try to export it as a pdf it only shows the first page, every page after is blank. How can I fix that?

szjani:
Is it available on Windows? I cannot find it but by friend sees it on his Mac.
Using `Microsoft Print to PDF ` yields weird result: e.g.: font glitches and no link support.

silverjava:
I've tried to use Printer to export PDF, but like @PreeyaDahya said only the first page has content, rest are all blank. Is there any way to fix it?

PR #756, opened 2017-08-01
Adapted height of dropdown fields to the height of the input fields

yashha:
What do you think?
Before:
screenshot from 2017-08-01 23-01-54
After:
screenshot from 2017-08-01 22-44-32


kazup01:
Thanks for you PR!
This is looks unnecessary, so I close this.
But as you said, the UI of this part should be improved, so I'll fix it later.

Issue #757, opened 2017-08-01
Trashed items are shown in the badge of the storage


yashha:
Maybe it's better to exclude the count of the trashed notes in the badge.

![screenshot from 2017-08-02 00-44-33](https://user-images.githubusercontent.com/4728786/28850373-8491d0a6-771c-11e7-83e4-ae4bc1ba26b1.png)

asmsuechan:
I'll improve it.

asmsuechan:
Fixed.

kazup01:
🎉

Issue #762, opened 2017-08-03
My storage fails to update the quantity of notes after deleting. 


wonderlives:
Please paste some **screenshots** with opening the developer tool if you report a bug.
<img width="497" alt="screen shot 2017-08-03 at 6 10 31 pm" src="https://user-images.githubusercontent.com/25517207/28946044-0bf2798c-7877-11e7-9b0f-4a8740bd018b.png">
After delete the notes, the number on the sidebar fails to update.

asmsuechan:
Hi, @wonderlives. Please paste some **screenshots** with opening the **developer tool** if you report a bug.

MrBMT:
@wonderlives Have you gone in to the Trash folder and fully deleted the notes from there?

At the moment the number of notes shown under My Storage doesn't update until they have been completely removed from the Trash, as mentioned in #757

asmsuechan:
@MrBMT Thank you for your support. I hope this (https://github.com/BoostIO/Boostnote/pull/764) helps.
And I close this because of duplication.

PR #764, opened 2017-08-05
Fix noteCount on a note trased

asmsuechan:
# context
This is an improvement of `Trash`.

# before
A note counter didn't decrement when the note was trashed.

# after
It works.
c1a8eabbc5f775dea412f01636150ad9

# ref
https://github.com/BoostIO/Boostnote/issues/757

# note
I assume this logic is not a better way.


wonderlives:
Awesome, the problem #762 I reported got fixed! Thank you! Really awesome work!

asmsuechan:
@wonderlives You're welcome! Have fun!

Issue #750, opened 2017-07-31
[feature request] show todo progress in list


XGHeaven:
I hope todo progress can show in list.
![image](https://user-images.githubusercontent.com/9291212/28785162-267d8624-7648-11e7-8e16-0c6588b11564.png)
There are my four todo notes, todo progress should shown in red rectangle or other place I hope.

And I very like this app, I hope you can distribute cloud service quickly.
There is no Chinese version of contributing.md. I can translate to Chinese.
And any other Chinese language needed, I can help you.

asmsuechan:
I am willing to agree to your request.

> I hope you can distribute cloud service quickly

Sorry, we don't have any plans to develop it.

> There is no Chinese version of contributing.md. I can translate to Chinese.

Yes, please! I'm really looking forward to seeing your PR indeed.

asmsuechan:
@XGHeaven Really thank you so much for your help! And this feature was released in v0.8.4! :smile:

PR #776, opened 2017-08-09
add todo percentage in item list

XGHeaven:
#750 

And I implement it like github todo list.

Before:
qq20170810-123435

After:
qq20170810-123357
qq20170810-123338

Color as same as TodoListPercentage.
Todo style is like github todo style. And todo list doesn't show in folded mode.

Added:
* browser/lib/getTodoState.js get percentage of complete todo helper function
* browser/compoents/TodoProcess.{js/styl}

### DIFF
--- a/browser/lib/getTodoState.js
+++ b/browser/lib/getTodoState.js
@@ -1,0 +1,25 @@
+export function getTodoState (content) {
+  let splitted = content.split('\n')
+  let numberOfTodo = 0
+  let numberOfCompletedTodo = 0
+
+  splitted.forEach((line) => {
+    let trimmedLine = line.trim()
+    if (trimmedLine.match(/^[\+\-\*] \[\s|x\] ./)) {
+      numberOfTodo++
+    }
+    if (trimmedLine.match(/^[\+\-\*] \[x\] ./)) {
+      numberOfCompletedTodo++
+    }
+  })
+
+  return {
+    total: numberOfTodo,
+    completed: numberOfCompletedTodo
+  }
+}
+
+export function getTodoPercentageOfCompleted (content) {
+  const state = getTodoState(content)
+  return Math.floor(state.completed / state.total * 100)
+}
--- a/browser/components/NoteItem.js
+++ b/browser/components/NoteItem.js
@@ -4,7 +4,7 @@
 import React, { PropTypes } from 'react'
 import { isArray } from 'lodash'
 import CSSModules from 'browser/lib/CSSModules'
-import { getTodoState } from 'browser/lib/getTodoState'
+import { getTodoStatus } from 'browser/lib/getTodoStatus'
 import styles from './NoteItem.styl'
 import TodoProcess from './TodoProcess'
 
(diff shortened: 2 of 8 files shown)


kazup01:
Hi @XGHeaven , thanks for your PR!
How about like GitHub todo list? I think that this UI is clearer.

screen shot 0029-08-09 at 11 54 05 am

XGHeaven:
@BoostnoteBot Why need to change `browser/main/Detail/SnippetNoteDetail.js`?
The todo percentage only show when note type is `MARKDOWN_NOTE`, snippet note haven't todo percentage.

asmsuechan:
@XGHeaven No worry, it's a bot. So it replays automatically when `MarkdownNoteDetail` is detected.

asmsuechan:
I don't think it should be shown in folded note list.

browser/lib/getPercentage.js
```diff
@@ -0,0 +1,17 @@
+export default function getPercentage(content) {
```
asmsuechan: Great!

XGHeaven:
@kazup01 I think this is a good idea. 
@asmsuechan What are you think about that? use github todo list?
If percentage don't shown in folded list, user want to see todo percentage only switch to unfolded list. It's a little trouble.

asmsuechan:
I think you don't have to implement the GitHub like todo list in this time. Perhaps I'll implement it later on.

I want the folded list to be simple. And come to think of the updated time of a note, it's not be shown on folded list.

XGHeaven:
I think github todo list looks like not bad.
image
If you implement it, that's good.

asmsuechan:
> I think github todo list looks like not bad.

I think so, but I have another idea the reason why I mentioned `you don't have to implement the GitHub like todo list in this time`.
I also want to put the numbers of todo percentage in `MarkdownNoteDetail`. And it's not only this item list's problem. So I want to separate them into other PRs.

asmsuechan:
Ah, of course, you can do it, please do it.

XGHeaven:
Ok

XGHeaven:
@asmsuechan I implement it. Can you review it?

browser/lib/getTodoState.js
```diff
@@ -0,0 +1,25 @@
+export function getTodoState (content) {
```
asmsuechan: I prefer `getTodoState` than `getTodoStatus`.

XGHeaven: How do you think about `getTodoState.js` file name? Should change to `getTodoStatus.js`?

asmsuechan: I want to change both of them.

XGHeaven: I changed that. And I also changed TodoProcess prop from `todoState` to `todoStatus`.
By the way, add transition for TodoProcess bar.

asmsuechan:
@XGHeaven  Well done! Thank you so much indeed :blush: I'll merge this later.

XGHeaven:
😄

PR #787, opened 2017-08-10
Fix configmanager

asmsuechan:
# context
It's a bug happens on v0.8.13.

# before
The settings of Boostnote was reset and used DEFAULT_CONFIG unless `~/.boostnoterc` was not created and set settings.

# after
Boostnote reads settings from localStorage and overwrites the settings when `~/.boostnoterc` exists.

### DIFF
--- a/browser/main/lib/ConfigManager.js
+++ b/browser/main/lib/ConfigManager.js
@@ -60,19 +60,18 @@ function _save (config) {
 }
 
 function get () {
-  let config = window.localStorage.getItem('config')
+  const storedConfig = window.localStorage.getItem('config')
+  let config = storedConfig
 
   try {
     const boostnotercConfig = RcParser.parse()
 
-    config = Object.assign({}, DEFAULT_CONFIG, JSON.parse(config))
-
-    config = Object.assign({}, DEFAULT_CONFIG, boostnotercConfig)
-    config = assignConfigValues(config, boostnotercConfig, config)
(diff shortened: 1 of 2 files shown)


browser/main/lib/ConfigManager.js
```diff
@@ -60,19 +60,17 @@ function _save (config) {
...
+  const rawStoredConfig = window.localStorage.getItem('config')
+  const storedConfig = Object.assign({}, DEFAULT_CONFIG, JSON.parse(rawStoredConfig))
+  let config = storedConfig
 
   try {
     const boostnotercConfig = RcParser.parse()
-
-    config = Object.assign({}, DEFAULT_CONFIG, JSON.parse(config))
-
-    config = Object.assign({}, DEFAULT_CONFIG, boostnotercConfig)
```
asmsuechan: This line is the reason why the bug happened.

vvs:
Is this really desired behavior when the `.boostnoterc` settings always override the user settings set in the UI?

So, a user modifies something in the UI, everything works as he/she likes, then he/she restarts the app, and the settings is gone, replaced by the one from the `.boostnote` file.

Or is this `.boostnoterc` file not really designed to be used by users, and only be used to fix/reset some of the user settings (for example, to be able to reset the settings if they gone bad for some reason)?

asmsuechan:
> Is this really desired behavior when the .boostnoterc settings always override the user settings set in the UI?

Yes, it is. But I'll reconsider it.

PR #793, opened 2017-08-12
Add file drop on NoteList

asmsuechan:
# context
Drag&drop on NoteList.

# before
Nothing.

# after
439d0e3727951b609025ae4bf92de217

# For test
* Drag&drop 1 .md file
* Drag&drop 2 .md files
* Drag&drop 1 .md and 1 .txt file
* Drag&drop 1 .html file
* Drag&drop .pdf file (not work)
* Drag&drop on trash
* Not working if the storage is empty

### DIFF
--- a/browser/main/NoteList/index.js
+++ b/browser/main/NoteList/index.js
@@ -348,6 +348,21 @@ class NoteList extends React.Component {
       properties: ['openFile', 'multiSelections']
     }
 
+    dialog.showOpenDialog(remote.getCurrentWindow(), options, (filepaths) => {
+      this.addNotes(filepaths)
+    })
+  }
+
+  handleDrop (e) {
+    e.preventDefault()
+    const filepaths = Array.from(e.dataTransfer.files).map(file => { return file.path })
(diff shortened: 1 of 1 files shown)


yashha:
Maybe also txt files?
We could also think about html files, but thats uncommon :) https://domchristie.github.io/to-markdown/

asmsuechan:
Hi, @yashha. Thank you for your proposal. Yes, it can. Actually .html file also work. But it is not converted to markdown style for now. And I have to implement validator because there are no restrictions on the type of file extension.
c0177f090eb70c0a1128102e015e114e

TechIsCool:
Do they get imported or are they linked to the files when dropped on the application. The reason I ask is because I think it would be awesome to have all my runbooks and README.md files from repositories linked inside of boost notes. It would allow editing them and then not having to copy and paste back into the repos.

yashha:
Good Idea @TechIsCool, would be interesting to have a way to access linked MD-Files.

kazup01:
Merged

PR #804, opened 2017-08-15
All shortcut configuarable

asmsuechan:
# context
I separated ConfigManager into ConfigManager and KeysManager to change shortcuts in menu into configurable.

# before
Users cannot set their own shortcuts.

# after
image

# for tests
* All default shortcuts are shown when user looks at it first time
* All of them can be changed after restart
* Enter blank and save
* Change shortcuts twice or more
* Work even if `~/Library/Application\ Support/Electron/config.json` doesn't exist
* Load from localStorage if the config of shortcuts doesn't exit
* .boostnoterc works
* Apply to localStorage.config if hotkey is updated
* Work even in the environment of first-time users

# note
The reason why I use `electron-config` is that because Boostnote has to read the shortcuts from `lib/main-menu.js`. Using CodeManager as well as Hotkey doesn't work because of webpack.
The data is located in `~/Library/Application\ Support/Electron/config.json`
In the current state, hotkey in `~/.boostnoterc` is ignored. In order to deal it, need to save the contents of `~/.boostnoterc` into localStorage.
I should add tests for KeysManager
Should KeysManager have attributes? I mean I thought it's better to make it class.

# For compatibility


yashha:
Thanks for the enhancement.
But why you do don't make one tab with "Hotkey" and "Shortcuts" merged?
Maybe with subsections?
I don't see a difference.
Also would a Key record feature be awesome. Maybe a red circle on the right of the input field or a Button next to the inputfield to record.

asmsuechan:
@yashha For ease, I call `shortcut` both `Shortcuts` and `Hotkey` this time. If you change a shortcut in `Hotkey`, you don't need to restart Boostnote. But if `Shortcut`, you have to restart Boostnote. It's a big difference between them. That's why I separated them. And also the structures are vastly different.

But well, as you can see, this is WIP PR. So there are a bunch of chances to merge them if I find any good way.

asmsuechan:
@yashha 
> Also would a Key record feature be awesome. Maybe a red circle on the right of the input field or a Button next to the inputfield to record.

Could you give me an example of a service which this feature is implemented? Actually I'm not sure what exactly it is.

yashha:
I think one tab would be better for UX. You could write down a note, that Boostnote has to be restarted.

I thought about the keyboard feature of ubuntu for example.
screenshot from 2017-08-15 11-14-48
There you click on the box and it gos into record mode, when you hit "esc" it stopps, when you hit "enter" it accepts.

yashha:
I haven't seen a electron app yet which implemented that though.

asmsuechan:
@yashha Merged them.

browser/main/lib/MenuManager.js
```diff
@@ -0,0 +1,35 @@
+const OSX = global.process.platform === 'darwin'
```
asmsuechan: The reason why I added `MenuManager` instead of `ShortcutManager` is that because of a preparation of Plugin feature. I plan to be able to add to the menu.

asmsuechan: Umm, I'm worrying it.

asmsuechan: Yes. I'll create `KeysManager` and transplant Hotkeys into it and will delete `MenuManager` since the day when I really need it.

browser/main/lib/KeysManager.js
```diff
@@ -0,0 +1,43 @@
...
+
+const DEFAULT_SHORTCUTS = {
+  newNote: 'CommandOrControl+N',
+  focusNote: 'Control+E',
+  deleteNote: OSX? 'Control+Backspace' : 'Control+Delete',
+  print: 'CommandOrControl+P',
+  nextNote: 'Control+J',
+  previousNote: 'Control+K',
+  focusSearch: 'Control+S',
+  hotkey: {
```
asmsuechan: I made this nested because of compatibility.

kazup01:
Hi @asmsuechan , how about this current state?

kazup01:
I close it, but please feel free to reopen. Thanks!

Issue #778, opened 2017-08-09
Feature request: Complex table markdown syntax


RedBug312:
I think it would be better to have extended table syntax like [MultiMarkdown](http://fletcher.github.io/MultiMarkdown-4/tables), or at least just `colspan` feature.
It's commonly used for agenda tables, and I think it troublesome to write them in HTML just for `colspan`.

BTW, there seemed no related plugins in markdown-it yet.

asmsuechan:
Hi, @RedBug312. Thank you for your suggestion. I'll consider it when a related plugin of markdown-it is released.

PR #811, opened 2017-08-18
Support for extended table markdown syntax

RedBug312:
Related to the issue #778 I posed.
I've written a markdown-it plugin for that, and here below is the demo:
demo


kazup01:
Thanks @RedBug312 !

PR #818, opened 2017-08-19
Make pasting an image in CodeEditor work

asmsuechan:
# context
I made pasting an image work.

# before
Nothing happens when I paste an image.

# after
I can paste an image from clipboard.
9d989912ab0c470243b2bbfefe78ef2f

# For tests
* does not work when I paste .pdf or something that are not images.
* Paste .png on macOS
* Paste .jpg on macOS
* Paste .png on windows
* Paste .jpg on windows
* Paste .jpg on Linux
* Paste .png on Linux

### DIFF
--- a/browser/components/CodeEditor.js
+++ b/browser/components/CodeEditor.js
@@ -3,6 +3,7 @@ import _ from 'lodash'
 import CodeMirror from 'codemirror'
 import path from 'path'
 import copyImage from 'browser/main/lib/dataApi/copyImage'
+import { findStorage } from 'browser/lib/findStorage'
 
 CodeMirror.modeURL = '../node_modules/codemirror/mode/%N/%N.js'
 
@@ -39,6 +40,7 @@ export default class CodeEditor extends React.Component {
       }
       this.props.onBlur != null && this.props.onBlur(e)
     }
+    this.pasteHandler = (editor, e) => this.handlePaste(editor, e).bind(this)
     this.loadStyleHandler = (e) => {
       this.editor.refresh()
     }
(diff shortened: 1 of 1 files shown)


asmsuechan:
This feature works even if the image is jpg or something that are not png.
I confirm it only on macOS Sierra 10.12

browser/components/CodeEditor.js
```diff
@@ -204,6 +208,29 @@ export default class CodeEditor extends React.Component {
...
+    const dataTransferItem = e.clipboardData.items[0]
+    if (!dataTransferItem.type.match('image')) return
+
+    const blob = dataTransferItem.getAsFile()
+    let reader = new FileReader()
+    let base64data
+
+    reader.readAsDataURL(blob)
+    reader.onloadend = () => {
+      base64data = reader.result.replace(/^data:image\/png;base64,/, '')
```
asmsuechan: The meaning of this line
https://stackoverflow.com/questions/11511511/how-to-save-a-png-image-server-side-from-a-base64-data-string#comment36092057_11511552

asmsuechan: I'm not sure the reason why the image is copied as png even if it's jpg. So this line works fine even if it's jpg.

asmsuechan: Here's the jpg file I pasted from the clipboard.
image

jlsjonas:
This feature seems to be broken...

It does seem to (try?) storing the images; but the resulting path doesn't actually render them
image

image


(v0.8.15)

xanderberkein:
Can confirm, got the same problem.
I'm on Boostnote 0.8.15, OS X 10.13

garulovilla:
I think it's important to check if the images are in use and remove those that are no longer necessary.

PR #877, opened 2017-09-20
Add RealtimeNotification

asmsuechan:
# context
We want users to notify tips.

# before

# after
image

# note
I'm not sure how I describe plural notifications, so I managed to show the first notification of them `notifications[0]`

# For tests
* it should not fail the application even if the notification repository does not exist
* it should show a notification


kazup01:
🎉

PR #879, opened 2017-09-20
Make line numbers in the editor optional

xxdavid:
Add an option to disable the line numbers in the Markdown editor.


kazup01:
Thanks @xxdavid ! We will check it!

asmsuechan:
image

browser/components/CodeEditor.js
```diff
@@ -50,7 +50,7 @@ export default class CodeEditor extends React.Component {
     this.value = this.props.value
     this.editor = CodeMirror(this.refs.root, {
       value: this.props.value,
-      lineNumbers: true,
+      lineNumbers: this.props.lineNumber,
```
asmsuechan: I prefer `lineNumbers` as the variable name.

browser/components/MarkdownEditor.js
```diff
@@ -213,6 +213,8 @@ class MarkdownEditor extends React.Component {
     if (!(editorFontSize > 0 && editorFontSize < 101)) editorFontSize = 14
     let editorIndentSize = parseInt(config.editor.indentSize, 10)
     if (!(editorFontSize > 0 && editorFontSize < 132)) editorIndentSize = 4
+    let editorLineNumber = config.editor.lineNumber
```
asmsuechan: I don't think the name `editorLineNumber` is appropriate because it's a boolean.

browser/components/CodeEditor.js
xxdavid: Yeah, I prefer `lineNumbers` too. I used `lineNumber` as it was already used somewhere else in the code.

browser/components/MarkdownEditor.js
```diff
@@ -213,6 +213,8 @@ class MarkdownEditor extends React.Component {
...
+    if (editorLineNumber === undefined) editorLineNumber = true
```
asmsuechan: `const editorLineNumber = config.editor.lineNumber || true` or something?

xxdavid: If `config.editor.lineNumber` were `false` (disabled), `config.editor.lineNumber || true` would be `true`. I don't think that's what you want.

browser/components/MarkdownEditor.js
xxdavid: Is 'editorShowLineNumbers` better? Or just `showLineNumbers`? Or what do you propose?

browser/components/MarkdownEditor.js
asmsuechan: What are cases that `config.editor.lineNumber` to be `undefined`? The setting should be `true` or `false`.

browser/main/Detail/SnippetNoteDetail.js
```diff
@@ -530,6 +530,7 @@ class SnippetNoteDetail extends React.Component {
             fontSize={editorFontSize}
             indentType={config.editor.indentType}
             indentSize={editorIndentSize}
+            lineNumber
```
asmsuechan: ?

browser/components/MarkdownEditor.js
xxdavid: Oh, sorry, of course, you are right. I used that before I declared the property in the `ConfigManager.js`.

browser/components/MarkdownEditor.js
asmsuechan: `isLineNumberShown`? Umm, I'm not sure it's better...

browser/main/Detail/SnippetNoteDetail.js
xxdavid: I thought Snippet notes should always have line numbers. Shouldn't they?
(`lineNumber` is equal to `lineNumber={true}`.)

browser/components/MarkdownEditor.js
asmsuechan: > If config.editor.lineNumber were false (disabled), config.editor.lineNumber || true would be true. I don't think that's what you want.

I got it.

browser/main/Detail/SnippetNoteDetail.js
asmsuechan: I think a little regarding the specification.

> (lineNumber is equal to lineNumber={true}.)

Understood.

browser/components/MarkdownEditor.js
xxdavid: I'm confused, sorry. The variable can't be undefined (as you said), can it?

browser/components/CodeEditor.js
asmsuechan: > it was already used somewhere else in the code.

Where is it? I could not find it... Does it mean you chose `lineNumber` because there are probabilities to conflict?

browser/components/MarkdownEditor.js
asmsuechan: Oh, I was confused, too lol

Forget about the ternary operation.

browser/components/CodeEditor.js
xxdavid: It's used for the line numbers in the preview. [Here](https://github.com/BoostIO/Boostnote/blob/95d74d1ca2206ed18add0348682af0063a5ac5b0/browser/main/lib/ConfigManager.js#L45), for example.

browser/components/MarkdownEditor.js
xxdavid: I don't think that fits the plural form (`lineNumbers`) you suggested...

browser/components/CodeEditor.js
asmsuechan: I see, and the name is also not good... I'll have to change it sooner or later.

browser/components/MarkdownEditor.js
asmsuechan: I often use `is` even if the var name is plural, and the prefix `is` describes it's a boolean value. `isLineNumbersShown`...?

xxdavid: OK then. I meant just the language (English) thing. Other than that, I think `isLineNumbersShown` is fine.

browser/components/CodeEditor.js
xxdavid: So should I change it to `lineNumbers`?

browser/main/Detail/SnippetNoteDetail.js
xxdavid: So do you want Snippet notes to respect this setting or should they always show the line numbers?

browser/components/MarkdownEditor.js
xxdavid: Or we can just use `config.editor.lineNumber` directly [here](https://github.com/xxdavid/Boostnote/blob/7a4258bb208d43be80e575fe61c0a0a187811e3f/browser/components/MarkdownEditor.js#L247), I think 😄 .

kazup01:
Hi @asmsuechan and @xxdavid , how is this state?

xxdavid:
@kazup01, I'm waiting for @asmsuechan to tell what we have agreed on and what to change.

browser/components/CodeEditor.js
```diff
@@ -141,6 +141,10 @@ export default class CodeEditor extends React.Component {
       this.editor.setOption('indentWithTabs', this.props.indentType !== 'space')
     }
 
+    if (prevProps.lineNumber !== this.props.lineNumber) {
+      this.editor.setOption('lineNumbers', this.props.lineNumber)
+    }
```
sota1235: I think you can set even if `prevProps.lineNumber !== this.props.lineNumber` is false.
Is it for performance?

sota1235:
@xxdavid Hi! I'm one of maintainers of Boostnote.
How's this PR status? I think you got some advices from @asmsuechan .
After you fix, plz tell me. Then I will check and merge!
Thank you

kazup01:
Hi @asmsuechan , we are waiting for your reply. Could you check @xxdavid's comments ?

browser/components/CodeEditor.js
xxdavid: Yes, it is and as you can see, the checks are performed for all the options in the [componentDidUpdate](https://github.com/xxdavid/Boostnote/blob/7a4258bb208d43be80e575fe61c0a0a187811e3f/browser/components/CodeEditor.js#L117) method.

xxdavid:
Sorry for still not updating the PR, but I'm unsure about the _variable naming_ (`lineNumbers`/`lineNumbers`) and whether should snippet notes _respect this option_ (and don't show the line numbers if it is disabled) or show it always.

sferra:
@xxdavid just in case you are still pondering on the variable naming: maybe consider the name `displayLineNumbers`. 

i would also suggest handling the snippets separately (as in: new ticket/discussion, new pull request). that way this pull request could be closed while still keeping the discussion/issue alive.

xxdavid:
@sferra, yeah, I agree, this suits the purpose best in my opinion.
I'll change all the variables to `displayLineNumbers`.

xxdavid:
I've pushed the changes. I also changed snippet notes to respect the option.
Could it be merged now, please?
And merry Christmas to you all! :)

Rokt33r:
@xxdavid 
Sorry for being late. I'll review this on this weekend and release within the next week.

PR #882, opened 2017-09-22
Fix layout

kazup01:


kazup01:
Fix the layout like below images
screen shot 0029-09-22 at 4 07 10 pm
screen shot 0029-09-22 at 4 07 17 pm

kazup01:
Merged

PR #884, opened 2017-09-22
Fix layout at bottom

kazup01:


kazup01:
Fixed like below images.
screen shot 0029-09-23 at 12 03 47 am
screen shot 0029-09-23 at 12 03 56 am

kazup01:
Merged

Issue #832, opened 2017-08-28
When in vim mode, typing :w<Enter> means nothing.


KuangLei:
I hope that in vim mode, `:w<Enter>`, `:q<Enter>`, `:wq<Enter>` should have the same function as pressing `Ctrl-s`.  

Thanks.

asmsuechan:
Hi, @KuangLei. Boostnote doesn't have `save` function. It saves a note automatically when you edit it. So I don't feel like I implement the features.

KuangLei:
Now `:w<Enter>` returns nothing, and `:q<Enter>` or `:wq<Enter>` returns "Not an editor command".  

So, I hope `:w<Enter>`, `:q<Enter>`, `:wq<Enter>` could have the function of exiting edit mode and going back to view mode, just like pressing `Ctrl-s`.  

Thanks.

KuangLei:
#883

asmsuechan:
https://github.com/BoostIO/Boostnote/pull/885

PR #885, opened 2017-09-22
iss #832 Make :q work in vim mode

asmsuechan:
# context
We want to exit from the editor in Vim mode.

# before
02a6570e0dd85231d3120aa861a95a83

# after
f76f1b74be813ee10f631b1cd41486d0

# For test
* :q exits from editor mode
* :q! exits from editor mode
* :qw exits from editor mode

### DIFF
--- a/browser/components/CodeEditor.js
+++ b/browser/components/CodeEditor.js
@@ -104,6 +104,12 @@ export default class CodeEditor extends React.Component {
 
     let editorTheme = document.getElementById('editorTheme')
     editorTheme.addEventListener('load', this.loadStyleHandler)
+
+    CodeMirror.Vim.defineEx('quit', 'q', this.quitEditor)
+  }
+
+  quitEditor () {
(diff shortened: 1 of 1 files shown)


kazup01:
awesome

KuangLei:
May I ask if `:wq` (Save and Exit) works either?  
I know it don't need to be saved manually but saving is a habit for people like me.  

Thanks.

asmsuechan:
Sadly it doesn't work because of https://github.com/codemirror/CodeMirror/blob/25e12b5fe1ee7b0be7e7ee4306608caa4e93c6d6/keymap/vim.js#L701.
It means I can define my own command by only 1 key.

KuangLei:
Thanks very much anyway.

KuangLei:
(It should be `:wq` instead `:qw`.)

asmsuechan:
Exactly :laughing:

PR #915, opened 2017-09-27
change grand to grunt

cofcool:


kazup01:
Wow I didn't notice this mistake. Thanks @cofcool !

PR #926, opened 2017-09-30
[feature][in-progress] Watch files for external changes

Overload119:
This is the approach I'll take with file monitoring.
It relies on my previous change for using prettifier.
It also adds chokidar which is recommended for file watching. This results in a `warning` during build, but the package works fine (https://github.com/paulmillr/chokidar/issues/77)

The goal is to fix the issue of external changes to a file while Boostnote is open.

There's a couple things that we should resolve with this feature.

- When I have Boostnote open on 2 computers, and synced through Dropbox, if I edit the file on one computer, Boostnote should let me know the backing storage file was changed. I can choose to pick up or ignore the changes, similar to how Atom works.
- If I use another editor (such as Sublime Text) to edit the `.cson` file Boostnote should let me know the file is changed on the storage layer.
- When viewing a note while Boostnote is open, we should lookup the actual file instead of the in-memory representation from Boostnote.

Feedback is encouraged!


kazup01:
Since there are many conflicts, I close this pull request. Thank you for your suggestion.

PR #956, opened 2017-10-11
Pin to top

asmsuechan:
# context
`Pin to top` feature is implemented.
6c5efcd90b5a0cf5b8d18a8ef1259eb5

# before
This is a new feature so there's no `before`.

# after
I'll write this later.

# for tests
* can add pin (snippet/markdown)
* can remove pin (snippet/markdown)
* not to be shown the context box on /home, /starred or /trash
* not to be shown pinned icons on /home, /starred or /trash

### DIFF
--- a/browser/components/NoteItem.js
+++ b/browser/components/NoteItem.js
@@ -44,10 +44,8 @@ const TagElementList = (tags) => {
  * @param {Function} handleNoteContextMenu
  * @param {Function} handleDragStart
  * @param {string} dateDisplay
- * @param {Function} handleNoteContextMenu
- * @param {string} pathname
  */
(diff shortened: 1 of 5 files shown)


KuangLei:
Seeing the picture above, I think the pin icon should be brighter, and I think a white pin to use in dark themes is neccessary too.  

Thanks.

asmsuechan:
I'm not in charge of style. That's why I ignore the ugly icon in spite of knowing that the style is not good. However, thank you for your comment.

kazup01:
Great contribution @asmsuechan 🎉
 I've been waiting for this feature. I will fix UI after merge it.

kazup01:
This feature has just released at v0.8.16! Thanks @asmsuechan  🎉
https://github.com/BoostIO/Boostnote/pull/1027

PR #961, opened 2017-10-12
Fix to create imageDir if it does not exist

asmsuechan:
# context
This PR fixes pasting the image from clipboard.
The reason why the issue happened is that because there's no image directory on the storage.

ref: https://github.com/BoostIO/Boostnote/pull/818


kazup01:
Merged

PR #968, opened 2017-10-12
Fix: "Uncaught TypeError: Cannot read property 'className' of null"

adrichey:
This fixes a bug where elements without a target offsetParent were throwing uncaught errors when trying to determine if a click event originated from an info panel element or not.

### DIFF
--- a/browser/main/index.js
+++ b/browser/main/index.js
@@ -27,7 +27,10 @@ document.addEventListener('click', function (e) {
   const className = e.target.className
   if (!className && typeof (className) !== 'string') return
   const isInfoButton = className.includes('infoButton')
-  const isInfoPanel = e.target.offsetParent.className.includes('infoPanel')
+  let isInfoPanel = false
+  if (e.target.offsetParent !== null) {
+    isInfoPanel = e.target.offsetParent.className.includes('infoPanel')
+  }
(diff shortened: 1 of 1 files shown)


browser/main/index.js
```diff
@@ -27,7 +27,10 @@ document.addEventListener('click', function (e) {
   const className = e.target.className
   if (!className && typeof (className) !== 'string') return
   const isInfoButton = className.includes('infoButton')
-  const isInfoPanel = e.target.offsetParent.className.includes('infoPanel')
+  let isInfoPanel = false
+  if (e.target.offsetParent !== null) {
+    isInfoPanel = e.target.offsetParent.className.includes('infoPanel')
+  }
```
asmsuechan: Cool! However, I prefer this way

```js
  const offsetParent = e.target.offsetParent
  const isInfoPanel = offsetParent !== null
    ? offsetParent.className.includes('infoPanel')
    : false
```

How about this?

adrichey: Sure. I will update that in a bit. The only reason I did it the way I did is for readability. Ternary statement it is!

asmsuechan:
Thank you so much!

kazup01:
Thanks @adrichey !

adrichey:
Anytime! Glad to assist!

Issue #571, opened 2017-05-14
[Feature] Github gist


kocoai:
Please add sync feature with github gist (like Lepton :P )

kazup01:
GitHub gist integration is sounds interesting!
Please also check #308 about cloud sync.

kazup01:
I will try Lepton! Thanks your info!

kazup01:
memo #507

enygma:
Looks like the API integration with the Gist API (https://developer.github.com/v3/gists/) and a token from the GitHub account wouldn't be too difficult. Though looking over the current features it doesn't look like there's any kind of API interface happening right now so it'd be something new.

I wouldn't think it would even need to be a sync (bi-directional) per-se. Even just a "publish" as a first step would be cool. If I were a better JS developer I'd happily help out on this one.

sanderman123:
I think this feature can grow the user base quite substantially because GistBox just announced they close as of December 1st 2017. 
https://app.gistboxapp.com
Their new app - "Cacher" - is so expensive they should have called it "Casher" instead.

It's at least why I ended up here... I would definitely move to Boostnote if I can access my old Gists.

@kocoai thanks, I'll have a look at Lepton as well.

phylor:
I'm looking into it by creating a prototype. A token based authentication seems to be doable. Will keep you posted 😄

skittlyoaten:
Github Gists sync would be great.  IMO, the best way to do that would be to make it so when a user goes  to "Add Storage", just allow for a "type" of gists.  That would keep it separate from other private, local, of cloud shared folders.

mrkeysh:
Agreed that a GIST would be best!
Using Atom.IO, there is a package called [Sync-settings](https://atom.io/packages/sync-settings) ( https://atom.io/packages/sync-settings ) that uses Gists.  This may be what you/we are looking for.  I like how this package also operates too with Git/Gists and it has an "autodetect" feature that it checks the date when you open it and prompts you for an update.

IssuehuntBot:
@kazup01 has boosted this issue with $60. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/571)

IssuehuntBot:
@yhaddad has started working. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/571)

IssuehuntBot:
[...]

Issue #455, opened 2017-04-19
Feature: Add "Send to gist" or upload to gist


Germanaz0:
It would be cool to add a button which gives the possibility to upload the note to gist.github.com

jasondavis:
I 2bd the vote.

Also export to CodePen.io API and JSFiddle API like some other code playground sites do

Germanaz0:
@jasondavis nice idea to export to jsfiddle

kazup01:
Hi @Germanaz0 san,
This feature is sounds interesting! Please also refer to #308

IssuehuntBot:
@kazup01 has boosted this issue with $40. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/455)

IssuehuntBot:
@jkgatt has started working. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/455)

IssuehuntBot:
@jkgatt has stopped working. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/455)

IssuehuntBot:
@amaseda has started working. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/455)

IssuehuntBot:
@amaseda has stopped working. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/455)

octavioamu:
was looking to start this one but gist can't be not longer anonymous, so for this we need github authentication  :/

IssuehuntBot:
@dharanee86 has started working. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/455)

IssuehuntBot:
@dharanee86 has stopped working. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/455)

IssuehuntBot:
@veegishx has started working. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/455)

IssuehuntBot:
@veegishx has stopped working. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/455)

IssuehuntBot:
@kazup01 cancelled funding, $40, of this issue. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/455)

IssuehuntBot:
[...]

PR #971, opened 2017-10-12
Sharing to Github Gist

phylor:
Hi there!

This is a work in progress pull request I'd love to hear your opinion on. It provides sharing to Gist (cf. https://github.com/BoostIO/Boostnote/issues/455 and https://github.com/BoostIO/Boostnote/issues/571). You need to create a Github API Token and enter it in preferences (new tab `Integrations`).

I added a button for markdown notes (top right, next to the info button). It directly shares it to Gist.

screen shot 2017-10-12 at 22 55 16
screen shot 2017-10-12 at 22 55 32

Thoughts I have:

- I imagine a popup (as when clicking the info button) instead of directly sending it to Gist. This could also host other sharing options (codepen, etc.) and sharing options (e.g. whether the Gist is public or not).
- Republishing should update the same gist instead of creating a new one
- It would also be nice to have the HTTPS link to the Gist visible in the share popup
- Snippets should be shareable as well

What do you think?

I'd love to continue on this if you think it goes in the right direction!


kazup01:
Thank you for your suggestion!
We think that this feature should be provided as plugin, not required at default.
There is none plugin feature yet, we will implement it later.

Your suggestion is so great. Thank you again.

PR #972, opened 2017-10-13
Replace pdf to print

asmsuechan:
# context
I replace `.pdf export` on InfoPanel to `Print` because I don't think it will be implemented soon and we can export as .pdf via `Print`.

image


kazup01:
Merged

PR #986, opened 2017-10-16
Snapcraft

asmsuechan:
# context
This change brings Boostnote to be released on `snap`. The release runs only master branch and node stable.


pat-s:
@asmsuechan I am currently reading into the snap building process. I am wondering why you merged this PR on Oct 16 implying that it works and then ask a [question](https://forum.snapcraft.io/t/how-can-i-contain-my-binary-in-prime-folder/2469) on the snap forums on Oct 17 stating that you cannot build it. 

When a user sees that a `snap` folder exists in the master branch, one assumes that there is an existing `snap` package and one also won't try to put work into it. So any plans on pushing this attempt further?

This is not meant to be offensive! I was  thinking that you had a working build until I took a deeper look at the travis build that was acutally passing but without doing anything.

PR #1010, opened 2017-10-24
Adjust notification HotKeyTab

PaulRosset:
1.  Adjust position of notification in HotKeyTab

2. Notification can disappear, to be more like a notification.
Would like to but break some test. For the moment, deleted feature in commit below.

boostnotepr


kazup01:
Merged. Thank you for your contribution @PaulRosset !

Issue #914, opened 2017-09-26
[Feature request] Checkboxes with color


MakisH:
I think it would be very helpful if the checkboxes changed color when checked. In a long todo list, I find it difficult to see the non-checked checkboxes (my open TODOs) and color would really help.

It would even be great if I could set these colors in the settings. Two hex color code fields would be enough.

What do you think?

kazup01:
Hi @MakisH , thanks for your suggestion! I think this is great. 
I also want to hear feedback from other people.

MrBMT:
Nice idea 👍 and if the colours were user configurable, if someone didn't like the change they could just set both checked and unchecked to use the same colour.

Nino9614:
And maybe not just change the color of the checkbox but also the text color or adding a background color to the text element. Then you dont need to try identifying on which line the checkbox is on.

piefel:
Colour is good, even if the colour is only grey for finished items.

It’s probably very difficult to colour the checkboxes in HTML, but coloured text should do the trick.

kazup01:
memo https://github.com/BoostIO/Boostnote/pull/1011

kazup01:
Hey guys, how about the current version? :) Thanks!

<img width="1165" alt="screen shot 0030-01-15 at 7 33 01 am" src="https://user-images.githubusercontent.com/8602615/34942646-6336d672-f9c6-11e7-9e64-30d9bd7fcae7.png">

MakisH:
Very nice! Is it already in v.0.8.20? I don't get colored checkboxes in neither of the White, Dark, and Dark-Solarized color themes.

PR #1011, opened 2017-10-24
Proof of concept: Feature checkbox styling

yashha:
peek 2017-10-25 01-57

#914


kazup01:
Hi @yashha , thanks for your suggestion and good for proof on concept.
How about this image?
todo-image

yashha:
Nice, I like your design.
But we should make it much faster. At the moment it is not really userfriendly to check the item.

kazup01:
Thanks for your feedback! I got it. How about like this?
howabout

MakisH:
I like the proposed design! Would it be super difficult to also get e.g. a bright green (or red, or configurable even through the config file) filling for the unfilled checkbox circles? :-D

kazup01:
Hi @MakisH , thank you for your feedback! Could you attach the example image?

MakisH:
@kazup01 this is what I mean:

32003455-0b4959d4-b9da-11e7-884f-3cfef2df88c6

This is only the idea, though. I am sure it can be improved!

yashha:
So any idea how to force an update of the preview?

kazup01:
I understand. Thank you :) @MakisH

kazup01:
Since I merged #1194, close this pull request.
This suggestion became very helpful and excited for our team :) Thanks again @yashha !

Issue #1009, opened 2017-10-24
Losing focus means leaving editing mode


piefel:
When I change windows back and forth, Boostnote editor will change from editing mode to previewing mode. This requires another click into the window or always pressing Ctrl-E before I enter more text.

(Using Ubuntu Linux.)

asmsuechan:
Hi, @piefel. Is your issue solved by https://github.com/BoostIO/Boostnote/pull/283? 
ref: https://github.com/BoostIO/Boostnote/issues/960#issuecomment-336595755

piefel:
Indeed, #960 is the exact same issue. Thanks for pointing that out!

I might add that I did not expect that functionality behind a _lock_ icon with the tooltip text _Lock_. OK, whatever icon, but the tooltip text could really be a bit more descriptive.

PR #1014, opened 2017-10-25
Change the tooltip text of the lock button for easy to understand what its role

asmsuechan:
# context
As @piefel mentioned in https://github.com/BoostIO/Boostnote/issues/1009#issuecomment-339235083, it's not easy to understand the role of the lock button. So I changed the tooltip text.

# before
image

image

# after
image

image


kazup01:
Thanks @asmsuechan !

asmsuechan:
I ignored the BoostBot's comment (https://github.com/BoostIO/Boostnote/pull/1014#issuecomment-339292318) because the lock button is not implemented in snippet notes.

PR #1016, opened 2017-10-25
ClearMessage notification HotKeyTab

PaulRosset:
1. Notification can disappear, to be more like a notification.

boostnotepr


kazup01:
Thanks @PaulRosset !

PR #1030, opened 2017-10-29
Fix custom preview fonts

RedBug312:
After upgrading to 0.8.16, I failed customize Preview Font Family. I found the CSS rule below failed due to escaped quotation marks, causing font-family inherited to another rule.

```css
body {
    font-family: 'Ubuntu','meiryo',''Microsoft YaHei'','helvetica','arial','sans-serif';
}
```

Removing the escaped quotation marks between `Microsoft YaHei` would solve the problem.
My environment is:
* OS: Ubuntu 17.04 x86_64
* node: v6.11.5

**Before**

**After**

with preview font set to `Ubuntu`


kazup01:
Merged. Thank you for your contribution @RedBug312 !

PR #1038, opened 2017-10-30
Fix enabling no focus right click pin to top and show adequate context menu

voidsatisfaction:
# WHY?

For user friendly movement in terms of pin to top interfaces

## BEFORE

pin-before

## AFTER

pin-after


kazup01:
That's soooooooooooooo good! Thank you for everyday support @voidsatisfaction 👍

voidsatisfaction:
@kazup01 nice to hear that :D

PR #1041, opened 2017-10-30
make ZZ work in vim mode

machellerogden:
As an avid vim user, I often use "ZZ" while in normal mode to quickly save and exit. This PR makes that bit of vim functionality work in Boostnote as well.


kazup01:
Thanks for your contribution @machellerogden ! This feature is looks good. 
Could you fix CI error?

machellerogden:
I'll check it out and update. Won't have time to loop back around to this until next week though. Thanks for your consideration!

kazup01:
Merged. I will fix CI error soon. Thank you for your contribution @machellerogden :)

PR #1042, opened 2017-10-30
only decode the url when the content of the link is a string

### LINKED ISSUE #1040: <a href="foo"> <img src="bar"/> </a> Syntax Broken
I'm not sure if this is an issue with Boostnote or codemirror, but as of Boostnote 0.8.16 `` tags within `<a>` tags are being removed. On Boostnote 0.8.15 these images were displayed and clickable as expected. [This JSFiddle](https://jsfiddle.net/wno1dnze/3/) is an example of how my clickable images used to look. Here is a gif showing the editor code, followed by the html page it generates when switching to preview:
img-355

Reason for needing this functionality:
I have a heavily customized version of Boostnote and it is all based around the ability to write html and js into notes to make clickable links to scripts (such as a script that adds expand/collapse sections to the current note). The ability to add to Boostnote's functionality without actually changing the app is a great feature and I hope to see clickable images fixed soon. 
Without the ability to make clickable images I couldn't create a personal note UI(once again great feature!) like below:
sectionednote

functionality shown: add section, reset title, link back to note that took you here(if linked), expand/collapse, expand/collapse all.

mslourens:
Fixes #1040

The cause of the problem is in ```MarkdownPreview.js```. The function ```fixDecodedURI``` only checks for innerText for every anchor element, but doesn't check if the anchor element contains other elements. In that case the uri is not decoded and the node remains as is.

### DIFF
--- a/browser/components/MarkdownPreview.js
+++ b/browser/components/MarkdownPreview.js
@@ -185,11 +185,13 @@ export default class MarkdownPreview extends React.Component {
   }
 
   fixDecodedURI (node) {
-    const { innerText, href } = node
+    if (node && node.children.length === 1 && typeof node.children[0] === 'string') {
+      const { innerText, href } = node
 
-    node.innerText = mdurl.decode(href) === innerText
-      ? href
-      : innerText
+      node.innerText = mdurl.decode(href) === innerText
+        ? href
+        : innerText
+    }
   }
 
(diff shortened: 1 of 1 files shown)


kazup01:
Merged. Thanks for your contribution! @mslourens 🎉

PR #1052, opened 2017-11-01
Add counter on deleted notes

PaulRosset:
* Add counter to see how many notes are in the trash.

boostnotecounterdeletednotess

### DIFF
--- a/browser/components/SideNavFilter.js
+++ b/browser/components/SideNavFilter.js
@@ -15,7 +15,8 @@ import styles from './SideNavFilter.styl'
  */
 const SideNavFilter = ({
   isFolded, isHomeActive, handleAllNotesButtonClick,
-  isStarredActive, handleStarredButtonClick, isTrashedActive, handleTrashedButtonClick, counterDelNote
+  isStarredActive, handleStarredButtonClick, isTrashedActive, handleTrashedButtonClick, counterDelNote,
+  counterTotalNote, counterStarredNote
 }) => (
   <div styleName={isFolded ? 'menu--folded' : 'menu'}>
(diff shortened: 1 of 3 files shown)


kazup01:
Hi @PaulRosset , thank you for always your contributing!

I think that "All Notes" and Starred" also have this feature better than "Trash" alone.  
How about do you think?

kazup01:
Hi @PaulRosset !
This(https://github.com/BoostIO/Boostnote/pull/1052/commits/875c4512212e4d04cc499c2305dfdf4f77d91388) is exactly what I meant.
Thank you for your support. We will check it soon!

sota1235:
@PaulRosset 
[Question] Why don't you show the number of each categories when the number is zero?

PaulRosset:
Hi @sota1235 

This is my personal point of view,  Less there is data on the UI more UI will fit for the user.
I prefer to use app which are pure.
So by adding always "0", it is an useless data for the user to know that.

kazup01:
Hi @PaulRosset ! Thank you for your reply.
We think that it is better to show "0" because it is displayed in the case of Folder. 
What do you think about it?
 
screen shot 0029-11-06 at 9 06 09 am

PaulRosset:
Oh ok, it make sense.
It is edited in the last commit.

kazup01:
Thank you for your quick response and great support :) We will check it!

kazup01:
Merged. Thank you for your contribution @PaulRosset 🎉

kazup01:
Hi @PaulRosset , I wrote your pull request to [release note](https://medium.com/boostnote/thanks-to-your-contribution-boostnote-v0-8-17-is-out-5fea9c7727f2).
Thank you again :)

PR #1064, opened 2017-11-04
enable importing file to empty folder

### LINKED ISSUE #1057: Cannot import into a specific folder
Video of problem (problem workflow first, then working workflow): http://www.youtube.com/watch?v=3kf_k5wTzNQ

Repro steps:

1. Select a folder (one of the storage options)
2. File > Import From > *select a file*
3. Nothing happens

**Here is a *working* workflow for import:**

1. Select the "All Notes" category
2. File > Import From > *select a file*
3. File successfully imports

------

System Info:

Windows 10.0.16299 (Windows 10 Pro Build 16299 x64)
Boostnote: 0.8.16? **(would be helpful to have an "About" under "Help" menu to see which version I'm using)**

yosmoc:
fix #1057

Current implementation is fetching the import destination storageKey and folderkey from a file in the folder. If no files are in the folder, it cannot fetch these keys. Use prop.params.storageKey / folderKey instead.

### DIFF
--- a/browser/main/NoteList/index.js
+++ b/browser/main/NoteList/index.js
@@ -434,10 +434,8 @@ class NoteList extends React.Component {
   addNotesFromFiles (filepaths) {
     const { dispatch, location } = this.props
 
-    const targetIndex = this.getTargetIndex()
-
-    const storageKey = this.notes[targetIndex].storage
-    const folderKey = this.notes[targetIndex].folder
+    const storageKey = this.props.params.storageKey
+    const folderKey = this.props.params.folderKey
 
     if (filepaths === undefined) return
     filepaths.forEach((filepath) => {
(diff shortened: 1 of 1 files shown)


yosmoc:
Sorry I closed this PR by mistake. I reopen again.

browser/main/NoteList/index.js
```diff
@@ -434,10 +434,8 @@ class NoteList extends React.Component {
   addNotesFromFiles (filepaths) {
     const { dispatch, location } = this.props
 
-    const targetIndex = this.getTargetIndex()
-
-    const storageKey = this.notes[targetIndex].storage
-    const folderKey = this.notes[targetIndex].folder
+    const storageKey = this.props.params.storageKey
+    const folderKey = this.props.params.folderKey
```
asmsuechan: I guess this is better.

```
const { dispatch, location, params } = this.props
const { storageKey, folderKey } = params
```

yosmoc:
Thanks for reviewing. I use the same storage/folder resolve way as creating a new note. I think this makes less confusing for user.

Would you please review it again?

PR #1081, opened 2017-11-06
solarized light/dark bug fix

yosmoc:
fixup commit for 0bf7e8b705793f432545c204749526d457547552
Solarized light/dark shares the same css file 'solarized.css'. Split by a space and use first one, since "solarized light" and "solarized dark" are the only theme which has a space in the theme name.

### DIFF
--- a/browser/components/MarkdownPreview.js
+++ b/browser/components/MarkdownPreview.js
@@ -259,7 +259,7 @@ export default class MarkdownPreview extends React.Component {
       ? theme
       : 'elegant'
     this.getWindow().document.getElementById('codeTheme').href = theme.startsWith('solarized')
-      ? `${appPath}/node_modules/codemirror/theme/${theme.split(' ')[0]}.css`
+      ? `${appPath}/node_modules/codemirror/theme/solarized.css`
       : `${appPath}/node_modules/codemirror/theme/${theme}.css`
   }
 
--- a/browser/components/MarkdownPreview.js
+++ b/browser/components/MarkdownPreview.js
@@ -258,7 +258,9 @@ export default class MarkdownPreview extends React.Component {
     theme = consts.THEMES.some((_theme) => _theme === theme) && theme !== 'default'
       ? theme
       : 'elegant'
-    this.getWindow().document.getElementById('codeTheme').href = `${appPath}/node_modules/codemirror/theme/${theme.split(' ')[0]}.css`
+    this.getWindow().document.getElementById('codeTheme').href = theme.startsWith('solarized')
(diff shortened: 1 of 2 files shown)


kazup01:
Thank you for your contribution! We will check it.

browser/main/lib/ConfigManager.js
```diff
@@ -122,7 +122,7 @@ function set (updates) {
     : 'default'
 
   if (newTheme !== 'default') {
-    editorTheme.setAttribute('href', '../node_modules/codemirror/theme/' + newTheme + '.css')
+    editorTheme.setAttribute('href', '../node_modules/codemirror/theme/' + newTheme.split(' ')[0] + '.css')
```
sota1235: This fixing is just only for solarized theme.
Can you add checking that `newTheme` is solarized or not?

yosmoc: Yes, this change is for solarized theme only. Currently, solarized dark/light is the only the theme which contains space, so this change will not cause the problem for other theme.

However, as you pointed out, if I add the `newTheme` is solarized or not, this makes more clear the intention of code.

Following code also doesn't check the theme, so I will also update this.
https://github.com/BoostIO/Boostnote/blob/993cb9cb0b7c72b13b1a1cba5ed26c7a29fbcdb7/browser/components/MarkdownPreview.js#L261

Thanks for reviewing!

browser/components/MarkdownPreview.js
```diff
@@ -258,7 +258,9 @@ export default class MarkdownPreview extends React.Component {
     theme = consts.THEMES.some((_theme) => _theme === theme) && theme !== 'default'
       ? theme
       : 'elegant'
-    this.getWindow().document.getElementById('codeTheme').href = `${appPath}/node_modules/codemirror/theme/${theme.split(' ')[0]}.css`
+    this.getWindow().document.getElementById('codeTheme').href = theme.startsWith('solarized')
+      ? `${appPath}/node_modules/codemirror/theme/${theme.split(' ')[0]}.css`
```
levrik: How about using `theme.substr(0, 9);` instead of splitting? Looks a bit clearer to me.

yosmoc: Thanks for suggestion.

Personally, I cannot see significant differences between your suggestion and original code from readable perspective. However now this code is only called for solarized dark/light theme, so I think
`${appPath}/node_modules/codemirror/theme/solarized.css`
is the most clear code. What do you think?

levrik: @yosmoc Oh.. yes. Completely overlooked that.

kazup01:
Merged. Thank you for your contribution @yosmoc :)

Issue #1090, opened 2017-11-09
Cannot read text when "Adding Storage" on Ubuntu


karanrajpal14:
Hello, 
I'm currently running Ubuntu Budgie and I'm unable to read any of the text. They all appear as boxes. Am I missing a particular font you guys use or something of that sort?
![boostnote](https://user-images.githubusercontent.com/3253145/32590917-e47dac78-c542-11e7-81c7-535d6e362e5c.png)

kazup01:
memo: https://github.com/BoostIO/Boostnote/issues/964

kazup01:
Hi @karanrajpal14 , @yosmoc has solved this problem. Could you try it?#1105

karanrajpal14:
Sure. Just give me a couple of hours and I'll try it once I'm back from work. Will report the status soon.

kazup01:
Thank you @karanrajpal14 :)

karanrajpal14:
@kazup01 @yosmoc how exactly do I upgrade electron for boostnote? If I'm not wrong, it needs to be updated within boostnote, correct? Or is there something I can do on my end?

yosmoc:
Hi @karanrajpal14,

If you can build my PR branch by yourself, you will get the boostnote with electron 1.6.15. You can find instruction in [here](https://github.com/BoostIO/Boostnote/blob/master/docs/build.md).

However, it's bit complicated, so I build for linux-x64 system. You can download from [here](https://send.firefox.com/download/4b47987d80/#IV9_wZrOD_8vxi-_OAICpA).
**Caution: This file link will not valid after 1 download or in 24 hours.**

Yesterday, I reproduced your problem and I confirmed it has been fixed by my PR #1105.

Enjoy!

karanrajpal14:
@yosmoc just downloaded your build and it worked. Thank you for doing that! Also, thank you for fixing the issue as well. I'll be closing this now since your fix will be merged eventually :)

PR #1105, opened 2017-11-11
electron-prebuilt is obsolete. use latest electron instead

yosmoc:
Today I investigate the #1090 problem, but unfortunately, I couldn't reproduce the problem.

According to this thread https://github.com/electron/electron/issues/9616, it has already been fixed in electron 1.6.x. So, I update electron to latest stable version.
And electron-prebuilt is obsolete. So I use latest electron instead. 

I have built in Ubuntu 17.10 in my Virtualbox and it seems working fine.
Would some one try this PR?

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "dev-start": "concurrently --kill-others \"npm run webpack\" \"npm run hot\""
   },
   "config": {
-    "electron-version": "1.2.8"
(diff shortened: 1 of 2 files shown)


kazup01:
Thank you @yosmoc ! I will ask others it, wait a moment :)

PR #1126, opened 2017-11-15
Add simple addon for CodeMirror

yamash723:
Hi. Thanks for Boostnote. It's very useful.
but, Boostnote can't syntax highlight Rust. CodeMirror need simple addon for the Rust. 

## Before

simple_addpm_before

## After

simple_addon_after


kazup01:
Thank you @yamash723 ! We will check it.

kazup01:
Merged. Thanks @yamash723 !

PR #1127, opened 2017-11-15
Add task item to indent adjustment target for list

yamash723:
Checkbox type item list is not supported for list indentation by tab key.
I would like to be able to make possible this. Is this specification is accepted?

## Before

indent_before

## After

indent_after


kazup01:
Cool! Thank you for great contribution @yamash723  :)

PR #1137, opened 2017-11-17
add timestamp hotkey

Defenderbass:


kazup01:
Hi @Defenderbass , thank you for your contribution! Could you paste image or gif?

Defenderbass:
Hi @kazup01! You are welcome!
screen1
screen2

browser/main/lib/ConfigManager.js
```diff
@@ -21,7 +21,8 @@ export const DEFAULT_CONFIG = {
   amaEnabled: true,
   hotkey: {
     toggleFinder: OSX ? 'Cmd + Alt + S' : 'Super + Alt + S',
-    toggleMain: OSX ? 'Cmd + Alt + L' : 'Super + Alt + E'
+    toggleMain: OSX ? 'Cmd + Alt + L' : 'Super + Alt + E',
+    timestamp: OSX ? 'Cmd + Alt + T' : 'Super + Alt + T'
```
asmsuechan: I don't think `timestamp` should be located in `hotkey` because `hotkey` was implemented based on the idea that it operates when a user doesn't focus on the application.

asmsuechan:
Please explain specifications which you think.

Defenderbass:
I think that need to leave `timestamp` in` hotkeys`, but separate hotkeys for the editor

kazup01:
Sorry for the long absence.
I close this because we think that this feature is not required by default.
We will make the pulgin later, please make this one at that time.

Btw your suggestion is so cool ;) Thanks again!

PR #1150, opened 2017-11-21
Fix status bar layout

kazup01:
## Before
screen shot 0029-11-21 at 1 23 21 pm

## After
screen shot 0029-11-21 at 1 21 23 pm


kazup01:
Merged

PR #1152, opened 2017-11-21
Fix layout at note item

kazup01:
## Before
screen shot 0029-11-21 at 5 56 33 pm

## After
screen shot 0029-11-21 at 5 53 53 pm


kazup01:
Merged.

PR #1168, opened 2017-11-24
store window size on resize

### LINKED ISSUE #1166: Window size is not saved
Every time I start application the window size is very small and I have to resize it. Previous version 0.8.4 did not have that problem and the window size was restored.

Console does not contain any errors

I'm using Fedora 27 and Boostnote build from master 626175b2b810bbb154afa9b91189efedbf4b4b41

mslourens:
Fixes #1166

```MainWindow.getBounds()``` is always 0 for height and width under Linux, so I store the window size on 'resize' event (with debounce, otherwise the function is called way too often).

### DIFF
--- a/lib/main-window.js
+++ b/lib/main-window.js
@@ -4,6 +4,7 @@ const BrowserWindow = electron.BrowserWindow
 const path = require('path')
 const Config = require('electron-config')
 const config = new Config()
+const _ = require('lodash')
 
 var showMenu = process.platform !== 'win32'
 const windowSize = config.get('windowsize') || { width: 1080, height: 720 }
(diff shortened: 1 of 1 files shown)


kazup01:
Thank you @mslourens for your contribution. We will check it tomorrow.

kazup01:
Hi @mslourens , this seems to have been fixed at # 1144, how about it?

mslourens:
That's also a good fix, I didn't see it. You can close my PR.

kazup01:
@mslourens  Thank you for your replying, I will close this PR.
Btw, your contributions are so helpful every time. Thank you again :)

PR #1113, opened 2017-11-12
React.PropTypes is deprecated from React 15.5

yosmoc:
Migrated by react-codemod + fixed trivial things by hand.

### Before

```
...
/Users/yoshi/src/github.com/yosmoc/Boostnote/browser/main/TopBar/index.js
    1:1   warning  React.PropTypes is deprecated since React 15.5.0, use the npm module prop-types instead  react/no-deprecated
    5:8   warning  'NewNoteModal' is defined but never used                                                 no-unused-vars
   10:9   warning  'dialog' is assigned a value but never used                                              no-unused-vars
   12:7   warning  'OSX' is assigned a value but never used                                                 no-unused-vars
  124:11  warning  'config' is never reassigned. Use 'const' instead                                        prefer-const
  124:19  warning  'style' is never reassigned. Use 'const' instead                                         prefer-const
  124:26  warning  'data' is assigned a value but never used                                                no-unused-vars
  124:26  warning  'data' is never reassigned. Use 'const' instead                                          prefer-const
  124:32  warning  'location' is never reassigned. Use 'const' instead                                      prefer-const
...
✖ 568 problems (0 errors, 568 warnings)
```

### After

```
...
/Users/yoshi/src/github.com/yosmoc/Boostnote/browser/main/TopBar/index.js
    6:8   warning  'NewNoteModal' is defined but never used             no-unused-vars
   11:9   warning  'dialog' is assigned a value but never used          no-unused-vars
   13:7   warning  'OSX' is assigned a value but never used             no-unused-vars
  125:11  warning  'config' is never reassigned. Use 'const' instead    prefer-const
  125:19  warning  'style' is never reassigned. Use 'const' instead     prefer-const
  125:26  warning  'data' is assigned a value but never used            no-unused-vars
  125:26  warning  'data' is never reassigned. Use 'const' instead      prefer-const
  125:32  warning  'location' is never reassigned. Use 'const' instead  prefer-const
...
✖ 527 problems (0 errors, 527 warnings)
```

PR #1170, opened 2017-11-24
fixed eslint warnings

### LINKED ISSUE #1169: Reduce lint warnings
Currently there are over 550 lint warnings. With this issue I will try to reduce them.

mslourens:
fixes #1169 

Reduces 550+ warnings to 47, which is this one in 47 different files:
```React.PropTypes is deprecated since React 15.5.0, use the npm module prop-types instead```
I couldn't fix that warning, because we are not using React 15.5.0 but 15.0.2, so this warning doesn't apply for this project.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -12,5 +12,10 @@
     "react/no-find-dom-node": "warn",
     "react/no-render-return-value": "warn",
     "react/no-deprecated": "warn"
+  },
+  "globals": {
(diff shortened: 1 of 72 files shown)


kazup01:
Thank you for your contribution. You are amazing!
I will check it tomorrow.

`React.PropTypes is deprecated` is fixed at #1113 and wait for review :)

PR #1179, opened 2017-11-26
Fix Margin at Note Detail

kazup01:
## Before
screen shot 0029-11-26 at 4 15 51 pm

## After
nov-26-0029 16-17-49


e-volusian:
I'm not sure the most appropriate place to put this comment, so I'll just put it here. Kazz: your responsiveness to user input is what creates a vibrant and enthusiastic user base. I came out here to suggest this change and to see that it's already implemented after <3 days is just...amazing. Thanks!

kazup01:
v0.8.18 is out! Thank you for waiting guys ;)
https://medium.com/boostnote/boostnote-v0-8-18-release-50e5c143778

PR #1180, opened 2017-11-26
SideNavItem: add background when hover

romainwn:
When SideNav is folded, there is no background-color when I hover on an item

Before : 
screen shot 2017-11-26 at 10 46 43

So I added a background-color in order to see better what the SideNavItem is about

After : 
screen shot 2017-11-26 at 11 13 34


kazup01:
Merged. Thanks @romainwn ;)

PR #1184, opened 2017-11-26
Fix surrogate pairs garbling on folded SideNav

cormoran:
I fixed emoji garbling on folded sidebar.

Before:

After:


kazup01:
Thanks @cormoran! Wait for the next update :)

PR #1202, opened 2017-11-29
Fix folded layout

kazup01:
## Before
screen shot 0029-11-29 at 4 43 53 pm

## After
screen shot 0029-11-29 at 4 42 24 pm


kazup01:
Merged

PR #1211, opened 2017-11-30
confirm tag at onBlur event

yosmoc:
## Before

tag_before

## After
tag_after

When user inputs the tag and leave the tag input box without fixing(enter or tab key), tag string is still there, but it is not stored as a tag.

This PR solves this problem. When the cursol is out of the tag input, it registers the current input contents as a tag.


kazup01:
Thank you for your suggestion @yosmoc ;)

PR #1217, opened 2017-12-01
Use existing file metadata for created and modified dates

SCdF:
NB: this definitely works on OSX. Other operating systems may have
slightly different interpretations of (eg) birthtime.

See: https://nodejs.org/api/fs.html#fs_class_fs_stats

### DIFF
--- a/browser/main/NoteList/index.js
+++ b/browser/main/NoteList/index.js
@@ -584,22 +584,29 @@ class NoteList extends React.Component {
     filepaths.forEach((filepath) => {
       fs.readFile(filepath, (err, data) => {
         if (err) throw Error('File reading error: ', err)
-        const content = data.toString()
-        const newNote = {
-          content: content,
-          folder: folder.key,
-          title: markdown.strip(findNoteTitle(content)),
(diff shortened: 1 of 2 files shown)


sota1235:
@kazup01 Could you check whether this PR works on windows? If it works, it's ok

kazup01:
@sota1235 Sure. I will check it tomorrow.

kazup01:
Hi @SCdF ,

An error occurs in the following procedure.
- Select note in the note list
- Move  note to another folder with drag and drop
- The number of notes is not counted in the number next to folder name
- An error occurs when selecting another note.

Attach gif and errors in the console.

dec-14-0029 13-16-21

screen shot 0029-12-14 at 1 16 59 pm

SCdF:
Nice catch! I didn't realise `createNote` is involved when you move notes. This has happened because I flipped the order of an Object assign to prioritise calculated dates over default "now" ones.

I've rejigged it so that it this is no longer a problem.

kazup01:
Merged. Thanks @SCdF ;)

PR #1241, opened 2017-12-05
fixed global styles

### LINKED ISSUE #1240: global.styl not working correctly
There are a couple of variables in global.styl which are not resolved, and therefore the style is not applied correctly. To give you an example:

image

The cause is that the vars.styl (where fontSize and textColor are defined) is not imported in global.styl. Also styles/main/index.styl is never used and has the same content as global.styl and can be removed. 

I will fis this issue myself.

mslourens:
fixes #1240


mslourens:
This is how it looks like now:
image

.gitignore
```diff
@@ -8,3 +8,5 @@ node_modules/*
 /compiled
 /secret
 *.log
+.vscode
+.idea
```
kazup01: I think that `.vscode` and `.idea` are unnecessary. Could you remove them?

mslourens: I just added them, because I use both editors and always have to be careful not to commit those IDE specific files. Is it a real problem that we add them here?

kazup01: I understand your opinion, but I think that we should merge only generally code (not only for us)  because Boostnote is open source.

sosukesuzuki: We don't think it is unnecessary to add `.idea` and `.vscode` into `.gitignore`. However, if many contributors want it, we'll add it. 😄 
And,  This pull request is "fixed global styles". could you create new pull request for add it.?🙏

mslourens:
I removed the gitignore changes and created a new PR with these changes: #1242

kazup01:
Thanks @mslourens ! I will check this tomorrow.

kazup01:
Merged. Thanks @mslourens !

PR #1244, opened 2017-12-05
fixed incorrect fontFamily for multiple fonts

### LINKED ISSUE #1232: Font family setting not working anymore
Font family for both editor and preview can not be set in config anymore.

Both editor and preview window always uses the default font from version 0.8.18 or maybe the previous version.

mslourens:
fixes #1232


kazup01:
Thanks @mslourens ;)

Issue #1191, opened 2017-11-27
Sidebar Collapse Styling


Nino9614:
I really like the new update, good job on the new look and fell. 
But for some reason my sidebar looks like this when i collapse it.

![boostnote01](https://user-images.githubusercontent.com/27356972/33256958-9c593d8e-d354-11e7-9c3f-a1a9437d5029.PNG)

![boostnote02](https://user-images.githubusercontent.com/27356972/33256969-a24e8dca-d354-11e7-8982-eabc2e9ccc4b.PNG)

Is this something you aware of? As far as i can see on other issues the sidebar looks fine.

hooskers:
I like this!
It seems like that's how it is supposed to work.

Out of curiosity, what do you think it should do?

The double arrow at the bottom left of the sidebar collapses it to a condensed version so you can still have your sidebar functionality, but with it taking up less room.
If you want to hide the sidebar completely you can use the button on the far right of the editor (third button down, below the star).

kazup01:
Hi thank you for feedback. I tried to fix this layout at #1202. How about it?

Nino9614:
@hooskers  

> Out of curiosity, what do you think it should do?

Its not about hiding the sidebar. The icons should align to the center like. 
The changes @kazup01 made in #1202 is what i meant.
@kazup01 thanks for the layout fix, looking good. When do you consider to make the change public?

kazup01:
@Nino9614  Thanks for reply 😄 
We will update to v0.8.18 in this week. Wait a moment.

kazup01:
Hi guys, v0.8.18 is out! Enjoy Boostnote ;)

Release Note
https://medium.com/boostnote/boostnote-v0-8-18-release-50e5c143778

Nino9614:
@kazup01 The new Update fixed the styling for me. Looking fire again :) 
You can close this issue if you want. Keep up the great work

Nino9614:
@kazup01 just found another weird behavior. When i collapse the sidebar everything seems fine. But when i select a element the icon gets pushed to the right: 
![image](https://user-images.githubusercontent.com/27356972/33711919-f985ca1a-db45-11e7-91e3-345433fa8cca.png)
Is this something you wanted to do or is this a bug? 
It doesn't happen when i click on the "All Notes" Element

kazup01:
I could reproduce it. 
[...]

PR #1254, opened 2017-12-08
Fix the folded side bar layout

kazup01:
Fix https://github.com/BoostIO/Boostnote/issues/1191#issuecomment-349934313


kazup01:
Merged

Issue #933, opened 2017-10-04
[Feature Request] Add per-folder sort


mpatraw:
## Context
Currently, when changing sorting method, this is an application setting, not a folder setting. So when you sort one folder and then go to another, it uses that same sort setting. Ideally this could be set on a per-folder basis.

## Benefits
* Some folders are used for different things that require different sorting
  * Logs, or daily notes would be creation date
  * Miscellaneous could be sorted alphabetically
* Predictable behavior when switching folders, it looks exactly the way you left it

kazup01:
+1

IssuehuntBot:
@mikaoelitiana has submitted [output](https://github.com/BoostIO/Boostnote/pull/2284). [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/933)

IssuehuntBot:
@kazup01 has rewarded. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/933)

PR #1259, opened 2017-12-08
sortby config per folder

mslourens:
implements #933


kazup01:
We think this feature is unnnecessary because complicated.
I close it, but thanks for awesome suggestion! @mslourens

PR #1269, opened 2017-12-11
Add the Solarized Dark Theme

kazup01:
screen shot 0029-12-11 at 7 00 10 pm


kazup01:
We will make more it better.

kazup01:
cc @ytk141

Issue #1273, opened 2017-12-11
No Pin Icon for pinned notes in Compact View


AlexEshoo:
When a note is pinned, there is a pin icon in card view which is nice.
![image](https://user-images.githubusercontent.com/8216513/33841022-87fa829c-de64-11e7-9696-7d3062ca6eed.png)

But when in compact view there is no indication that the note is pinned. Perhaps making the note icon red is sufficient in this view?
![image](https://user-images.githubusercontent.com/8216513/33841039-953419fa-de64-11e7-8e18-72713d8084ce.png)

kazup01:
Hi @AlexEshoo , thanks for your suggestion! I will add it later :)

kazup01:
Added at #1275.

Nino9614:
@kazup01 are you considering to also show the star icon when in compact mode? Or is that to much inside the layout?

kazup01:
v0.8.19 is out!
https://medium.com/boostnote/boostnote-v0-8-19-is-out-with-new-theme-markdown-live-preview-and-more-25f6caaad61e

kazup01:
Hi @Nino9614 , Since Star can be used from SideNav, I think unnecessary to display it on Notelist. What do you think about it?

PR #1275, opened 2017-12-11
Add Pin to NoteItemSimple

kazup01:
#1273 

## Before
screen shot 0029-12-12 at 2 28 06 am

## After
screen shot 0029-12-12 at 2 26 45 am

### DIFF
--- a/browser/components/NoteItemSimple.js
+++ b/browser/components/NoteItemSimple.js
@@ -14,7 +14,7 @@ import styles from './NoteItemSimple.styl'
  * @param {Function} handleNoteContextMenu
  * @param {Function} handleDragStart
  */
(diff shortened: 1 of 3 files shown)


browser/components/NoteItemSimple.js
```diff
@@ -30,6 +30,10 @@ const NoteItemSimple = ({ isActive, note, handleNoteClick, handleNoteContextMenu
         ? <i styleName='item-simple-title-icon' className='fa fa-fw fa-code' />
         : <i styleName='item-simple-title-icon' className='fa fa-fw fa-file-text-o' />
       }
+      {note.isPinned && !pathname.match(/\/home|\/starred|\/trash/)
```
sota1235: writing logic on stateless component does not make sense because testing component will be harder.
plz receive boolean of which means same as `!pathname.match(/\/home|\/starred|\/trash/`

kazup01:
Merged

Issue #1261, opened 2017-12-08
Export as MarkDown (.md) Image Link Broken


DevelopRx:
**Export as MarkDown (.md)** only exports literal .md and there is no path to images in :storage.

It would be nice if associated images in the .md file could also be exported from :storage into a separate local folder (images) and the .md revised accordingly. 

Current image link export output in .md: `![MyImage.jpg](\:storage\pfvw27psq2cblnmi.jpg)`

**What I Suggest:**

1. **pfvw27psq2cblnmi.jpg** renamed **MyImage.jpg** and copied into local export **images** folder
2. Link to image is updated in .md: `![MyImage](images/MyImage.jpg)`

Thanks!

nlopin:
I'll fix it this week.

DevelopRx:
Thanks!

nlopin:
@kazup01 this one is done

kazup01:
Thank you guys :)

PR #1256, opened 2017-12-08
first attempt to export html

mslourens:
minimum viable implementation of #1233, please let me know what you think.

PR #1306, opened 2017-12-17
Export note with local images

nlopin:
Looks through the note and searches for local images. Copies them to ‘images’ folder in the export path and replaces affected links.

#1261


nlopin:
@kazup01 any news?

kazup01:
@nlopin  Thanks for your support! We will review it at sometime soon. 
Thank you for waiting :)

kazup01:
Sorry for the long absence. Could you fix some conflicts?

nlopin:
Sure

nlopin:
@kazup01 it will be an issue with html export (#1256). There is a code to receive an HTML for export:
```
this.exportAsDocument('html', (value) => {
      return this.refs.root.contentWindow.document.documentElement.outerHTML
})
```

As you can see, the whole HTML of the preview window is fetched and exported then. But, it contains relative links to local styles and fonts. For example, all of those styles will be lost after user copy the exported HTML to another computer:
```
<link rel="stylesheet" href="file:///Users/nlopin/Workspace/Boostnote/node_modules/katex/dist/katex.min.css">
<link rel="stylesheet" href="file:///Users/nlopin/Workspace/Boostnote/node_modules/codemirror/lib/codemirror.css">
<link rel="stylesheet" id="codeTheme" href="file:///Users/nlopin/Workspace/Boostnote/node_modules/codemirror/theme/dracula.css">
```

It is not a good idea to export content like this because it does not provide consistent view on different machines. 

We do have `markdown-it` plugin and I want to try it for rendering an HTML output. But we need to decide what to do with our custom styles:
1. Should we keep them and include to the export, or ignore them?
2. Custom font "Lato". It's impossible to embed it into the page. Any ideas what can we do here?

Rokt33r:
@nlopin
1. I think we can insert style into `<head>` element of html file.
2. We can't. The user must install their custom font to each machine.

nlopin:
@Rokt33r thanks for your response. 

Looks like you didn't catch the issue in the first question. The question is not about where should we put styles. The issue is: We'll have to export those css files along with html. Otherwise it won't work properly.

If it's ok I'll do this.

Rokt33r:
@nlopin It makes sense. I've realized now that it is just like what Chrome does.

Yea, we should export multiple files, style sheet, images and any other static stuff, including fonts if url is available.

Rokt33r:
@nlopin ping.

Rokt33r:
@nlopin ping.

nlopin:
@Rokt33r hi. Sorry for delay. I'm working on it right now.

nlopin:
@Rokt33r @kazup01 I've finished. 

Export is divided by two parts:
1. export additional files and paths modifications
2. export the content of the note

Additional files are exported via copy tasks. Task is an object `{'src': pathfrom, 'dst': pathto}`. Destination path can be relative or absolute. 

If all additional files are exported, the content will be exported at the end. If an error happened during export, boostnote remove all already exported files and folders.


Feel free to ask questions here or at Slack.

Rokt33r:
@nlopin Great work! 👍 I'll review this tomorrow!

tatoosh:
Will u also include CSS ?

nlopin:
@tatoosh yes. If you export to HTML two additional folders will be exported – `images` and `css`

tatoosh:
THX @nlopin!
Do you have an option for Export CSS, so Form dark theme the Export will transformed to white?

nlopin:
No, it will export the same styles you see on preview. Create feature request if you need it

browser/main/lib/dataApi/exportNote.js
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    return task
+  })
+}
+
+function saveToFile (data, filename) {
+  return new Promise((resolve, reject) => {
+    fs.writeFile(filename, data, (err) => {
+      if (err) throw err
```
Rokt33r: I assume this makes the error go limbo. `return reject(err)` should be right.

nlopin:
@Rokt33r done! Thank you

browser/main/lib/dataApi/exportNote.js
nlopin: Fixed

Rokt33r:
Now we need to implement UI. I'm going to create another PR against this branch.

nlopin:
Could you explain what do you mean? 
There is no sense in current version of HTML export, because it doesn't work properly. I think it's ok to merge PR because it fixes that issue.

nlopin:
@kazup01 @Rokt33r I'm worried that leaving this PR without merge will lead us to new set of conflicts.

Rokt33r:
@nlopin sorry for my late answer. Everything looks great. I'll merge this today!

Rokt33r:
I found a problem.

screen shot 2018-02-10 at 9 11 57 am

Styling of task list seems to be broken. I'm investigating now.

Rokt33r:
Merged!

nlopin:
This is the fundamental issue. Some styles applies manually to Preview HTML in `rewriteIframe` method. It's not a good practice, so I removed that. 

I add class to the `li` tag at markdown to html transformation and then style it with css. @Rokt33r  Please, have a look

Rokt33r:
@nlopin Could you create another PR? I can review it now!

nlopin:
Done #1531

PR #1307, opened 2017-12-18
Split markdown

sosukesuzuki:
I implemented the new markdown editor.
It is split in half by markdown preview, and half by text editor.


kazup01:
Merged

PR #1312, opened 2017-12-20
Fix switch button layout

kazup01:
## Before
screen shot 0029-12-20 at 6 51 50 pm

## After
screen shot 0029-12-20 at 6 50 19 pm


kazup01:
Merged

PR #1354, opened 2018-01-01
Elixir syntax highlighting

trinode:
# Syntax Highlighting for Elixir:

Things it should do:-
- [x] Highlighting in code editor
- [x] Highlighting in preview
- [x] Highlighting in snippets
- [x] Detect Elixir from .ex extension in a snippet file
- [x] Appear in the list of syntax types for manual selection
- [x] Apply when a code block is declared with ` ```elixir `
 

## Note
image

## Snippet
image


trinode:
Added a second commit to sort the languages in the menu as they were appearing out of order

trinode:
So I had this built a few times and it seems to randomly fail either LTS or stable, but only one, and not the same one each time... Is there some build issue somewhere?

trinode:
Awesome, thanks.

@kazup01 do we need to change the label from "In Review" so it will be merged? Are we in time for the next release?

kazup01:
Merged. Thanks for your contribution @trinode :)
We will release it at v0.8.20!

kazup01:
Hi @trinode , thanks for your support! We were released v0.8.20.
[Release Note](https://medium.com/boostnote/boostnote-v0-8-20-release-4e1c1b1e1dbd)
Enjoy Boostnote :)

PR #1368, opened 2018-01-05
fix(code-editor): use correct prop for font family

### LINKED ISSUE #1358: Invalid font-family for editor
I am experiencing an issue where the font for the editor is by default always set to an invalid value, instead of the font I specify in the settings. This is not the case for the preview font-family, so it's not as if the font could not be found or is provided incorrectly.
The value provided is always the following:
```css
font-family: MarkdownEditor__codeEditor___browser-components-, Monaco, Menlo, "Ubuntu Mono", Consolas, source-code-pro, monospace;
```

This issue was introduced in the previous version (0.8.18).
Even after upgrading to the newest version which supposedly fixes being able to set fonts from settings.

**OS**: Arch Linux
**Version**: 0.8.19-1
**Package**: https://aur.archlinux.org/packages/boostnote/

**Settings**:
image

**Issue**:
image

Any idea on what's going on?
So far I've been able to deal with it by opening the developer tools and changing things there, but I shouldn't have to do this each time.

marcusstenbeck:
Closes #1358 

Bug was introduced here: https://github.com/BoostIO/Boostnote/commit/3fbc749395b7c90643d05ee09b8e7f74e00eb190#diff-4771352a7e4e4f9d82f0be3e914757adL245

### DIFF
--- a/browser/components/CodeEditor.js
+++ b/browser/components/CodeEditor.js
@@ -250,7 +250,7 @@ export default class CodeEditor extends React.Component {
 
   render () {
     const { className, fontSize } = this.props
-    let fontFamily = this.props.className
+    let fontFamily = this.props.fontFamily
     fontFamily = _.isString(fontFamily) && fontFamily.length > 0
       ? [fontFamily].concat(defaultEditorFontFamily)
       : defaultEditorFontFamily


kazup01:
@marcusstenbeck  Awesome! Thanks for your contribution 😄
We will update to v0.8.20 soon.

kazup01:
Hi @marcusstenbeck , thanks for your support! We were released v0.8.20.
[Release Note](https://medium.com/boostnote/boostnote-v0-8-20-release-4e1c1b1e1dbd)
Enjoy Boostnote :)

PR #1412, opened 2018-01-15
Fix the Solarized Dark Styling

### LINKED ISSUE #1409: Solarized Dark Styling Problem
I noticed the styling is not applied correctly when using the solarized dark theme: 
unbenannt

Also there is a black rectangle that appears as soon as i change a preference and save it. 
This rectangle "follows" me on every menu i open

kazup01:
Fix #1409 

screen shot 0030-01-15 at 8 18 43 am


kazup01:
Merged

PR #1413, opened 2018-01-15
Fix the folded layout on side bar

### LINKED ISSUE #1408: Sidebar Layout Problems
I already opened a similar issue about this a while ago see: #1191
The Icons get pushed to the side when selected. 
Ironically only the "star" and "trash" Icon has this behavior
but the issue still appearing on version:  0.8.20: 
unbenannt

Also as you can see in the screenshot there is no icon for the preferences menu when sidebar is "collapsed". Is this wanted or just a bug? Previously there was some sort of menu icon

unbenannt

kazup01:
Fix #1408

### DIFF
--- a/browser/components/SideNavFilter.styl
+++ b/browser/components/SideNavFilter.styl
@@ -92,7 +92,6 @@ body[data-theme="white"]
     color $ui-inactive-text-color
 
   .menu-button--active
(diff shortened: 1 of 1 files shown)


kazup01:
Merged

Issue #1405, opened 2018-01-14
Tweak "Info" popup


ww7:
![](https://i.imgur.com/5uo26HB.png)
- resize to fit all info without scroll
- add  copy to clipboard 📋  button for note link

kazup01:
v0.9.0 is out! Thank you for your suggestion @ww7 :)
[Release Note](https://medium.com/boostnote/boostnote-v0-9-0-release-5a376e21b625)

PR #1416, opened 2018-01-15
added button for copy note link

mslourens:
implements #1405 - question: do we want to show the copy notification, like in the markdown preview once the button has been clicked and the link has been copied to the clipboard?


kazup01:
Cool! Thanks @mslourens :)

PR #1419, opened 2018-01-15
added encoding meta tag

### LINKED ISSUE #1399: Text encoding in the exported HTML
Thanks for releasing 0.8.20, especially the HTML export feature!

Everything seems perfect, except the character encoding issue in some browsers. The exported html doesn't seem to specify its character set, so when I open an exported html file in FireFox (57.0.4, Mac), it uses whatever encoding is configured as the default (Western in my case).

That makes some characters displayed incorrectly, such as smart quotes, LaTeX math equations, etc. If I force the encoding to Unicode, they are displayed correctly.

Chrome seems to use Unicode by default and is ok.

Please specify the character encoding in the header of the generated html. I think something like
```html
<meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
```
will work.

Thanks!

mslourens:
fixes #1399


kazup01:
Merged.

PR #1429, opened 2018-01-17
Remove inline-style from markdown link list

### LINKED ISSUE #1401: Markdown list
Bug with list when item without spaces (links) not fit in one line

yamash723:
Fix #1401

### Before
before

### After
after


kazup01:
Thank you @yamash723 -san :)

PR #1470, opened 2018-01-26
Ct portable

CTrauma:
If the application finds a file named '.portable' in it's working directory it will set userData to a local directory instead of the user's home directory. This allows the application to run from an external drive and moved from system to system.


Rokt33r:
Your code looks good.

Which OSs have been tested? And, could you update readme.md about this?

Rokt33r:
Ok, I just want you to check that you've tested it on Linux. It sounds fine.
Could you disable this feature on macOS? It doesn't seem to work on macOS.

You can do it like the below.
```js
if (process.platform !== 'darwin') {
}
```

And, you can ignore the error from our CI. We don't support the current stable version, v9.x, yet.

Rokt33r:
And I assume portable app is not compatible with auto update and installer.
How are you going to install the app on Windows?

I assume we have to disable auto update too.

CTrauma:
ok.. let me look at the auto update issue, haven't had much time lately, will have a look as soon as I can.

RoyiAvital:
Great.
Looking forward to see it in released version.

Thank You.

Rokt33r:
Close this PR until you come back. @CTrauma

PR #1481, opened 2018-01-30
Remove leading # when creating tag

### LINKED ISSUE #1476: [Feature Request] Automatically remove #'s if you put one in the tag area
I often find that in the tag area, I'll write `#engineering` instead of `engineering`, causing it to be called `##engineering` - triggering.

It'd be pretty :fire: to be honest

andyklimczak:
- Remove leading # from tags that are created with leading # (user types in `#tagName`)
- Convenience method for users who tend to type them, but did not want a tag with two #
- If a user wants a tag with a leading #, then can enter `##OfPeople`, and only the first will be removed
- Resolves #1476


kazup01:
Merged. Thanks for your contribution @andyklimczak :)

PR #1483, opened 2018-01-30
Fix permanently deleting note

### LINKED ISSUE #1465: Deletion of a note in Trash isn't effective until reload
In previous version, if you delete a note that is already in Trash, that note was permanently deleted and that was effective right away. But with v0.8.20, it is not effective immediately and the note still remains in the Trash UI. Only after I reload (View > Reload, Cmd-R on macOS), it is removed from Trash.

andyklimczak:
- Permanently deleting note would not remove note from list until after refresh
- Fix so that permanently deleted note is removed from list
- Resolve #1465 

This bug was created in 9f9463f0e8932ae02dec4d4b6bd243e409d71e25 (git bisect), and I'm not sure of what review comments the commit message is referring to (looked through a few PRs around that time). So this fix might not be the fix we really want, depending on those comments.

### DIFF
--- a/browser/main/Detail/MarkdownNoteDetail.js
+++ b/browser/main/Detail/MarkdownNoteDetail.js
@@ -210,10 +210,9 @@ class MarkdownNoteDetail extends React.Component {
         }, () => {
           this.save()
         })
-
-        ee.emit('list:next')
       }
     }
+    ee.emit('list:next')
   }
 
   handleUndoButtonClick (e) {
--- a/browser/main/Detail/SnippetNoteDetail.js
+++ b/browser/main/Detail/SnippetNoteDetail.js
@@ -203,10 +203,9 @@ class SnippetNoteDetail extends React.Component {
         }, () => {
           this.save()
         })
-
-        ee.emit('list:next')
       }
     }
+    ee.emit('list:next')
   }
 
   handleUndoButtonClick (e) {


sosukesuzuki:
thanks for your contributing.
Sorry, the code causing this bug was written by my review. 
The reason why I did such a review is that because if we select `cancel`, Boostnote select next note, like below.

(This video was captured in this PR.)
If you can, I want you to fix it too.

andyklimczak:
I see, I will take a look!

browser/main/Detail/MarkdownNoteDetail.js
```diff
@@ -200,6 +200,7 @@ class MarkdownNoteDetail extends React.Component {
             }
             ee.once('list:moved', dispatchHandler)
           })
+          .then(() => ee.emit('list:moved'))
```
andyklimczak: The issue really is that `dispatchHandler` is never called. So we `emit` the event for the dispatch handler to be run.

I think this is better than just directly calling the dispatch handler (`dispatchHandler()`) because it remains asynch using the event emitter (I believe).

andyklimczak:
Updated and included small explanation. I'm not 100% sure if this is the best way, there are a few ways to solve this (by calling the dispatch handler) that seem to have subtle differences.

Thanks!

browser/main/Detail/MarkdownNoteDetail.js
sosukesuzuki: I see. I think it is better than calling the dispatch handler.
but, if we delete a note in trash, Boostnote select nothing, like below.

this behavior is not we hope.

andyklimczak:
Updated to use `list:next` so that the next note in the trash is selected (if one exists)

andyklimczak:
Test failure seems like a flaky test (also seen happening [here on master](https://travis-ci.org/BoostIO/Boostnote/jobs/335525728))

Issue #1493, opened 2018-02-01
Not using newer version of Katex


cenhao:
Katex has added quite a few more syntax support like `overbrace` after 0.8.0, would be super useful that we can update the katex version from 0.7.1 to newer one.
![image](https://user-images.githubusercontent.com/2842841/35704984-fc225158-0755-11e8-9794-aeb1e28af4d1.png)

Rokt33r:
Fixed.

PR #1494, opened 2018-02-02
Update KaTeX library from 0.7.1 to 0.8.3

Paalon:
See #1493.


Paalon:
Do I need to add the generated yarn.lock?

Rokt33r:
@Paalon Please ignore CI failing. We don't support the current stable version of Node.js yet.

PR #1502, opened 2018-02-04
I fixed the behavior of the editor when the enter key is pushed.

Matts966:
On the enter key pushed, if already the line contains a token in markdown list, no completion is needed, I think.

For example, if I push the enter key when md is like below, and my cursor selects the head of the line,
```
- [ ] a
```

the behavior before
```

- [ ] - [ ] a
```

the behavior after.
```

- [ ] a
```

By this change, node_modules/codemirror/addon/edit/continueList.js could be trashed, but I can't because of the gitignore.

### DIFF
--- a/browser/components/CodeEditor.js
+++ b/browser/components/CodeEditor.js
@@ -92,7 +92,7 @@ export default class CodeEditor extends React.Component {
         'Cmd-T': function (cm) {
           // Do nothing
         },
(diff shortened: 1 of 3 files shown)


Rokt33r:
GJ!!:+1: I'll review this by tomorrow. How about to make another pull request against Codemirror to fix this problem?

Matts966:
Thank you!
I try to make it.

Issue #1482, opened 2018-01-30
[Feature Request] Bring in web page title when pasting URL on editor


avence12:
I would like to translate plain URL into markdown link with web page title.

Example:
1. Pasting https://github.com/BoostIO/Boostnote/issues on editor
2. Automatically transform `https://github.com/BoostIO/Boostnote/issues` into `[Issues · BoostIO/Boostnote](https://github.com/BoostIO/Boostnote/issues)` by extracting the title tag from HTML

![image](https://user-images.githubusercontent.com/4282367/35544656-8706da02-05a6-11e8-99d0-0cb2517b570d.png)

glego:
You should have the ability to disable it and take into consideration when pasting between braces it doesn't nest.

like
```
([Issues · BoostIO/Boostnote](https://github.com/BoostIO/Boostnote/issues))
```

PR #1509, opened 2018-02-05
Feat/display url title

Antogin:
Hi, this is my proposition for request #1482 

peek 2018-02-04 17-46

The user can opt out (or should it be a opt in functionality?
peek 2018-02-04 17-51

This is what happens when there is no connection
fail

I use <kbd>[</kbd> <kbd>]</kbd> to tag the url and target it, if there is an invisible character that i can use, it would be cool. I did not find anything usable.
peek 2018-02-04 17-52


Rokt33r:
How about to use `<>` rather than using invisible character?

`<https://github.com>` should be rendered as a link like the below.

<https://github.com>

Antogin:
Works perfectly! Thanks

Issue #755, opened 2017-08-01
[Feature request] Highlight searched keyword


melibou:
Hi, 

When searching something, to highlight the keyword on the text and show a  < > sings for next and after. Additionally, to show how many times this keyword show up. 

Regards,
Meli

BobTB:
Oh,... this note taking app also does not have this.... I had so high hopes for this. It is very hard to use if search result is not highlighted in the text.

yahyavi:
Yes, this is a deal breaker for a note taking app.

MakisH:
The "Find" (inside one note) highlights the terms, but the global "Search" not. Maybe a good idea would be to merge the "Search" and "Find" features. Why are they both needed?

yahyavi:
Find does highlight but then how do you go to the next item? hitting enter writes over the selected value. Using n as in vim also doesn't work.

sjtindell:
Bump, this would be a great feature.

lotuc:
Currently It supports searching in md editor, but search in rendered view will be very useful, desperately need this.

PR #1524, opened 2018-02-09
Highlight global search matches on code editor

Redsandro:
I like to see the global search matches in the markdown editor for productivity reasons. I implemented this for personal use, and I hope it will be useful to others.

This partially addresses #755.

This PR makes it more evident that the global search box could really use both an "x" symbol that appears to clean/delete the current input (for touch users) or having the escape key clean the input (for keyboard users), but that is beyond the scope of this PR. See #1525.

### DIFF
--- a/browser/components/CodeEditor.js
+++ b/browser/components/CodeEditor.js
@@ -7,6 +7,7 @@ import path from 'path'
 import copyImage from 'browser/main/lib/dataApi/copyImage'
 import { findStorage } from 'browser/lib/findStorage'
 import fs from 'fs'
+import eventEmitter from 'browser/main/lib/eventEmitter'
 
 CodeMirror.modeURL = '../node_modules/codemirror/mode/%N/%N.js'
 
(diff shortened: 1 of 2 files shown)


Redsandro:
image

sosukesuzuki:
If we update keyword in search box, code editor is hidden and markdown preview is shown.
This behavior is not great.

Redsandro:
@sosukesuzuki I hadn't observed that. I think it's because I have split view enabled by default, and you have the preview by default.

I can fix this, if you would consider it.

Redsandro:
@sosukesuzuki I have examined this. What you are observing is original behavior by Boostnote. If you search, your list of notes refreshes, the code editor closes, and a new note from the list is viewed in the preview. It was already like this, by design (not by me). 

This PR adds global search highlighing in the code editor. Especially useful in dual pane editing. Whether or not the editor should close when search is used, is not decided, designed, or changed by this PR. It is also not in the scope of this PR.

If the note selected by search should be opened in the code editor by default (e.g. when the code editor was previously selected), this should be a different PR or feature request.

Please consider merging this PR as it will be useful for some, and consider writing a PR for the extra features you think would be useful to others as a result of the functionality this PR brings to the table.

PR #1527, opened 2018-02-09
Trashed notes should not be counted

Redsandro:
The selected note category count should equal the amount of notes visible in the note list. Currently, trashed notes are counted as notes, but are not visible, so the counts don't add up.

image

This PR fixes that.


kazup01:
Merged.

Issue #1507, opened 2018-02-05
[Bug] New notes and the all of the top notes aren't shaded


Paalon:
When I started boostnote, the folder's top note "test" was selected and should be shaded but not.
![2018-02-05 13 14 40](https://user-images.githubusercontent.com/11065457/35788468-7d3d8ee0-0a78-11e8-8b2b-9160494d4157.png)

I selected the note "test" and it was shaded.
![2018-02-05 13 15 05](https://user-images.githubusercontent.com/11065457/35788464-77e1ddb6-0a78-11e8-85ae-d7d9b4b647f4.png)

Then I created new note but the new "Empty" note was not shaded. The previous selected note "test" is shaded.
![2018-02-05 13 14 53](https://user-images.githubusercontent.com/11065457/35788461-72104616-0a78-11e8-802f-eb8e6bac91fe.png)

nlopin:
I've noticed that too. I'll fix it

Redsandro:
Both issues are fixed in #1528.

PR #1528, opened 2018-02-10
UX: Keep selection synced with note visibility

Redsandro:
This PR fixes two UX bugs:

1. The visible note should always be selected or part of a selection.
1. When navigating from a set to a superset or vice-versa, the visible (selected) note should not change if it's part of both sets.

Previously, it was confusing to follow what's going on, because the selection and note visibility were not synced. This demonstrates the old behavior. **Play close attention to the note in view on the right, and the selection in the note list:**

peek 2018-02-10 01-23 wrong

With this PR, the selected note will remain selected as long as it's part of the subset (e.g. when searching or changing storage), and if not, when the first note is automatically viewed (old behavior) the selection will be updated to match:

peek 2018-02-10 01-20

### DIFF
--- a/browser/main/NoteList/index.js
+++ b/browser/main/NoteList/index.js
@@ -109,14 +109,27 @@ class NoteList extends React.Component {
 
   componentDidUpdate (prevProps) {
     const { location } = this.props
+    const { selectedNoteKeys } = this.state
+    const visibleNoteKeys = this.notes.map(note => `${note.storage}-${note.key}`)
+    const note = this.notes[0]
+    const prevKey = prevProps.location.query.key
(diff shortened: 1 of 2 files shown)


Redsandro:
I forgot the case where someone creates a new note and fixed it in this PR. This closes #1507.

Redsandro:
Hey @kazup01, got any update on the review process? This PR is really nice to have. I notice when I switch laptops (one with this PR applied and one with vanilla Boostnote).

kazup01:
Hi @Redsandro , yeah this PR is awesome. We have many tasks now, so wait a while please.
Thanks!

kazup01:
Merged. Thank you for your contribution and waiting @Redsandro :)

Issue #1523, opened 2018-02-09
[Feature Request] Edit on double-click in stead of single-click


Redsandro:
I find myself constantly toggling the edit mode (preview -> editor) when I don't want to.

A solution would be to (have the option to) only switch to the editor when double-clicking on the preview, in stead of single-clicking.

kazup01:
Hi @Redsandro , thank you for your suggestion. Btw you can use the option that right click in Preferences.

<img width="1150" alt="screen shot 0030-02-10 at 4 16 53 pm" src="https://user-images.githubusercontent.com/8602615/36059568-f1ef3d8e-0e7d-11e8-92da-6fc2b0ab386e.png">

PR #1543, opened 2018-02-13
Switch from preview to edit by double click

nlopin:
User want to click markdown to work with text (for example copy).
If they have "Switch to Preview" setting in "When Editor Blurred",
they enter edit mode after click. To fix that issue I introduced new
value to the "Switch to Preview" setting that do that thing. User can
enter edit mode by double click on the editor and leave it on blur.

Unfortunately, it's impossible to switch both ways by double click
because the editor mode is listening for double click event.

#1523


kazup01:
Merged. Thanks for your contribution @nlopin !

Issue #1506, opened 2018-02-04
 [Feature Request] Global "Pin to Top"


tatoosh:
Working with more then one storage the "Pin to Top" will only work per storage, not for "all notes":

![grafik](https://user-images.githubusercontent.com/3670907/35782963-ab017b84-0a00-11e8-9623-b80d1286e135.png)

nlopin:
@kazup01 this one is merged

PR #1544, opened 2018-02-14
Show pinned notes in All notes view and allow to pin them there

nlopin:
All restrictions of pin functionality were removed. Now user can see pinned notes and also pin notes right from 'home'. Technically, a note still pins to the storage it belong.

#1506


kazup01:
Merged. Thanks @nlopin :)

Issue #1325, opened 2017-12-22
More intuitive UI for restoring items in trash


initializard:
Current restore icon: 
![screen shot 2017-12-22 at 11 09 55 am](https://user-images.githubusercontent.com/2012737/34306440-b785cf06-e708-11e7-8bb7-6b2131d2d7c0.png)

It took me longer than I'd like to admit to figure out the "fa-undo" icon at the top restored/undeleted the note. My initial impression was that it was some kind of refresh. 

Some possible ideas: 
- Alternative icons: https://github.com/FortAwesome/Font-Awesome/issues/4090 , https://www.materialui.co/icon/recover
- Adding "Undelete" or "Restore" text beside the icon might help. 
- Moving the restore icon to the right-hand side with the info and trash icon since they are all actions that are taken on the current note. 

This goes along with https://github.com/BoostIO/Boostnote/issues/1311 about hover-text on icons being needed in case icons aren't self-explanatory.

Issue #1311, opened 2017-12-19
Add alt/hover text to UI icons


initializard:
Especially when UI elements are moving around and changing between updates, it would be nice to have alt/hover text when you mouseover icons with a description.

Boostnote icons
![screen shot 2017-12-19 at 1 13 52 pm](https://user-images.githubusercontent.com/2012737/34174555-fd80ecee-e4be-11e7-98c1-45d338d335f9.png)

VS Code hover text example (also shows keyboard shortcuts)
![screen shot 2017-12-19 at 1 17 08 pm](https://user-images.githubusercontent.com/2012737/34174707-70375b06-e4bf-11e7-97ae-b5b3311828a1.png)

PR #1549, opened 2018-02-16
Restore note UX enhancements

nlopin:
It solves issue #1325. I decided not to change the icon, but enhance UX at the Trash view.

1. RestoreButton component was added. The button has tooltip (tiny part of issue #1311)
2. Added "Restore note" item to the context menu in the Trash view


Rokt33r:
I've asked a super minor fix. If you don't have a time to fix this, I'll do it tomorrow!

Issue #1545, opened 2018-02-14
Mark notes by storage/folder color in "All notes" and "Notes in storage" views 


nlopin:
Now we set colors to folders, which affects only the title of preview. 
![image](https://user-images.githubusercontent.com/1469636/36226964-85caba3c-11e0-11e8-9a37-77125f524bf5.png)

But it's much more important to understand the folder of note in the storage view and the storage of note in "All notes" view. 

It would be great if we mark notes with corresponding color. We can color title icon or add thin colored vertical line at the beginning of the block. On screenshot below I colored title icons of the notelist: 
![image](https://user-images.githubusercontent.com/1469636/36227334-d6f04110-11e1-11e8-97e7-81ba3f5099f6.png)

nlopin:
We need to find more elegant solution. The one with icons is too aggressive

pfftdammitchris:
I'm working on this now.

pfftdammitchris:
Sorry I think I misunderstood the post a little bit, but I think I know what you are trying to tell now.
I have a solution and idea for this, i'll proceed it to #1556 if you want me to. I think coloring the icon is a better elegant approach in the storage view.

Coloring the title of the note could be good too but I think if we do that then the icon next to it becomes redundant. What do you think?

Flexo013:
It the latest version of Boostnote the notes in the sidebar have their folder listed like so:
![image](https://user-images.githubusercontent.com/4507015/62027143-3b29e180-b1dd-11e9-8442-185aa6d9fa72.png)

PR #1551, opened 2018-02-17
Storage/folder color labelings in Side nav/Note lists

pfftdammitchris:
References #1545 

Features:
1.  In the "All Notes" default view, each note item in the note list now shows their storage label in color
2. If you go to compressed view it also shows the same thing to the right
3. You are now able to customize a color given to a storage of your choice in the side navigation by clicking the color picker button (next to the + icon) --- choosing a color updates the newly added "color" key in the storage's data storage
4. Added third option for the modal---maybe we can use that to customize the modal in some ways (I used it here to disable the background when the color picker comes up)
5. Color coded folder icons show in the nav list when in storage list view

#### **This is not complete yet** but I wanted to just create a pull request and leave it here while I receive feedback or some suggestions / re-do's etc and continue developing this feature

Does anyone like where this is going or is this something that is totally not ideal to the app?
Some notable things are the modal component (I added in a 3rd prop to the ModalBase component so it now has a configuration to config the modal on the fly on load---maybe we can use that for some interesting things in future developments), the color picker modal (I had to make a separate component as a modal (ColorPickerModal) so that it gets re-used. My first approach kept making duplicate color pickers so I had to move it all the way up front to the Main component---I ended up liking this idea because now the Main component can share the one color picker to the entire component tree as well as the re-usable getColorPicker func to pass around everywhere (look at Main on methods getColorPicker() and ColorPicker())

Todos: 

- [ ] Auto refresh nav list on color change
- [ ] Render color picker next to the cursor
- [ ] Do cooler things

storagecolorization

### DIFF
--- a/browser/components/NoteItem.js
+++ b/browser/components/NoteItem.js
@@ -46,51 +46,69 @@ const TagElementList = (tags) => {
  * @param {Function} handleDragStart
  * @param {string} dateDisplay
  */
-const NoteItem = ({ isActive, note, dateDisplay, handleNoteClick, handleNoteContextMenu, handleDragStart, pathname }) => (
-  <div styleName={isActive
+const NoteItem = ({ isActive, note, dateDisplay, handleNoteClick, handleNoteContextMenu, handleDragStart, pathname, storage, folder, isAllNotes }) => {
+  return (
+    <div styleName={isActive
       ? 'item--active'
       : 'item'
     }
-    key={`${note.storage}-${note.key}`}
-    onClick={e => handleNoteClick(e, `${note.storage}-${note.key}`)}
-    onContextMenu={e => handleNoteContextMenu(e, `${note.storage}-${note.key}`)}
-    onDragStart={e => handleDragStart(e, note)}
-    draggable='true'
-  >
-    <div styleName='item-wrapper'>
-      {note.type === 'SNIPPET_NOTE'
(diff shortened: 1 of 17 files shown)


pfftdammitchris:
Could get some help as well

pfftdammitchris:
Damn on second thought I think it would be a lot more nicer if only the storage's text is colored in the side navigation.... maybe the background is a little too much?

pfftdammitchris:
Going to work on this again in a little bit.
For the mean time, could I get some opinions on my contribution so far?

I'm not sure if I'm wasting my time or if I should stop now (If people disagree and do not like what I am doing, I prefer to be told so so that I can go on ahead and think of other ideas and approaches. I am perfectly fine with anyones opinion)

nlopin:
Thanks for your contribution! It looks pretty cool so far.
In my opinion It's too much color for storage menu item in the left menu. Have you tried to color the font?

pfftdammitchris:
I haven't done that. I'll try it and update the post with the new looks :)
By the way for the "All Notes" view, I have a huge hunch to put a storage icon next to the colored storage labels in the nav list, like the side nav has for the folders. What would everyone think of that? Would an extra icon might be redundant to the eye?

pfftdammitchris:
Im probably just going to scratch this and do small pull requests at a time
I'll re-code this

Issue #1555, opened 2018-02-19
[Feature Request] 'Empty trash' button when right clicking the trash can


patrick-iv:
Did a quick scan at the Issues page but couldn't find this (if there already is a feature request for this, please close this and link the other).

Like the title say, I'd like a "Delete all" or "Empty trash" button when right clicking the Trash tab. I know that it's possible to select all items in the category and right click, delete them but by default I always try to right click the Trash tab first (like you do when emptying the trash can in Windows).

pfftdammitchris:
If no one works on this i'd like to implement this one soon

Rokt33r:
Fixed by #1558.

PR #1558, opened 2018-02-19
Added option Empty Trash

pfftdammitchris:
Feature request from #1555 

I was contemplating to add a window to allow the user to "confirm" the deletion, but I think that wouldn't be needed since the user already technically "confirmed" that decision when deleting a single note beforehand. Apps nowadays don't really give the user any notification of emptying the trash anyways because 99% of the time the user never really empties the trash on accident. nor would the notes being deleted from the trash be of importance anyway, being the reason they got there in the first place

emptytrash


pfftdammitchris:
Resolved #1555

PR #1570, opened 2018-02-22
Fix memory lickage

Rokt33r:
There is no `componentDidUnmount`.


JohnMunsch:
Step One: *Do Not* lick the memory.

Issue #1072, opened 2017-11-05
Journal Mode


rolandjohann:
As said at #1071 this can be a perfect solution for a journaling app, but a feature is missing:

# Journal Mode

This will be a separate view where one can create a journal entry with meta:
- tags
- heading/title
- content

Based on the journal entries the notes will be built. The tags define the notes they belong to, heading defines subheadlines. If the same headings are defined multiple times, the content get appended to that section.

It's worth to think about this and tweak tagging/heading to get a better UX - for example:
A note is likely to have more than two levels of headlines. This structure needs to be mapped to the concept depicted ahead: tagging and heading of the journal entry. Maybe this can be done by hierarchical tags. These can be of special syntax `some tag::and it's headline::second headline`.

At the end of the day it should be possible to throw a journal entry into the app specifying its content, some tags and a heading/title and are built to a collection of notes acting as wiki.

Additional meta information will be stored automatically, like date time, maybe it should be possible to define further ones like `project`. I can imagine that meta like `project` can be mapped into tags and become a separate note for each project...

GaryFurash:
LOVE IT!

markerikson:
Yep, would love to see some kind of journaling support.  Playing around with Boostnote some for the first time, it looks like I could sorta simulate things by giving notes names/first lines like "2017-12-03", and maybe creating categories per month or something.

I've been using https://rednotebook.sourceforge.io/ .  I really like its date-based UI and behavior, but the lack of Markdown support is driving me nuts.  Anything that could help mimic some of that capability would be great.

ttdtrang:
I'm trying to implement some kind of Journal mode. It works something like this
* To create a journal entry, you must be in a **JOURNAL** folder. This way specific requirements of a journal entry is satisfied without interfering with other note collections.
* A journal entry can only be created once for a given date.
* A journal entry is created by clicking on the target date on a calendar.

What's hard to decide is whether to use the automatically recorded creation date as a journal date or to add another date field to record the journal entry date. For example, you are on Monday, Feb 26, but want to write the journal entry for a day in the past, Sat Feb 24, the journal mode should give you a new journal entry for Sat Feb 24, but will annotate it with creation date at Feb 26.

Right now I'm leaning toward keeping these dates separate. Please let me know if you have any thoughts about that.

rolandjohann:
@ttdtrang the question in mind is if we mean the same with our journaling feature. It seems that you would like to have a daily journal.

[...]

Issue #125, opened 2016-09-22
Feature request


ArtemMedvedev:
Thanks for awesome app, exactly what I was looking for.

I might have miss, but for the first look haven't found this in issues. 
It would be cool have possibility to make nested folder structure, adding subfolders.
Also, default syntax for folder and opportunity to edit syntaxes list (hide unused), would accelerate and simplify adding new notes.
And lastly, would be very useful have more options for customizing Finder. Search by tags, change font, theme, sizes, hide sidebar, etc.

Everything else has already been said in the other threads.
I am very happy that found Boostnote, it will be very actively used.

Rokt33r:
Thank you for your feedback.

> It would be cool have possibility to make nested folder structure, adding subfolders.

We need to discuss this issue more. How many notes you have? And, I'll make a new issue about it later.

> default syntax for folder and opportunity to edit syntaxes list (hide unused)

This acutally is actually planned. I think It will be shipped within 2 weeks

> lastly, would be very useful have more options for customizing Finder. Search by tags, change font, theme, sizes, hide sidebar, etc.

It would be great. But, all of problems is how it should be.(For me, designing app is MUCH HARDER than writing code) Could you give me more information here?  #85

ArtemMedvedev:
@Rokt33r Hey! Sorry for delay, a little bit busy this weekend.

> We need to discuss this issue more. How many notes you have? And, I'll make a new issue about it later.

Currently I have approximately 170+ code snippets which I want to keep structured by categories, for a quick example:

JavaScript:
- NodeJS
- ES6
- React
- Electron
- Gulp
- ...

DevOps:
- Nginx
- Docker
- Varnish, Memcached
- ...

AppleScript:
- General
- Apps
- Finder
- Script Debugger
- ...

Shell scripts:
- Bash
- Logs
- Vim
- ZSH, Prompts
- ...

Automation:
- Alfred
- Hazel
- Keyboard Maestro
- ...

and so on.

Also I have about 400+ bookmarks in the [raindrop](https://raindrop.io). I really like it. but it's absolutely not suitable for storage code snippets. I tried to use SnippetsLab, Dash, GistBox, CodeBox, and others. In most cases opportunities are very limited, I can't feel like in the editor/IDE. Even such simple things as make a comment on the usual hotkey, unfortunately almost impossible anywhere, about a deeper customization can only dream of, and usually support with updates always leaves much to be desired.
[...]

Issue #123, opened 2016-09-22
Journal storage(or folder)


Rokt33r:
# Note storage as a Journal

Recently, I needed this feature to record daily note.
But, It has some different characters from common note format.
- It might has no topic
- It might not be able to categorized by folder or tag
- It could be indexed by some moment not the updated time.
## Specs
- Only one note for each day
  - Note will be created automatically by daily basis.
  - If the note has no content, It will be hidden.
- Indexing notes by calendar
- Integrations
  - Calendar(Import all user's event and include it to journal note)
  - Todo List(Written tasks in reminder automatically is sent to user's TodoList app.

PR #1574, opened 2018-02-23
Journal tree

ttdtrang:
Implemented a tree view based on Note creation date. This implementation would form the basis to implement further journal-style note taking functionalities, as in [#1072](https://github.com/BoostIO/Boostnote/issues/1072), without the need to significantly change the current directory layout (as in [#123](https://github.com/BoostIO/Boostnote/issues/123) or [#125](https://github.com/BoostIO/Boostnote/issues/125))
journal-view


ttdtrang:
Submitting another pull request after code linting.

Issue #1578, opened 2018-02-24
Pictures can't move automatically when changing the folder a note belongs to


kikyoulin:
This situation only occurs at moving a note to a folder that is in a different Storage Location.

![problem](https://user-images.githubusercontent.com/30592039/36626181-d83ce72e-1968-11e8-8121-017e8edf9d99.png)

If the two folders are in the same storage location, it shows images normally. If the two folders are not in the same storage location, I need to move these corresponding pictures manually.

Flexo013:
This issue has been fixed. Thanks for reporting it!

PR #1579, opened 2018-02-24
Move images between storages together with note

nlopin:
1. Added new step of moving note's images to `moveNote` function
2. Changed behavior of sidebar navigation after move finished

#1578

P.S. We should refactor moveNote function soon: all operations there are synchronous and we don't process errors at all


Rokt33r:
@nlopin Should we wait for refactoring?

nlopin:
@Rokt33r no, let's do that in separate PR

PR #1612, opened 2018-02-28
Properly decoding for the title of pasted website

kawmra:
# Problem
When we paste the URL, the title of the website is automatically fetched, but if it was encoded by non-utf8 encoding it could not be decoded correctly and it garbled.

# My proposal
Check the `Content-Type` header of the HTTP response when fetching the website from the URL, and if the `charset` is specified, decode response with iconv as that charset before getting title.
If `charset` is not specified or not supported, the response will be decoded as utf-8.

## before

before

## after

after


Rokt33r:
I surprise that yahoo is still using `EUC-JP`... 😱 😱 😱 Thanks for your contribution.

kawmra:
yeah I surprised too.. Thank you for your review!

Issue #1589, opened 2018-02-26
No longer shows check list progress bar


tewe:
In v0.10.0 I can no longer see in the note list how many `- [ ]` I checked off. Downgrading to v0.9.0 helps.

RomanKlopsch:
Which operating system are you using?
And which way did you install the new version?
I can't reproduce the bug on Ubuntu / 17.10 (installed due the deb package)

tewe:
Sorry, macOS 10.13.3, both auto-update and the DMG.

pfftdammitchris:
Working on windows for me. Seems like a macOS issue. Hmm

Rokt33r:
Hmm... that's weird. It works fine for me too. I'm using macOS 10.13.3 too.

tewe:
It does not work fine for you. Here is how 0.9.0 looks, not the progress bar:

<img width="1401" alt="boostnote" src="https://user-images.githubusercontent.com/645412/36740168-ef79deb2-1be1-11e8-8117-f835b3fedab7.png">

RomanKlopsch:
Well, seems like we misunderstood you.
I have to correct myself: It does **not** work for me

RomanKlopsch:
I managed to get the progress-bar back by changing the display type of
>.NoteItem__item-bottom___browser-components-

from *flex* to *block*.
I don't know, whether this tweak breaks something else but it seems to solve the issue.

Rokt33r:
@RomanKlopsch If you don't mind, you can submit a PR with your solution! So, we can review and merge it. How about it? 😄

RomanKlopsch:
I am looking forward to fix this in the afternoon

Rokt33r:
I'll gladly wait for it! ❤️ ❤️

RomanKlopsch:
#1614 here we go...

yosmoc:
@Rokt33r please close this issue.

PR #1614, opened 2018-02-28
Fix missing progress bar in note list

RomanKlopsch:
This solves #1589


pfftdammitchris:
I think the reorderFolder test should be temporarily disabled for now. Its failing half the pull requests

Rokt33r:
Great work!

PR #1621, opened 2018-03-01
Added support for toggling smart quotes in preview

### LINKED ISSUE #1520: Cannot disable smart quotes tranlation on preview screen
I wrote as a markdown source: `curl 'https://www.google.co.jp/'`
then I got in preview: `curl ‘https://www.google.co.jp/’` 

It's just annoying to me. Please make some options to disable it.

rayou:
Fix #1520

`markdown-it` doesn't support options update on the fly, if user changes the smart quotes setting, a restart/reload is required which doesn't seem a good UX to me.

So I made a slightly change to `lib/markdown.js`, I put the initiate processes into a class constructor, which make possible to create a new instance without restart.

Default setting of smart quotes is `enabled`.

mar-02-2018 00-31-40


rayou:
fixed conflicts.

Issue #1518, opened 2018-02-08
[Feature Request] add Nested sub folder support


mehrdad-shokri:
I'm coming from [this](https://github.com/BoostIO/Boostnote/issues/125) feature request,  
as it states there is a WIP for adding sub categorical authoring just like a book support.  
One of the main maintainers (@Rokt33r) stated was working on this feature but there wasn't something announced since then.  
Any updates on this?

Flexo013:
This feature will be continued on #125.

PR #1627, opened 2018-03-02
Journal entry

ttdtrang:
Supported Journal mode as described [here](https://github.com/ttdtrang/Boostnote/releases/tag/v0.1.0)

Inspired by ideas and comments in issue #123, #125, #1072 and #1518. Considering that attachments have been supported, I think journal mode can be supported without nested folder structure, at least for the need of daily journaling. A nested folder feature may still be necessary for other reasons.

boostnote-journal-480


Rokt33r:
@ttdtrang Should we close #1574?

ttdtrang:
I think so. This one includes the previous.

Rokt33r:
@ttdtrang Sorry for making you wait. I'm going to merge this on v0.12.0. Could you fix conflicts again?

ttdtrang:
@Rokt33r Done.

Rokt33r:
I think I should close this. I think we should provide this in a separate app.
Let's discuss this more in the slack! Please join!

markerikson:
Awww, this is a feature I was really looking forward to!  I would still love to see this added.

joshlemer:
Just wondering if this ever got picked up in an other form, either in a new app or the existing one? Cheers!

Flexo013:
@joshlemer There is no designated journal mode.

I am curious though, with the new app, what is stopping you from making a daily note title with the date in a folder for the current year/month? With subfolders and note titles you could achieve a journal right?

markerikson:
Yes, and I've been doing that for years now:

image

However, from an organizational perspective, it would be nice to at least have the folders list turned into a tree so that I could collapse older years.

Also, every time I create a folder for a new month, it's created at the bottom of the list, and I have to manually drag it _all_ the way to the top of the list, which is annoying.

Flexo013:
Yeah, I understand the frustration there. Subfolders in the old app #125 would be a solution to that. Or maybe one day when the new app supports full storage transfer you could switch to that; since that version has subfolders (but lacks some other features).

Issue #1609, opened 2018-02-28
The error or code snippet


shunchuan:
Perhaps it's a bug,you see:

![image](https://user-images.githubusercontent.com/28803269/36777478-938e6992-1ca4-11e8-9ca3-642f4b8eaefb.png)

![image](https://user-images.githubusercontent.com/28803269/36777288-f4050e80-1ca3-11e8-894f-003df6c4be9e.png)
when I save as .html :
![image](https://user-images.githubusercontent.com/28803269/36777356-26af9558-1ca4-11e8-8b84-7838fc9b6c4b.png)

![image](https://user-images.githubusercontent.com/28803269/36777809-caba0a56-1ca5-11e8-895b-60f44baa3852.png)

bimlas:
Seems to be solved - if you agree, please close the issue.

Rokt33r:
Let us know if the bug still exists.

PR #1655, opened 2018-03-11
fixed missing classname in lib/markdown highlight render function

rayou:
Fixed missing classname in lib/markdown due to the same file was modified in multiple commits. Related to #1609


Rokt33r:
Thanks!!

Issue #1672, opened 2018-03-13
HTML tags not rendered anymore?


tlwt:
I think since last update HTML tags are not rendered anymore. is that by design? Currently running Version 0.11.1 (0.11.1) on Mac OS High Sierra 10.13.3 (17D102)

```
<h2>test</h2>
<b>test</b>
<li>test
```

Exporting it to HTML gives this output. All tags are stripped. 
[test.txt](https://github.com/BoostIO/Boostnote/files/1806942/test.txt)

Rokt33r:
https://github.com/jch/html-pipeline/blob/master/lib/html/pipeline/sanitization_filter.rb
Okay, let's make same settings to the one of github. 😄

Rokt33r:
If someone has another idea, let me know.

tlwt:
two additions would be nice - allowing the style attribute ( `<div style="...">`)  to individually format html tags. and the underline `<b>` tag is helpful, as it is not available within markdown to my understanding.

pmood:
If it's fine to allow `style` attr, then style tags could be added as well so that people don't have to copy and paste. I don't know of any XSS that would work in CSS in Chrome, so it *should* be fine.

Rokt33r:
https://www.owasp.org/index.php/XSS_Filter_Evasion_Cheat_Sheet

I've tried all attacks using `style` attribute in the above link. But, none of them works. I assume it should be fine to allow it.

Rokt33r:
@tlwt Done. https://github.com/BoostIO/Boostnote/pull/1677

Rokt33r:
@emrusso I couldn't find any problem from `style` tag. @Redsandro How do you think?

Redsandro:
~~I recommend~~ The _Open Web Application Security Project_ says [style is not safe](https://www.owasp.org/index.php/XSS_(Cross_Site_Scripting)_Prevention_Cheat_Sheet#RULE_.230_-_Never_Insert_Untrusted_Data_Except_in_Allowed_Locations) and should not be used.

pmood:
[...]

PR #1677, opened 2018-03-14
Allow b tag and style attribute

Rokt33r:


browser/lib/markdown.js
```diff
@@ -55,13 +55,14 @@ class Markdown {
...
     this.md.use(sanitize, {
-      allowedTags: ['iframe', 'input',
+      allowedTags: ['iframe', 'input', 'b',
         'h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'h7', 'h8', 'br', 'b', 'i', 'strong', 'em', 'a', 'pre', 'code', 'img', 'tt',
         'div', 'ins', 'del', 'sup', 'sub', 'p', 'ol', 'ul', 'table', 'thead', 'tbody', 'tfoot', 'blockquote',
         'dl', 'dt', 'dd', 'kbd', 'q', 'samp', 'var', 'hr', 'ruby', 'rt', 'rp', 'li', 'tr', 'td', 'th', 's', 'strike', 'summary', 'details'
       ],
       allowedAttributes: {
         '*': [
+          'style',
```
Redsandro: After taking some time to reflect, I feel in my heart that this is very tempting but a bad idea. Please see https://github.com/BoostIO/Boostnote/issues/1672#issuecomment-373236851 where I try to explain that it doesn't matter that you cannot do a known xss. style is a known problematic attribute, and a potent target for _new_ exploits.

PR #1678, opened 2018-03-14
Added unit tests for lib/markdown

rayou:
Added unit tests for the issues in recent releases.


browser/lib/consts.js
```diff
@@ -1,10 +1,9 @@
 const path = require('path')
 const fs = require('sander')
 const { remote } = require('electron')
-const { app } = remote
```
rayou: I couldn't make `app` extracted from `remote` while running tests, following error throws.
```bash
var app = remote.app;
                ^

TypeError: Cannot read property 'app' of undefined
```
Any ideas why this happened?

package.json
```diff
@@ -92,7 +92,7 @@
     "uuid": "^3.2.1"
   },
   "devDependencies": {
-    "ava": "^0.16.0",
+    "ava": "^0.25.0",
```
rayou: updated to 0.25.0 for snapshot support

tests/lib/rc-parser-test.js
```diff
@@ -29,5 +29,5 @@ test('RcParser should return a json object', t => {
 })
 
 function filePath (filename) {
-  return path.join('boostnoterc', filename)
+  return path.join(`${__dirname}/boostnoterc`, filename)
```
rayou: breaking change in ava [v0.17.0](https://github.com/avajs/ava/releases/tag/v0.17.0)

browser/lib/consts.js
Rokt33r: I assume we have to make `remote` Mockable.

tests/helpers/setup-electron-mock.js
```diff
@@ -0,0 +1,11 @@
+import mock from 'mock-require'
+
+const noop = () => {}
+
+mock('electron', {
```
rayou: added mok electron.

browser/lib/consts.js
rayou: As result, `require('electron')` just return the path of electron binary, a string.

I think it's an good idea to mock `electron`, so I created a new mock helper in `tests/helpers/setup-electron-mock.js`, we can add more mocked function into it if needed.

rayou:
As result, `require('electron')` just return the path of electron binary, a string.

I think it's an good idea to mock `electron`, so I created a new mock helper in `tests/helpers/setup-electron-mock.js`, we can add more mocked function into it if needed.

The change in `browser/lib/consts.js` has been reverted.

Rokt33r:
Great work!! 👍 👍 👍

rayou:
fixed conflicts.

PR #1684, opened 2018-03-15
Prevent generating a link tag when pasting URL to inside of a link tag.

### LINKED ISSUE #1680: Prevent fetch the title when paste URL into inside of a link tag
I like the function to fetch the titles when pasting URL, but sometimes I would like to decide a title manually.

I think it makes no sense the title will be fetched automatically even when I paste a URL into inside of link tag manually.

For example, I don't want the title to expand automatically in the following case.

boostnote

How do you think?

And I can submit PR if you like this proposal.

kawmra:
closes #1680

When pasting URL to inside of a link tag, the title not fetched anymore.

## before

boostnote

## after

boostnote2


Rokt33r:
Look awesome!

PR #1691, opened 2018-03-16
Fixed smart quotes when posting a note to blog

rayou:
Hi @lijinglue,

the default smart quote setting of `lib/markdown` was `true` and didn't follow global config in ConfigManager. I have modified `lib/markdown` to use `config.preview.smartQuotes` value as default. 

`markdown` instance was created at the beginning in `NoteList/index.js` and never reloads, when user changes `Smart Quotes` setting, markdown instance in `NoteList/index.js` won't be updated.

So I changed the markdown init behavior to only initiate when it's needed for `Post to Blog` feature.

**Before:**
mar-16-2018 23-19-07

**After:**
mar-16-2018 23-47-59

### DIFF
--- a/browser/lib/markdown.js
+++ b/browser/lib/markdown.js
@@ -8,7 +8,6 @@ import {lastFindInArray} from './utils'
 
 // FIXME We should not depend on global variable.
 const katex = window.katex
-const config = ConfigManager.get()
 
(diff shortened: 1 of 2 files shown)


lijinglue:
Thank you for spotting and fixing this!

Issue #1645, opened 2018-03-07
Ruler [feature request]


gpresland:
Request to add zero or more visible rulers to the editor that show line length, along with the colour for them. This is the same feature found in Atom, Sublime, Visual Studio Code, etc.

e.g.

```javascript
"rulers": [80, 100, 120],
```

rayou:
I am working on this one

bimlas:
It seems implemented, issue may be closed.

PR #1709, opened 2018-03-21
Added editor rulers

rayou:
Feature requested in #1645 

Rulers is disabled by default.

I will add unit tests once @nlopin adds jest package into his commits.

mar-21-2018 22-40-51
Note: Ruler 120 was hidden due to small viewport.


lib/main.html
```diff
@@ -57,6 +57,10 @@
     opacity: 1 !important;
     pointer-events: auto !important;
   }
+  .CodeMirror-ruler {
+    border-left-color: rgba(142, 142, 142, 0.5);
```
rayou: overwrite the default CodeMirror Ruler color.

The default ruler color is #ccc which is too bright in dark themes, so I made this tweak to ensure the ruler fits all kind of themes.

lib/main.html
```diff
@@ -57,6 +57,10 @@
...
+    mix-blend-mode: difference;
```
rayou: overwrite the default CodeMirror Ruler color.

The default ruler color is #ccc which is too bright in dark themes, so I made this tweak to ensure the ruler fits all kind of themes.

Issue #1694, opened 2018-03-18
Notes not rendered anymore


dyge:
Boostnote with a storage on a nas accessible via local Network is not displaying all notes anymore. Just 2 notes are displayed. I see the missing notes in the NAS folder and can access, but they will not be displayed.

The difference is between snipped and markdown mode. Markdown will not be displayed anymore

Rokt33r:
Could you attach a screenshot of the error message? You can found it from devtool.

dyge:
there is no error message, just not displaying older notes.

dyge:
did some additional tests, even adding will show the notes in the actual session, but after restart the notes are not displayed anymore

Rokt33r:
Hmm...🤔 tell me if it happens again. I'll close this issue for now.

dyge:
still not displayed, seems a misunderstanding? finally, i lost my notes in the tool but i´m capable to read the cson files and add to another tool .... Losing data through a tool because of proprietary format is a lesson i learned the hard way .....

Rokt33r:
I'm sorry. I misunderstood. 😢 

Btw, there should be some errors in the console of devtool. Could you submit a screenshot to prove that there are not any errors. I've never seen the case like this.

dyge:
![error](https://user-images.githubusercontent.com/22254452/37585234-b809df6e-2b58-11e8-91b7-2404af19e657.JPG)

Rokt33r:
Yea, those are the errors what I've meant. Open those files and investigate what happened.

myxo:
It's worth mention, that if you still have cson files, you didnt lost your data; cson files have quite simple format, you can restore data by hand (or by write simple script)

dhaskew:
I had this same problem.  For me, if I created notes on my Ubuntu 16.04 system using an older version (0.8.x) then I could see my notes on my mac with the newest (0.11.x) versions.  If I created notes on my mac, then I couldn't see those notes on my Ubuntu system.  Presumably, there was a breaking change in how notes are created in newer versions that is not backwards compatible.

Flexo013:
This issue shouldn't be occurring anymore. If it does happen again then please let us know and we will try our best to deploy a fix.

Issue #1485, opened 2018-01-30
Missing notes after updating


hmojicag:
Not sure how it happened but some of my notes are unlinked.

Here's a screenshot of my boostnote, as you can see boostnote only "opens" 3 notes for the folder "haza"
![image](https://user-images.githubusercontent.com/6140325/35585010-5975706a-05bc-11e8-8a57-4900c447bc39.png)

Here's the hashcode of the folder "haza":
![image](https://user-images.githubusercontent.com/6140325/35585047-765bd3d6-05bc-11e8-8cc5-751cb53f0b9f.png)

But when I search the notes folder looking for my missing notes using the hashcode/key of the folder as search parameter I can see all my notes which must be 6 for this folder, other folders suffer the same issue:
![image](https://user-images.githubusercontent.com/6140325/35585125-a871a9fe-05bc-11e8-967d-af98b71f5066.png)

Why Boostnote is not loading those notes?
What can I do to fix that?

anhhuyalex:
Hello,

I also experience a similar issue. After an update, some notes seem to have disappear from Boostnote. I can't see them, nor can I search them. Nevertheless, everything is linked (all the folders are visible). Furthermore, the .cson files still exist in the linked folder. Most notes remain, but some notes have mysterious disappeared.

ithinuel:
Hi, I had a similar issue solved by Ctrl+R.

Flexo013:
@hmojicag Have you experienced this issue with any of the latest builds of Boosnote?

hmojicag:
Noup, sorry guys after that bug/thing/feature deleted my notes I got to move to a note keeper app that let handle and backup my notes in a simpler way, something that is not tightly coupled with the app itself and more like a simple text/markdown editor.
Thanks anyway.

PR #1737, opened 2018-03-23
remove broken note from note list

yosmoc:
related #1597 #1485 #1694

I also have same behaviour reported in above issues. 

I don't know broken .cson error is made by Boostnote or not, but it is quite scary that notes seems to be disappeared. Investigating the reason why we get the broken the .cson is important, but I think quick fix is more important. 

In current implementation, when .cson is broken and catch error in processing this file,  undefined is collected in notes. If undefined is in the notes, we cannot see the broken note and following notes even they are not broken.

My implementation is just removing undefined from  the note list.

### DIFF
--- a/browser/main/lib/dataApi/resolveStorageNotes.js
+++ b/browser/main/lib/dataApi/resolveStorageNotes.js
@@ -27,8 +27,11 @@ function resolveStorageNotes (storage) {
         data.storage = storage.key
         return data
       } catch (err) {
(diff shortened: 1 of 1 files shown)


Rokt33r:
LGTM! 久しぶりです！

yosmoc:
@Rokt33r Thanks!またちょこちょこ手伝いますね！

Issue #1738, opened 2018-03-24
Linux: Closing the window via GUI doesn‘t close processes


wAuner:
Linux Mint 18.3 Cinnamon
Boostnote 0.11.3

As pointed out in another [thread](https://github.com/BoostIO/Boostnote/issues/1264#issuecomment-375712671) for windows, closing boostnote via the x-button of the window doesn‘t close the processes properly. This prevents the app from repoening after closing. Opening the app again is only possible after manually killing all related processes.
Closing the app via `ctrl+q` works well though.

Rokt33r:
It should be same to #1679 

More than a year ago, I'd been using cinnamon too. At that moment, some people asked me to make the app hide, rather than quit itself, when its window closed. I assume this sounded reasonable because we can easily unhide it from the dock of cinnamon, just like the dock of macOS.

But, this is old story. So, do you want to make the app quit ASAP when you close its window, Right?

@jkbecker @ushell How do you think?

If you guys like it, I'll re-deploy the fixed version on next week.

wAuner:
Ok after testing it again, I can not confirm what you are saying about "unhiding" in cinnamon with boostnote. If I close the app via GUI, there is no icon etc left which I could use to start again.

I'm not sure what you mean by dock, but I tried this with the cinnamon panel and I'm also using planck as a macOS like dock. I have the boostnote icon in my planck dock and it definitively doesn't work like macOS. If I try to start the app again, a window is opened for a split second and then it is gone.

So in its current form I'd argue that this should be handled as a defect. I don't see a reason why the "gui-quit" should be different from `ctrl+q`. This may make sense in macOS where this is the default behavior for many apps, but this is not true for Linux (at least none distro that I've ever used).

Rokt33r:
> but this is not true for Linux (at least none distro that I've ever used).

Okay, let's make it simple. Do you want to try to fix this? Modifying a single line should be enough.
https://github.com/BoostIO/Boostnote/blob/master/lib/main-window.js#L43

Rokt33r:
If you don't have time to do, I'll fix this week anyway. Just let me know if you could afford.

wAuner:
ok I've made an attempt to fix it and sent a pull request. Hope that works. I've never had any contact with javascript before though.
Keep up the good work, I really like the app!

jkbecker:
[...]

PR #1741, opened 2018-03-24
fix incorrect app closing for linux mint cinnamon

wAuner:
excluded the cinnamon desktop for the macOS-like behavior which closes the window, but does not shutdown the app processes as discussed in #1738 with @Rokt33r


jkbecker:
looks right to me

PR #1742, opened 2018-03-24
Add selector to sort tags by counter or alphabetically

bimlas:
screencast

### DIFF
--- a/browser/main/SideNav/SideNav.styl
+++ b/browser/main/SideNav/SideNav.styl
@@ -8,6 +8,29 @@
   display: flex
   flex-direction column
 
+.control
+  user-select none
+  height $control-height
+  font-size 12px
+  line-height 25px
+  display flex
+  color $ui-inactive-text-color
+
+.control-sortTagsBy
+  flex 1
+  padding-left 15px
+
+.control-sortTagsBy-select
+  appearance: none;
+  margin-left 5px
+  color $ui-inactive-text-color
+  padding 0
+  border none
+  background-color transparent
+  outline none
+  cursor pointer
+  font-size 12px
+
 .top
   padding-bottom 15px
 
@@ -82,6 +105,10 @@ body[data-theme="white"]
     background-color #f9f9f9
     color $ui-text-color
 
+  .control
+    background-color $ui-white-backgroundColor
+    border-color $ui-white-borderColor
+
 body[data-theme="dark"]
   .root, .root--folded
     border-right 1px solid $ui-dark-borderColor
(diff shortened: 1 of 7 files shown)


Rokt33r:
I like this feature. And, I want to suggest a little layout change.

Current layout:
```
Tags
↓alphabetically
#a
#b
```

Suggestion:
```
Tags ↓alphabetically
#a
#b
```

bimlas:
It would be OK, but I'm working on another feature related to tags which needs control buttons, thus I thougth that the same layout can be used as the control of note list.

Rokt33r:
What kind of buttons are you going to add for tags?

bimlas:
I like to narrow down the tags to **related** ones.

Example:

* string (9)
* integer (5)
* array (4)
* print (2)
* length (1)

When you select a tag by clicking on it (after you enabled this feature on the "control bar"), the related tags geting listed. For example if I select the `string` tag, the other tags of the related entries geting listed, like:

* print (1)
* length (1)

The counter now shows the number of entries with `string` **and** `print`/`length`. If you want to get the full list back, then you can do this by clickin on the "reset" button on the "control bar" (or a dedicated "All" tag - I'm not sure which one is better).

It turns the tag list in to some kind of "dynamic table of contents". This behaviour is exists in CintaNotes and PiggyDB thus I think a lot of users would migrate to Boostnote.

I cannot find the PiggyDB documentation, thus I recommend to try it out on http://piggydb.jp/sandbox/home.htm (guest/guest):

* Select a tag from the tag palette ([python](http://piggydb.jp/sandbox/tag.htm?id=357) for example)
image
* Clicking on it show the related tags - select the plus sign near Ruby to get "notes" tagged with Python **and** Ruby
image


> Clicking on a tag activates it, and only notes having this tag will be displayed. Using Ctrl+Click or Shift+Click, you can expand the tag selection to additional tags. Keep in mind that with Ctrl+Click, a tag intersection is used when multiple tags are selected (i.e. notes having all of the selected tags will be included)
[CintaNotes manual - Using the Tag Sidebar](http://cintanotes.com/help/#managing)

To get the most out of this:

> I recommend that you keep individual fragments (notes) relatively small. Smaller fragments, which focus on single issues, give you more flexibility when you organize and relate them using the tags.
[PiggyDB manual - Knowledge Fragments](http://piggydb.jp/en/tag.htm?id=12)

bimlas:
I thought about it again, here's a concept:

slice_1

The plus sings used to narrow down the list (as I said above), "All tags" resets the narrowing.

One drawback about placing the order selector near to "Tags" title: since i18n is supported, it may be take too much space (for example "Tags" could be "Abcdef ghijklm", "Alphabetically" (order) could be "Nop qrstuv wxyz whatever"), therefore it may be hidden if the sidebar is too strait (as it works on note list's control bar).

bimlas:
Moved the selector next to the title:

browser/main/SideNav/SideNav.styl
```diff
@@ -53,11 +30,33 @@
...
-.tag-title
-  padding-left 15px
-  padding-bottom 13px
-  p
-    color $ui-button-default-color
+.tag-control
+  display flex
+  height 30px
+  line-height 25px
+  overflow hidden
```
bimlas: This is the correct way  to hide the selector when the sidebar is too narrow?

bimlas: Is this the correct way to hide the selector when the sidebar is too narrow?

Rokt33r: I think so. :)

PR #1751, opened 2018-03-25
Add folding to CodeEditor

gediminasml:
This PR adds (much needed 😄) Code folding functionality to `code editor` by passing additional `options` to the `CodeMirror`

Folding


msbentley:
This looks great, but are there limitations in which languages are supported? For example it works fine for json, but not at all for python (v0.11.4 on ubuntu). Thanks!

gediminasml:
@msbentley I suppose we should refer to [CodeMirror](https://github.com/codemirror/CodeMirror) documentation 🙂

msbentley:
If I compare to the examples given at: https://codemirror.net/demo/folding.html then I get:

- python - no folding indicators shown
- HTML - only one indicator shown (line 7)
- JavaScript - no indicators shown

PR #1752, opened 2018-03-25
Add jest and simple test

nlopin:
I've added Jest to the project. 

There is an issue when you test components with external js libraries. For example, `MarkdownPreview` works with `Raphael` and its' extensions to draw graphs. I haven't figured out yet how to mock these stuff properly.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -5,7 +5,7 @@
       "presets": ["react-hmre"]
     },
     "test": {
-      "presets": ["react", "es2015"],
+      "presets": ["env" ,"react", "es2015"],
       "plugins": [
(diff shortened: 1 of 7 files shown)


nlopin:
One big issue I'm thinking about. Jest doesn't work with `refs` properties, but `MarkdownPreview` is the wrapper around iframe and we interact with that iframe via refs.

flaviusone:
> Jest doesn't work with refs properties

Can you detail this? Maybe you are referring to `react-test-renderer` and [ref mocking](https://reactjs.org/blog/2016/11/16/react-v15.4.0.html#mocking-refs-for-snapshot-testing)?

Did you consider adding enzyme?

Great job for adding Jest. It's a big and important step for this project! 🎖

PR #1807, opened 2018-04-13
Edited theme and language init code to avoid code repetition

ZeroX-DG:
Hi Boostnote team,
Thanks for this amazing app. I noticed some code repetitions that can be avoided, I don't know if they was there on purpose or by accident. If there is a reason for them to stay there then please tell me why, maybe I can learn something from this :smile:


kazup01:
@ZeroX-DG  Thank you for your contribution 🎉

Issue #1822, opened 2018-04-17
Boostnote paste screenshot doesn't work properly


ZeroX-DG:
# Current behavior

When I paste my screenshot into boostnote the image won't display unless I trigger the mode
![peek 2018-04-17 20-00](https://user-images.githubusercontent.com/12984316/38871095-0af0898e-427a-11e8-901f-fae7a983f82c.gif)

# Expected behavior

The screenshot display immediately

# Steps to reproduce

1. take screenshot
2. paste into boostnote

# Environment

- Version : 0.11.4
- OS Version and name : Linux Mint 18.2 Cinnamon 64 bit

ZeroX-DG:
I've came up with a fix, PR is on its way

ZeroX-DG:
@ehhc Sorry for asking this but English isn't my first language so I'm not quite understand your question :smile: .  What do you mean by "I hope to integrate it in my one though" ? Is it means you have a PR for the drag&drop case and you want to intergrate my fix in yours? If that's what you mean then what do you want me to do?

ZeroX-DG:
BTW, I also have a fix for the drag&drop case on my mind :smile: so if you already fixed then please tell me so that I won't have to code it :smile:

ehhc:
Hi as you see in #1825 I'm currently trying to refactor the whole attachment management. E.g. i would like to have it in one class that is responsible for that. With my previous comment i meant that i might fix your bug through my refactoring as well :)

ZeroX-DG:
Oh so should I withdraw my PR?

ehhc:
i'm not the owner of the project so i don't know. Since you already have a solution i think you should just wait for your PR to be merged. It might take some time till my fix is ready.. :)

ZeroX-DG:
Ok good luck with your PR :smile:

ehhc:
@ZeroX-DG I've submitted a lot of my refactoring by now :) Hope it will be accepted :) I've fixed this issue as well :)

PR #1824, opened 2018-04-17
Fixed image not displaying after paste

ZeroX-DG:
This PR is a fix for this issue: https://github.com/BoostIO/Boostnote/issues/1822


kazup01:
Thanks @ZeroX-DG :)

PR #1836, opened 2018-04-18
Updated tag note count to float to the right like the storage list

bah2830:
If a tag has a number in the name or there are a lot of notes with a tag the current count of notes in a tag can be confusing with a single space between the tag name and the count.

This update borrows the styling from the storage item to float the count to the right side of the sidebar.


kazup01:
Merged. Thanks @bah2830 😄

PR #1852, opened 2018-04-21
Fixes #1825 Refactoring of the attachment/image management

### LINKED ISSUE #1825: Add attachment to note
In #1723 i asked for the feature to include attachments in nodes. I've submitted a really ugly quick fix. Unfortunately there are still some points:

1)	Why are the dropped files named with Math.random().toString(36).slice(-16) (CopyImages.js) instead of something more resilient like Hash(Timestamp + filename) ? Math.random() might return the same result again. Especially when not properly initialized! 
2)	Why are there (at various places) different definitions of markdown-templates to include images? I’ve found at least 3 without being able to figure out why they aren’t at one central place and in what cases they are used 
3)	Why are there multiple definitions of the image folder path? e.g. moveNote.js:77, exportNote.js:9, copyImage.js:22 and none global definition? A global reliable definition of this constant variable is missing. 
4)	No documentation (e.g. JavaScript Doc) is provided. You have to guess what a method does and why it is necessary
My quick fix hasn’t fix these issues.
**I’m working on them at the moment and will submit a more suitable solution as soon as possible.**

ehhc:


ehhc:
I have no idea why the tests are failing. Locally 'yarn jest' finishes successfully..

ehhc:
@Rokt33r can you have a look at this? It was a lot of work and since some other issues are related to the attachment management it might be worth to review this pull request soon so that none work is wasted for issues allready solved.. at least i think so :)

Rokt33r:
@ehhc I'll do it today. :)

PR #1857, opened 2018-04-21
Issue 1827

ehhc:


ehhc:
This change is based on the refactoring done in https://github.com/BoostIO/Boostnote/pull/1852

Rokt33r:
Could you create this issue again? Showing diff of github seems to be broken. And, it would be nice you can fix the lint errors.

PR #1858, opened 2018-04-23
Fix #1797: In preview window, open external link in external browser.

### LINKED ISSUE #1797: Don't open external links in the preview window
OS : Windows 10 64 bits.

Since version 0.11.4(2018/4/11), when clicking on a link from preview, the link is loaded into the preview webview.

image

When switching to another note **into the same folder**, the note is displayed with styles included from the loaded website.

image

Previewing a note contained in another folder obviously resets stylesheets to the default appearance.

Links should be opened with the OS default browser instead.

zzdjk6:


Rokt33r:
LGTM!

PR #1875, opened 2018-04-27
Monokai theme

yougotwill:
This has been long overdue.

I'm planning on looking into custom themes otherwise I'll make a poll on slack for 3 more themes that we could add. 😄 

monokai

**Note** 
With the current way that themes are set up the codebase will become to bloated if we add a whole bunch of themes. 
I'll try look into either making an adaptive theme based on the currently set syntax theme or allow people to select specific colors that the theme will use from the settings.


kazup01:
Merged. Thank you for your amazing contribution! 
And your plan is sound great. I can not waiting for it 😄

Issue #1788, opened 2018-04-09
After moving the note from one folder to other, the images attached in the note automatically deleted 


lkmokadam:
After moving the note from one folder to other, the images attached in the note automatically deleted. Fortunately, i am using Dropbox location and restored the images. 

OS: Ubuntu 16.04
Version using : Boostnote 0.11.2

davidmigloz:
I'm also experiencing this issue in Mac with Boostnote 0.11.3.

ShurikuS57:
I have the exact same problem on linux.

Abracakebabra:
+1. 

This is a major problem. Leads to data loss.
(Seems not to be a problem for all file formats, e.g. .pdf)

ehhc:
I've submitted a fix for the problem based on the refactoring of the attachment management -> https://github.com/BoostIO/Boostnote/pull/1860

PR #1876, opened 2018-04-27
Moving notes with attachments

ehhc:
Fix for #1788 which dosen't suffers from the limitation mentioned in https://github.com/BoostIO/Boostnote/pull/1805

New pull request due to changes in master (@Rokt33r )


Rokt33r:
I think content should not change after moving. But, it replaces links of attachments.

Rokt33r:
I think we need to discuss this. Is it possible for you to join our slack channel?

PR #1890, opened 2018-05-06
Fixed print font color in dark theme

BerniWittmann:
When using the printing option in dark theme the font color is grey and not perfectly readable. So i just tweaked the styling a little.

Before:
before

After:
after


BerniWittmann:
Sorry just saw that this was fixd in the meantime while I worked on this.

PR #1923, opened 2018-05-15
Move note with attachment to different storage Fix for #1788

ehhc:
@Rokt33r we had a slack discussion that it is neccessary to change the note content since the noteKey is changed during the move and the attachments should be in the correct folder.
We discussed whether it is possible to simply remove this new-creation of the note key. I tried it but using the old noteKey leads to some strange issues when the app tries show the moved note. So i guess that's the reason why you initially included the new-key-feature. -> it can't be removed.

We agreed that it's a good idea to have the attachements in a folder matching the noteKey. So this PR is again needed for moving a note with its attachment.

You wished for a PR that you can auto-merge.
Here it is.


browser/main/lib/dataApi/attachmentManagement.js
```diff
@@ -3,6 +3,7 @@ const fs = require('fs')
 const path = require('path')
 const findStorage = require('browser/lib/findStorage')
 const mdurl = require('mdurl')
+const fse = require('fs-extra')
```
Rokt33r: Why do we need `fs-extra`? I assume `sander` can do the same job.

ehhc: I've used fs-extra because it has a good move-directory function. As far as i know sanders is not able to do so.. I think it contains some fallback implementation if the quick/efficient ways don't work..

PR #1932, opened 2018-05-18
Refactoring of store.js

bimlas:
I'm not sure, but the original logic seemed buggy to me: the `uniqueKey` was not added to the state (see the patch).

### DIFF
--- a/browser/main/store.js
+++ b/browser/main/store.js
@@ -87,21 +87,7 @@ function data (state = defaultDataMap(), action) {
         }
 
         // Update foldermap if folder changed or post created
-        if (oldNote == null || oldNote.folder !== note.folder) {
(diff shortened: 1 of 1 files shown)


ehhc:
Why is the same code used in so much different places without extracting it to a method/function?

bimlas:
You are right, I will refactor it.

bimlas:
Just realized that there is no problem with the original logic, but there are a LOT of duplicates in _store.js_. Refactoring in progress, but please notify me if I'm doing it wrong.

bimlas:
I think I'm finished: I did what I can.

bimlas:
@kazup01, please change to 'in review' state, I'm finished.

Issue #1893, opened 2018-05-09
Double quotes rendered as encoded string within MD frames in HTML exported notes


Bubka:
Hello,
The PR #1753 has broken correct display of double quotes in MarkDown frames (4 spaces as line prefix)

# Current behavior
The exported HTML
![image](https://user-images.githubusercontent.com/858858/38861585-33bce412-4233-11e8-866a-2bc1eaa320fb.png)

# Expected behavior
The exported HTML
![image](https://user-images.githubusercontent.com/858858/38861534-12263826-4233-11e8-8f45-39d81f0d0185.png)
(like it used to work before PR #1753)

# Steps to reproduce

1. Create new note
2. Type in
```
    "}, "2":{"
```
3. Export the note as HTML

# Environment

- Version : 0.11.4
- OS Version and name : Win 7 64b

IssuehuntBot:
@crowdy has started working. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/1893)

IssuehuntBot:
@kazup01 has boosted this issue with $20. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/1893)

IssuehuntBot:
@zerox-dg has started working. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/1893)

IssuehuntBot:
@zerox-dg has submitted [output](https://github.com/BoostIO/Boostnote/pull/1935). [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/1893)

IssuehuntBot:
@kazup01 has rewarded. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/1893)

Issue #1728, opened 2018-03-23
Tags not escape on exporting to .html


hos2002820:
If my content has tags something like
```
<br>
```
or
```
<VirtualHost *:80>
</VirtualHost>
```

they will not show on exported .html

nlopin:
@Rokt33r put it on me

Rokt33r:
@nlopin Gladly. 👍 👍 👍

Bubka:
Hello,
The PR #1753 broke correct display of quotes in Mark Down Frames (4 spaces as line prefix)

my MD code:
```
    "}, "2":{"
```

The html exported code is:
```
<code>
    &quot;}, &quot;2&quot;:{&quot;
</code>
```

So in the browser it's looks like:
![image](https://user-images.githubusercontent.com/858858/38861585-33bce412-4233-11e8-866a-2bc1eaa320fb.png)

instead of:
![image](https://user-images.githubusercontent.com/858858/38861534-12263826-4233-11e8-8f45-39d81f0d0185.png)
like it used to work

PR #1935, opened 2018-05-19
Allow customizing html escape when export note

ZeroX-DG:
When exporting note to html they are escaped before export, if user doesn't want to escape it (https://github.com/BoostIo/Boostnote/issues/1893) Then they can customize it in export tab in setting.

image


ZeroX-DG:
Well, good news ! it's still works.
image

However, the old bug "image-not-show-when-drop" occurred again (https://github.com/BoostIO/Boostnote/issues/1822). And of course the image displayed again when I toggle the mode.

ehhc:
@ZeroX-DG to be honest, with my refactoring i broke your fix again..
i also opened a PR for it but it was closed "since it is already fixed".. -.-

You could simply insert another key then the preview should work as expected .. :)

ehhc:
conserning your fix. I would vote against giving the option to disable the html escaping since it is a very very important security feature.
Btw: the problem of https://github.com/BoostIO/Boostnote/issues/1893 is *not* that the html get escaped but that it is down in a wrong way.

input:
`    "}, "2":{"`

expected:
`<pre><code>&quot;}, &quot;2&quot;:{&quot; </code></pre>`

acutal:
`<pre><code>&amp;quot;}, &amp;quot;2&amp;quot;:{&amp;quot;</code></pre>`

as you see the problem is that the `&` is wrongly escaped as well.. that's the problem.. not the escaping itself

ZeroX-DG:
@ehhc  Oh, I see the problem there! thank you. Hmm...I also think that allow unescaped html is pretty dangerous, but I want to keep the option there to increase the "customizability" of boostnote, beside, the html is escaped by default, if the user choose to turn that feature of, he/she probably knows what he/she is doing (I hope so :smile: ). Anyway, I'll try to investigate the "double-escaped-html" issue and submit a fix right here.

ehhc:
Even though i understand your argument about customization, i think giving users the possibility to do potential harmful things seems not to be a good idea in my opinion..

ZeroX-DG:
@ehhc I fixed the double escaped html. We gonna need to vote whether the option should be there or not :smile: @kazup01  What do you think?

ehhc:
and i think we should ask @Rokt33r  as well

browser/main/modals/PreferencesModal/Export.js
```diff
@@ -0,0 +1,120 @@
...
+  clearMessage () {
+    _.debounce(() => {
+      this.setState({
+        UiAlert: null
+      })
+    }, 2000)()
+  }
+
+  render () {
+    const ExportAlert = this.state.ExportAlert
```
sosukesuzuki: Please use split assignment: like below
```
const { ExportAlert } = this.state
```

browser/main/modals/PreferencesModal/Export.js
```diff
@@ -0,0 +1,120 @@
...
+    const ExportAlertElement = ExportAlert != null
```
sosukesuzuki: I think that it is better than now, you don't declare the variable `ExportAlertElement`.
I think you should writer like below on [this line](https://github.com/BoostIO/Boostnote/pull/1935/files#diff-5411ae75aacdd9d44387d3d6712ffdf3R112):
```jsx
{
   ExportAlert != null &&
    <p className={`alert ${ExportAlert.type}`}>
       {ExportAlert.message}
    </p>
}
```

browser/lib/utils.js
```diff
@@ -15,43 +15,12 @@ export function escapeHtmlCharacters (text) {
...
-    }
-
-    lastIndex = index + 1
-    html += escape
-  }
-
-  return lastIndex !== index
-      ? html + str.substring(lastIndex, index)
-      : html
+  return str
```
nlopin: Have you tested the speed between the current and new version of the function?

From my point of view, 6 replaces is the slower solution than previous. Instead of searching the whole text (which can be big) only once it is done 6 times in a row.

ZeroX-DG: Yes, I've noticed that and discovered that your version seem to work fine and not the source of the problem as when I revert to your version and change the escape for the `&` char to `'&ampssssss;'` The html exported still use `&amp;` which is quite wierd and I'll re-investigate the problem.

ZeroX-DG:
@nlopin @Rokt33r @ehhc , The problem is more complicated than I thought.
- If we don't escape html -> markdown it will filter non-html tag which causing this problem: https://github.com/BoostIO/Boostnote/issues/1728
- If we escape html -> tags will be escaped and can be render in html but if escaped strings is a part of a code block -> markdown it will escape everything in code block -> double escaped strings, which is this problem: https://github.com/BoostIO/Boostnote/issues/1893
So, I have to make a new escape html function that can detect code block and skip all characters needed to be escape that located in the code block.

Sorry for the delay, I almost forgot about this PR :smile:

ZeroX-DG:
After reading the issue: https://github.com/BoostIO/Boostnote/issues/1728, I think the option for customizing whether the html should be escape or not is redundant so I won't add it into the code.

Rokt33r:
@rokt33r After merging, a sanitizing method for code fences must be replaced with this escape method for better performance. #2191 
(this is a reminder for me)

PR #1942, opened 2018-05-20
Extracting attachment management method to distinct files -> fixes #1940

### LINKED ISSUE #1940: [Refactoring] attachment management
In the attachment management file are multiple functions.
That makes it hard to mock special functions

it would be better if the functions would be located in unique files

ehhc:
It would be better if the attachment management method would be in distinct methods.
That would simplfy the implementation/testing of new features connected with the attachments..

@Rokt33r  sorry for changing so much again..


ehhc:
I will close this issue since it think i caused enough trouble with my last refactoring.. 
I think i should wait and fix issues in that before doing so much refactoring again.. 
AND you won't be able to automerge this at the moment since there are other PR using the all-in-one-file-code of the attachment management

What do you think about this? @Rokt33r  @kazup01 ?

PR #1987, opened 2018-05-27
[Feature Request] - 1638 - Support for admonitions

johannbre:
Added support for markup admonitions by including markdown-it-admonition. Added material icon support and updates to styles for the rendering of six different types of admonitions.

admonitions

admonitions


johannbre:
Committed and pushed

johannbre:
Committed and pushed

TobseF:
Doesn't work on Windows if the **Material Icon** font isn't installed.
https://github.com/BoostIO/Boostnote/issues/2083

PR #1993, opened 2018-05-28
Include markdown-it-smartArrow. Toggle it via settings and disable it…

ehhc:
… by default since it might affect HTML comments in markdown


ehhc:
@kazup01 the merge issues are resolved :)

PR #1996, opened 2018-05-29
Added custom markdown css

ZeroX-DG:
With this feature, users can now customize how the markdown should look like. More info: https://issuehunt.io/repos/53266139/issues/184

Here is how it looks
peek 2018-05-29 11-44

As you can see, the user can choose to use custom css or not by clicking on the allow custom css checkbox in the preview section and then specify the custom css in the code editor bellow.

By default the custom css is disabled.

If the user export the note as html, their custom style will be applied too.

### DIFF
--- a/browser/main/modals/PreferencesModal/UiTab.js
+++ b/browser/main/modals/PreferencesModal/UiTab.js
@@ -29,7 +29,7 @@ class UiTab extends React.Component {
   componentDidMount () {
     CodeMirror.autoLoadMode(this.codeMirrorInstance.getCodeMirror(), 'javascript')
     CodeMirror.autoLoadMode(this.customCSSCM.getCodeMirror(), 'css')
-    this.customCSSCM.getCodeMirror().setSize(null, '300px')
+    this.customCSSCM.getCodeMirror().setSize(null, '250px')
     this.handleSettingDone = () => {
       this.setState({UiAlert: {
         type: 'success',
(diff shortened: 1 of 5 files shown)


browser/main/modals/PreferencesModal/UiTab.js
```diff
@@ -569,6 +574,24 @@ class UiTab extends React.Component {
...
+              />&nbsp;
+              {i18n.__('Allow custom CSS for preview')}
+            </label>
+          </div>
+          <div styleName='group-section'>
+            <div styleName='group-section-label'>
+              {i18n.__('Custom CSS')}
+            </div>
+            <div styleName='group-section-control'>
+              <ReactCodeMirror onChange={e => this.handleUIChange(e)} ref={e => (this.customCSSCM = e)} value={config.preview.customCSS} options={{ lineNumbers: true, mode: 'css', theme: codemirrorTheme }} />
```
sosukesuzuki: This `ReactCodeMirror` is too huge...
Set the width of  this according to [the editor](https://github.com/BoostIO/Boostnote/blob/master/browser/main/modals/PreferencesModal/UiTab.js#L260).

ZeroX-DG:
@sosukesuzuki  I have changed my code, please have a look :smiley:

J-F-Liu:
Suggest define Custom CSS as multiple theme, and user can associate different notes with different CSS themes.

PR #2016, opened 2018-05-31
Markdown Admonitions: Enhanced coloring and rendering for icons.

johannbre:
Markdown Admonitions: Enhanced coloring 
and rendering for icons.
Bug Fix: Fixed a tiny issue with fonts not loading properly on windows when running in dev mode.

admonitions


package.json
```diff
@@ -98,7 +98,8 @@
     "superagent": "^1.2.0",
     "superagent-promise": "^1.0.3",
     "unique-slug": "2.0.0",
-    "uuid": "^3.2.1"
+    "uuid": "^3.2.1",
+    "yarn": "^1.7.0"
```
Rokt33r: You **must not** install yarn locally. Please uninstall it.

johannbre:
Removed the yarn dependency

browser/components/MarkdownPreview.js
```diff
@@ -26,7 +26,7 @@ const dialog = remote.dialog
 const markdownStyle = require('!!css!stylus?sourceMap!./markdown.styl')[0][1]
 const appPath = 'file://' + (process.env.NODE_ENV === 'production'
   ? app.getAppPath()
-  : path.resolve())
+  : path.resolve().replace(/\\/g, '/'))
```
ehhc: What is this doing? May it cause problems on windows maschines, where there is another separator?

johannbre: This is to solve a problem on windows.  This replaces all back-slashes (\\) with forward-slashes (/). Without this, in dev mode, on windows, it's broken and some of the fonts do not get loaded. On windows, you end up with a mixture of slashes since the appPath starts with `file://`, but `path.resolve()` obviously returns windows style paths containing back slashes.

ehhc: Okay, Thanks for the information :)

johannbre:
I've made the changes as requested. Please let me know if there is something else I need to change?

PR #2033, opened 2018-06-05
fixed disappearing scroll bar

ZeroX-DG:
The disappearing scroll bar bug is extremely annoying and has been reported here: https://issuehunt.io/repos/53266139/issues/1632

After working on my personal project, I have found a fix for this issue.

This is bug exists due to the style of the scroll bar of codemirror, instead of applying for just codemirror, it applied to the whole window, everything that has its scroll bar unstyled. So the solution is to override that style!

### DIFF
--- a/browser/components/MarkdownPreview.js
+++ b/browser/components/MarkdownPreview.js
@@ -127,10 +127,6 @@ body p {
 }
 
 const scrollBarStyle = `
-::-webkit-scrollbar-track {
-  border-radius: 10px;
-}
-
 ::-webkit-scrollbar {
   width: 12px;
 }
--- a/browser/components/MarkdownPreview.js
+++ b/browser/components/MarkdownPreview.js
@@ -132,7 +132,15 @@ const scrollBarStyle = `
 }
 
 ::-webkit-scrollbar-thumb {
-  box-shadow: inset 0 0 6px rgba(0, 0, 0, 0.3);
+  background-color: rgba(0, 0, 0, 0.15);
+}
+`
+const scrollBarDarkStyle = `
+::-webkit-scrollbar {
+  width: 12px;
+}
+
+::-webkit-scrollbar-thumb {
   background-color: rgba(0, 0, 0, 0.3);
 }
 `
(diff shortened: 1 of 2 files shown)


Rokt33r:
@ZeroX-DG Could you attach a screenshot after the fix?

ZeroX-DG:
No problem.

BEFORE:

peek 2018-06-08 18-13

AFTER:

peek 2018-06-08 18-15

ZeroX-DG:
@sosukesuzuki As you can see the current scroll bar will disappear after a short period of time and only reappear if you resize the app or focus on a text box and hit the page up or page down key. The new one make sure that the scroll bar will never disappear.

sosukesuzuki:
@ZeroX-DG I know. However, I think that we are not in trouble. Why are you troubled?

ZeroX-DG:
I started to think this is not related to codemirror but instead electron? I've removed codemirror scroll addon and the problem is still persist and also I've found the same issue on Atom github page: https://github.com/atom/atom/issues/15107 , but it seem that their solution is just like me. But how can you guys don't experience this issue? I experienced this issue on both Boostnote and my side project which is fixed by using the traditional way to create scroll bar.

Rokt33r:
Hmm @sosukesuzuki For me, it looks a problem to be fixed. I'll review this.

Rokt33r:
@ZeroX-DG How can I reproduce this problem? It works fine for me.
screen shot 2018-06-27 at 14 55 28

ZeroX-DG:
Hey @Rokt33r , why your scroll bar is the default scroll bar of chrome? Mine is not! 
Here is how you can reproduce the problem:
1. Open something long enough to make the scroll bar appears (the interface setting tab for example)
2. Wait for it to disappear
peek 2018-06-27 13-00

Rokt33r:
I'll try again.

Rokt33r:
Hmm.. mine never disappear... Do you have any clue why it doesn't happen to me? Btw, I didn't configure my scroll. I assume the reason might be that I'm running windows via Parallels.

ZeroX-DG:
I have no idea what's happening but I think if your scroll bar is still the default one then it will not disappear, do you have a linux machine to test it? Maybe it's a linux problem?

ZeroX-DG:
@Rokt33r I've just updated to the latest version of Boostnote (0.11.7) and the scroll bars became the default scroll bar of chrome (just like what you have) and it doesn't disappear anymore??? I gonna have to investigate the commit that change the scroll bar, but that's also means the issue is now solved, the scroll bar doesn't disappear anymore, now it just looks ugly and that's a different issue. Should we close this pull request?

Rokt33r:
Close this issue because the problem is suddenly gone after v0.11.7.

Rokt33r:
L

PR #2057, opened 2018-06-09
Add missing Japanese translations

kawmra:
I add some missing Japanese translations.

But I couldn't translate `Disable Direct Write (It will be applied after restarting)` because I didn't understand what `Direct Write` means.
(And I couldn't find code that uses `disableDirectWrite`.)


kazup01:
Merged. Thank you @kawmra for your contribution!

PR #2060, opened 2018-06-09
[Dark Theme]: Fixes the text color for an active note with an empty title

yougotwill:
This was bugging me so I fixed it :)

Before:
screen shot 2018-06-09 at 2 35 48 pm

After:
screen shot 2018-06-09 at 2 35 12 pm


kazup01:
Merged. I didn't notice this bug, thank you @yougotwill :)

PR #2077, opened 2018-06-13
When printing

Matts966:


Matts966:
Sorry, I forgot to write a description...

I would like to remove the padding for scrolling preview when printing notes(or make pdf) , because the padding make the last page of pdf white one.

Also, when printing, I would like to see code even if a line of it is too long to see without scrolling. Pdf docs and papers can not be scrolled, so I break the line when it is long.

browser/components/MarkdownPreview.js
```diff
@@ -125,6 +125,12 @@ body p {
   .clipboardButton {
     display: none
   }
+  pre.CodeMirror>code {
+    white-space: pre-wrap;
```
sosukesuzuki: please add `word-wrap: break-word;`.
Now, output PDF is broken like below:

Matts966:
Thank you for your review!
It may take time for me to fix it, because I don't know what is best for this problem...

I have two ideas about this.
1. adding `<br>` tag to the number of code line.
2. using other codemirror type? like CodeEditor.

Sorry for my confirmation. 
Thank you.

Matts966:
Also, it is not good to break DOM structure and I reverted the commit.

sosukesuzuki:
@Matts966 I think the second one of your ideas is great:+1:
We should change DOM structure of code block in preview like CodeEditor.

Rokt33r:
@matts966 @sosukesuzuki Could you tell me the current status of this pr?

Matts966:
Sorry for my delaying.

Matts966:
Sorry for my being late and a mistake to close pr.
I'm currently so busy with my work, and the repest will not be satisfied soon.
The modification causing a bug was already reverted.

Issue #2101, opened 2018-06-18
MAC OS - boostnote toggling in a weird way.


flecho:
# Current behavior

Hello, I am a user of Boostnote. :-)
After the recent update to 0.11.6, boostnote behaves in a weird way.
Every time I press the 'm' button, the `edit` mode changes into `view` mode, which hampers my editing the note.
Are there anybody who suffer from the same issue?
Thanks,

# Expected behavior
It shouldn't toggle the mode, when I press `m` button.

# Workaround
I just found out my hotkey setting was set to toggle for `Cmd + M` button.
But, I don't know why just pressing `m` leads to toggle. 
To avoid this matter, I changed my key setting for toggle to some different key combinations, so I do not suffer from this. But, anyway, I think someone needs to check! :-)

# Environment

- Version : 0.11.6
- OS Version and name : MAC OS Sierra, 10.12.3

dbburgess:
Likely duplicates: #2101, #2098, #2094, #2086

ax-vasquez:
@dbburgess confirming your suspicion - I created #2098 and it is the same issue as this one.

voidsatisfaction:
meee tooo

dpo:
Same here. Makes Boostnote unusable really.

voidsatisfaction:
As I checked source code, problem is here

```js
// main/lib/ConfigManager.js
  hotkey: {
    toggleMain: OSX ? 'Cmd + Alt + L' : 'Super + Alt + E', // 'Cmd' should be altered as 'command' and M also should be m
    toggleMode: OSX ? 'Cmd + M' : 'Ctrl + M'
  },
```

these hotkeys are linked to `Mousetrap library` which makes shortcut possible at global level.

check https://github.com/ccampbell/mousetrap

@kazup01 

p.s I want to fix this problem but currently I have npm error so I will try this after several hours later

PR #2109, opened 2018-06-20
Fix MacOs 'M' key issue

voidsatisfaction:
related: #2101 

# Before

In editor mode, pressing M key trigger editor mode toggle(Preview <=> Split)

# After

1. change wrong default value
2. But **those who already set hot key as 'Cmd + M' should change 'Command + M' manually**  (Because Boostnote use localStorage to save user configuration including hotkey. But It can only be changed by user)
3. change `Command + M` action to more desirable way(I make it focus after change editor mode so that It sustains focus when editing)


Rokt33r:
LGTM. I'll also make another pr to correct user's config automatically.

Issue #1979, opened 2018-05-26
behavior about DropDownList for storage folders 


KOROKORO-Manbo:
# Current behavior
DropDownList for storage folders are automatically(=against my idea) expanded on selection.It's kind of inconvenient when you have looong list of strage.
![expandingfolder](https://user-images.githubusercontent.com/16297739/40571426-7e775a28-60d3-11e8-8997-2f953c7b1afa.gif)
# Expected behavior
Keep folded til I expand it.
if possible, keep the state of dropdownlist when starting app next time.

# Environment

- Version :0.11.5
- OS Version and name :High Sierra 10.13.4

IssuehuntBot:
@kelvin-wong has started working. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/1979)

IssuehuntBot:
@kelvin-wong has submitted [output](https://github.com/BoostIO/Boostnote/pull/2144). [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/1979)

IssuehuntBot:
@kazup01 has rewarded. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/1979)

Issue #1911, opened 2018-05-12
Retain collapsed storage location state


jonblack:
When a folder is selected, the collapsed state of storage locations in the left pane is lost - all locations are uncollapsed.

Steps to reproduce:

1. Add multiple storage locations with multiple folders
2. Collapse all but one of the storage locations in the left pane
3. In the non-collapsed location in the left pane, select a folder
4. All storage locations will uncollapse

Expected behaviour:

After selecting a folder, the collapsed state shouldn't change

IssuehuntBot:
@kazup01 has boosted this issue with $20. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/1911)

IssuehuntBot:
@kelvin-wong has started working. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/1911)

IssuehuntBot:
@kelvin-wong has submitted [output](https://github.com/BoostIO/Boostnote/pull/2144). [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/1911)

IssuehuntBot:
@kazup01 has rewarded. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/1911)

PR #2144, opened 2018-06-28
Add collapsed state for storage

kelvin-wong:
The root cause of this issue is that when the folder is clicked,
the router pushed the path and the StorageItem component has been
refreshed and isOpen has been reset

- Add storing collapse state for storage
- Add tests
- Default as collapsed for fallback

fix BoostIo/Boostnote#1979 BoostIo/Boostnote#1911


kelvin-wong:
@sosukesuzuki Thanks!

Issue #2159, opened 2018-06-30
Export to HTML Broken in Windows


svdborre:
# Current behavior
When exporting a note to HTML, the below error occors:

![boostnote-error](https://user-images.githubusercontent.com/36859918/42123821-d2a289d8-7c58-11e8-9bb0-80f1bdd6a02b.jpg)

# Expected behavior

HTML export should work.

# Steps to reproduce

1. Export the note to HTML

# Environment

- Version : 0.11.7
- OS Version and name : Windows 7

fromtier:
same as windows 10 /ver 0.11.7

IssuehuntBot:
@boostio funded this issue with $20. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/2159)

IssuehuntBot:
@abadewitz has started working. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/2159)

IssuehuntBot:
@abadewitz has stopped working. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/2159)

IssuehuntBot:
@yamash723 has started working. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/2159)

IssuehuntBot:
@yamash723 has submitted [output](https://github.com/BoostIO/Boostnote/pull/2171). [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/2159)

IssuehuntBot:
@kazup01 has rewarded. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/2159)

PR #2171, opened 2018-07-03
Fix search value for html export path

yamash723:
Fixed issue: #2159

The path generated by `file-url` begins with `file:///`.

```javascript
file.replace('file://', '')  //    /C:/Users/........  wrong path.
file.replace('file:///', '') //    C:/Users/........   correct path.
```

### DIFF
--- a/browser/components/MarkdownPreview.js
+++ b/browser/components/MarkdownPreview.js
@@ -225,7 +225,7 @@ export default class MarkdownPreview extends React.Component {
       const attachmentsAbsolutePaths = attachmentManagement.getAbsolutePathsOfAttachmentsInContent(noteContent, this.props.storagePath)
 
       files.forEach((file) => {
-        file = file.replace('file://', '')
+        file = file.replace('file:///', '')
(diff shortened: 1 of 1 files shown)


browser/components/MarkdownPreview.js
```diff
@@ -225,7 +225,7 @@ export default class MarkdownPreview extends React.Component {
       const attachmentsAbsolutePaths = attachmentManagement.getAbsolutePathsOfAttachmentsInContent(noteContent, this.props.storagePath)
 
       files.forEach((file) => {
-        file = file.replace('file://', '')
+        file = file.replace('file:///', '')
```
Rokt33r: This fix doesn't work on macOS. I think we have to check which os is used. 
```js
if (global.process.platform === 'win32') {
  file = file.replace('file:///', '')
} else {
  file = file.replace('file://', '')
}
```

yamash723: Thanks for your review. I fixed it.

PR #2177, opened 2018-07-03
Spell check

### LINKED ISSUE #2176: Spellcheck
It has been mentioned several times that it would be nice to have a spell check feature.
I implemented one for markdown notes but it is some kind of slow and the performance is not that good. Maybe you can have a look at it and decide if we want to include it in boostnote (ignoring the performance issues) or if you can help to improve it..

Eventhough it's not perfect, it has been a lot of work so i would love to hear your suggestions.. @Rokt33r ?

ehhc:
Attempt to fix #2176


ehhc:
@Rokt33r unfortunately i still didn't had the time to include the debouncing of the live spell check.. it's on my todo list :)
but therefore it should still be 'in progress' and not 'in review'..

ehhc:
I will make a new PR branching a newer version of the master..

Issue #2165, opened 2018-07-01
[Feature request] Hotkey for "Fullscreen" toggle button.


StevenJL:
Can we assign a hot key for the Fullscreen toggle?  I'm referring this button which makes the notepad expands over the file navigation panels.

![boostnote_fullscreen](https://user-images.githubusercontent.com/2191808/42139561-fab2a074-7d44-11e8-9eb6-a4ce1abf0422.png)

Can we make a hot key to toggle this instead of always clicking this "Fullscreen" toggle button? Thanks

Rokt33r:
It is possible. How do you think which key should be set there?

StevenJL:
> It is possible. How do you think which key should be set there?

How about Shift-Cmd-F? I often use Shift-Cmd-L to focus the search bar, so it seems like Shift-Cmd is being used.  So how about F for "Fullscreen"?

IssuehuntBot:
@stevenjl funded this issue with $20. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/2165)

IssuehuntBot:
@enyaxu has started working. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/2165)

IssuehuntBot:
@enyaxu has submitted [output](https://github.com/BoostIO/Boostnote/pull/2178). [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/2165)

enyaxu:
How about use 'CommandOrCtrl +B' for hot key? **B** short for 'Side **B**ar'. I pushed a pull for this  feature. Please review

Rokt33r:
Great! @StevenJL How do you think?

StevenJL:
@Rokt

IssuehuntBot:
@kazup01 has rewarded. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/2165)

piotrgajow:
Is this feature available currently? If not when will it be?

Rokt33r:
@piotrgajow https://github.com/BoostIO/Boostnote/pull/2178 It is available now.

PR #2178, opened 2018-07-04
Hotkey for toggle editor fullscreen

enyaxu:
Add hotkey for toggle side bar
> <kbd>Command+B</kbd> for macos
> <kbd>Ctrl+B</kbd> for others

The **B** abbreviation for 'Side  **B**ar'.
#2165 
kapture 2018-07-04 at 12 08 19


ZeroX-DG:
Hmm...I think the word full screen will confuse users. There is already a shortcut for full screen mode F11. How about full screen edit mode or something like that, I'm bad at naming stuff :smile:

Rokt33r:
But, `F11` is a reserved keybinding of macOS!

Issue #1001, opened 2017-10-21
[Feature request] Back button


lorean1024:
It would be great to have the ability to go back to the previously displayed note. It would be extremely helpful while navigating links to different notes, wiki-style.

kazup01:
Hi @lorean1024 , how about this one? https://github.com/BoostIO/Boostnote/pull/693

lorean1024:
Hi! I'm not sure, but I can only see linking functionality there, without the ability to go back.
I'm thinking about the case when we have different notes linking to one note and after clicking the link we want to go back to the note that got us there - I think it's currently impossible.

jyhong836:
Hi, I also think this is necessary. Jumping between notes may distract me from previous notes. A button to get back to **history notes**  will be helpful.

IssuehuntBot:
@zerox-dg has started working. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/1001)

IssuehuntBot:
@zerox-dg has stopped working. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/1001)

IssuehuntBot:
@subtleparesh has started working. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/1001)

SubtleParesh:
I am working on this one, **back Button feature**, I am creating a **back** and **forward** button to provide navigation to previous notes we have opened,  for Markdown-notes only. Its working basically, but any details on this feature ... please comment

jyhong836:
About the feature, I found that the navigation will ignore the folders. For example, when I click a link from one note to the other, the presented note changes but the folder doesn't. When you are working on this, can you make the folder change consistently?

Besides, could you make the note history stored somewhere? Just like the Safari at Mac. When you long press the new-tab button (a plus button), you can view the recently closed tabs. I love that function.

jyhong836:
Sound great! Thank you for your work.

jyhong836:
@SubtleParesh Could you sustain the history of scroll position. Now, when I jump between notes, the pages will automatically change to the beginning. For me, I'd like to continue where I stopped.

IssuehuntBot:
@subtleparesh has submitted [output](https://github.com/BoostIO/Boostnote/pull/2179). [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/1001)

IssuehuntBot:
[...]

PR #2179, opened 2018-07-04
#1001 Back button Feature

SubtleParesh:
Back Button #1001  
A feature to show  all the history of notes is also added please review and suggest changes


browser/main/Detail/MarkdownNoteDetail.js
```diff
@@ -220,7 +362,7 @@ class MarkdownNoteDetail extends React.Component {
     }
   }
 
-  handleUndoButtonClick (e) {
+  handleRestoreButtonClick (e) {
```
SubtleParesh: Changed for more concise word

Rokt33r:
@SubtleParesh Could you fix the conflicts and lint errors from the ci?

SubtleParesh:
Cleared Errors

Rokt33r:
Thanks!!

But, I should ask one more thing. I don't think the history menu is that useful. It just looks confusing.
Could you remove it? I think the arrow buttons are enough.

And, why do we need `.yarnclean`?

SubtleParesh:
sorry for the late replay and... someone asked me the history menu would be good.... will be removing it right away .... and we dont need yarn clean will add it to git-ignore

SubtleParesh:
Done .... removed history menu..... now its just forward and backward buttons... good to merge @Rokt33r

browser/main/Detail/MarkdownNoteDetail.js
```diff
@@ -91,6 +105,120 @@ class MarkdownNoteDetail extends React.Component {
...
+    if (back.future.length > 7) {
+      back.future.splice(0, 1)
+      this.setState({
+        backStack: back
+      }, () => {
+        this.save()
+      })
+    }
+    if (back.past.length) {
+      if (!this.isBackActive) {
```
Rokt33r: This suppose to be `this.state.isBackActive`. Does it make sense?

mbarczak:
Hi, just throwing my 3 cents here : I think that the history menu could be very useful and not confusing, if it’s exposed properly to the end user. An example of good implementation could be back/forward feature from Thunderbird : 
back_forward_demo
So the main functionality is going back/forward, and if anyone requires deeper control , can use the “history menu” available as a pull down menu glued to the back/forward button. @SubtleParesh  worked hard on this feature, and It would be a waste to get rid of the “history menu” feature, especially if most of it is already coded ;)

daiyam:
Hi, I was checking this feature which looks good but I found an issue: when navigating, it's saving a note even if I was just seeing.

About the history menu, as it is, the navigation buttons are good for me. But maybe for other people,  a drop down menu can be added between the 2 arrows. In the menu, the current note can be marked as active.

SubtleParesh:
Thank you for your reply @mbarczak  and @daiyam,
@daiyam you say the note is saved even when navigating, but most of the times we just see the note and do not edit it and maybe i can add some kind of timer so that if 3+ sec viewed then it will be added to the history. And yes the menu item should be surely marked as active
As @mbarczak  has asked should we keep the history menu (and yes it needs some edits)?

daiyam:
@SubtleParesh A note shouldn't be marked as updated when we only view it.

browser/main/Detail/MarkdownNoteDetail.js
SubtleParesh: Yeah... my bad... Thank for your help @Rokt33r

SubtleParesh:
Made some changes ... I am not able to get dailyam 
I dont think the note being edited or not .... has any relation with history .... though i would like to make any changes required.... 
Some changes i have made and i have highlighted the active note 
screenshot from 2018-09-15 11-58-11

daiyam:
@SubtleParesh, the history is being hiding by the preview (maybe due to some changes in the `master`)

Yes,  I agree whatever we edit or preview a note, that note should be added to the history.
But you should save a note only when we are editing it.

Here a part of the problematic code:
```
if (back.past.length > 7) {
  back.past.splice(0, 1)
  this.setState({
    backStack: back
  }, () => {
    this.save()
  })
}
```

It should be just:
```
if (back.past.length > 7) {
  back.past.splice(0, 1)
  this.setState({
    backStack: back
  })
}
```

SubtleParesh:
Yes.....  now I get it... Thank you..... really appreciate your guidance.... especially to @daiyam  and @Rokt33r

daiyam:
@SubtleParesh thanks

The history menu seems to have problems with notes with the same title (only show one) and notes without title.
The next button is not disabled when reading the lastest note.

Rokt33r:
@SubtleParesh I'm sorry for being late to response. Could you fix the conflicts again? I'm going to merge this in this week.

sv3ndk:
@SubtleParesh, are you still working on this one ? Looks like a great feature!

SubtleParesh:
@Rokt33r, conflicts are fixed if your could merge it. sorry for the delay

Rokt33r:
@SubtleParesh Awesome! I'll review in this week. Hang on a little bit more!

Rokt33r:
@SubtleParesh I found it doesn't work well with snippet notes. Could you make another pr? I found the problem after I merged so I rolled back.

PR #2180, opened 2018-07-04
Add Ctrl+I and Ctrl+B support to format text

noelzubin:
Ctrl+I Now toggle Italics
Ctrl+B Now toggle Bold

Note: If there is a newline between the selected text this wont
work since codemirror doesn't wrap text with newline in them anyway

### DIFF
--- a/browser/components/CodeEditor.js
+++ b/browser/components/CodeEditor.js
@@ -153,8 +153,69 @@ export default class CodeEditor extends React.Component {
         'Cmd-T': function (cm) {
           // Do nothing
         },
+        'Ctrl-I': function(cm) {
+          if(cm.somethingSelected()) {
+            let selection = cm.getSelection()
+            let str = selection.trim()
+            let index = selection.indexOf(str)
+
+            if(str.length === 0 || str.match(/\n\s*\n/) !== null)
+              return
+
+            let newString 
+            let boldAndItalic = str.match(/^\*\*\*(.*)\*\*\*$/)
+            let bold = str.match(/^\*\*(.*)\*\*$/)
+            let italic = str.match(/^\*(.*)\*$/)
+            if(boldAndItalic !== null) {
+              newString = boldAndItalic[1]
+              newString = `**${newString}**`
+            } else if(bold !== null) {
+              newString = bold[1]
+              newString = `***${newString}***`
+            } else if(italic !== null) {
+              newString = italic[1]
+              newString = newString
+            } else {
+              newString = `*${str}*`
+            }
+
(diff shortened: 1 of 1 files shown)


browser/components/CodeEditor.js
```diff
@@ -153,8 +153,69 @@ export default class CodeEditor extends React.Component {
         'Cmd-T': function (cm) {
           // Do nothing
         },
+        'Ctrl-I': function(cm) {
```
Rokt33r: For macOS, we have to implement this to `Cmd-I` and `Cmd-B`.
So, we have to extract the handler and make it work depending on OS.
Like the below:
```js
'Cmd-I': function (cm) {
  if (global.process.platform !== 'darwin') { return }
  applyItalic(cm)
},
'Ctrl-I': function (cm) {
  if (global.process.platform === 'darwin') { return }
  applyItalic(cm)
}
```

browser/components/CodeEditor.js
```diff
@@ -153,8 +153,69 @@ export default class CodeEditor extends React.Component {
...
+              newString = newString
+            } else {
+              newString = `*${str}*`
+            }
+
+            let newSelection = selection.substr(0, index) + newString + selection.substr(index + selection.length)
+            cm.replaceSelection(newSelection)
+          }
+        },
+        'Ctrl-B': function(cm) {
```
Rokt33r: Same to here

browser/components/CodeEditor.js
```diff
@@ -153,8 +153,69 @@ export default class CodeEditor extends React.Component {
...
+              newString = `**${str}**`
+            }
+
+            let newSelection = selection.substr(0, index) + newString + selection.substr(index + selection.length)
+            cm.replaceSelection(newSelection)
+          }
+        },
         Enter: 'boostNewLineAndIndentContinueMarkdownList',
         'Ctrl-C': (cm) => {
+          console.log('here', cm)
```
Rokt33r: Please remove this unnecessary logging

noelzubin:
wokay done

Rokt33r:
I found another problem that is the shortcut has a conflict with https://github.com/BoostIO/Boostnote/pull/2178 .

How do you think @enyaxu?

enyaxu:
I noticed this problem, Maybe we should replan Bootnote shortcuts?

Right now shorcuts used by Bootnote don’t have explict config file, maybe we could create keybindings.json for  system use, and another keybindings.json file for user themselves to customize.

By do this , we can prevent shortkeys conflict, and user can now shortkeys clearly.

PR #2183, opened 2018-07-05
Flickering on create note

max-buranbaev:
Hello!
First of all, thanks for the great product. 👏 
Issue is when you create a new note you still see previous note for a sometime. The cause of issue is that re-rendering of code-editor happens after closing create-a-note section. I've added a small timeout to avoid it. 

Before: 
before
After fix:
after


Rokt33r:
It looks better! Thanks!

Issue #2133, opened 2018-06-26
Shortcut of Previous note is not working when focus is on the editor.


flyaway1217:
# Current behavior
When the focus is on the note list, ^J and ^K works fine.
But, when the focus is on the editor, ^J still works fine, ^K is not working any more.
I am using Vim in normal mode.

# Expected behavior

# Steps to reproduce

1.  Change into a Vim mode.
2. Create a new note and start writing note.
3. Try to use ^J or ^K to change note.

# Environment

- Version : 0.11.6
- OS Version and name : MacOS High Sierre 10.13.5

IssuehuntBot:
@boostio funded this issue with $10. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/2133)

IssuehuntBot:
@enyaxu has started working. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/2133)

IssuehuntBot:
@enyaxu has submitted [output](https://github.com/BoostIO/Boostnote/pull/2192). [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/2133)

IssuehuntBot:
@kazup01 has rewarded. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/2133)

modmod24:
I never used this, how does it work? I tried `Ctrl + [` and `Ctrl + ]` on linux nothing happens.

flyaway1217:
@modmod24 It works in my Mac. But the shortcuts are `Commend+[` and `Commend+]`.

modmod24:
I see now it works when using the GUI. Shortcuts don't!

-----
OS and Version: Ubuntu 16.04
Version: 0.11.9.pre

PR #2192, opened 2018-07-07
Replace shortcut for 'Next Note' and 'Previous Note'

enyaxu:
**Why**
Now shortcut key for **Next Note**(<kbd>ctrl+J</kbd>) and **Previous Note**(<kbd>ctrl+k</kbd>)had been conflicted with **Emac** Key map setting or **sublime** or **vim**, one of these key map settings also had It's inner use of key.   like **Emac** using  <kbd>ctrl+k</kbd> for repeatedly to kill multiple lines, but **sublime** using <kbd>ctrl+k</kbd> for combine action key. 
also see #2133 

**Do**
as above said, we need to change key map binding for 'Next Note' and 'Previous Note', I suggested that using <kbd>CommandOrCtrl+]</kbd> for 'Next Note', <kbd>CommandOrCtrl+[</kbd> for 'Previous Note'.
please review.

### DIFF
--- a/lib/main-menu.js
+++ b/lib/main-menu.js
@@ -235,14 +235,14 @@ const view = {
     },
     {
       label: 'Next Note',
-      accelerator: 'Control+J',
+      accelerator: 'CommandOrControl+]',
       click () {
         mainWindow.webContents.send('list:next')
       }
     },
     {
       label: 'Previous Note',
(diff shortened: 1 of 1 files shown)


Rokt33r:
Sound reasonable. How do you think, @sosukesuzuki ?

Issue #2204, opened 2018-07-13
Missing Note Preview When Building From Source


alecbcs:
# Current behavior
After building the application from source the note preview pane crashes with the error  `Uncaught Error: Cannot find module 'markdown-it-admonition'`. I've checked and markdown-it-admonition does seem to be installed by npm.

![screenshot from 2018-07-13 08-57-54](https://user-images.githubusercontent.com/19558067/42701551-be70e78e-867b-11e8-9f7e-36a2c963f2f2.png)
![screenshot from 2018-07-13 08-58-36](https://user-images.githubusercontent.com/19558067/42701569-ce616646-867b-11e8-8bf1-51db3c18a584.png)
This is the `package.yaml` for building the application on Solus.
![screenshot from 2018-07-13 09-00-05](https://user-images.githubusercontent.com/19558067/42701581-d39aaad2-867b-11e8-8c81-e70fbae74f54.png)

# Expected behavior
The note preview pane should be visible. 

# Steps to reproduce

1. Build application from source using pre-build
2. Install & open the application

# Environment

- Version : 0.11.7
- OS Version and name : Solus (3.999)

cdayjr:
Hi, I had this same error after @saaguero helped me out fix my issue. In my issue they suggested installing `markdown-it-admonition` in the `dist/Boostnote-linux-x64/resources/app` folder. This works, but I wanted to use the `rpm` package. It turns out all you need to do is run `yarn add markdown-it-admonition` in the source code root folder after running `yarn` the first time, and this'll cause the packages to be built with it when you run the build commands.

saaguero:
@cdayjr I suggest that you open a PR with your fix so that it can be reviewed.

cdayjr:
@saaguero Great idea, I submitted a pull request here: #2214

dbjsy:
Hello,

Do we know when the fix for the issue will be merged?

Thanks!

IssuehuntBot:
@boostio funded this issue with $30. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/2204)

dbjsy:
:+1: @Rokt33r thank you :)

arcturus140:
@Rokt33r what is the status of this issue? I think it has been merged into 0.11.9 and can be closed?

PR #2214, opened 2018-07-18
Fix for BoostIO/Boostnote#2204

cdayjr:
A few others and myself were having some issues compiling from source. This was fixed when I swapped out the `markdown-it-admonition` repo to use whatever is in the yarn registry rather than the custom github repo for it, so I'm submitting that change.


Rokt33r:
Thanks! I'll check it work later.

PR #2302, opened 2018-08-15
restore window position

daiyam:
This change allows the window to be restored at the same position as when it was closed.


ZeroX-DG:
LGTM

Issue #2267, opened 2018-08-07
Todo percentage not updated correctly...


gufertum:
# Current behavior

If you check off checkboxes in markdown view, the percentage does not update - only when using the rendered view. But switching to rendered view and ticking a checkbox still show the incorrect number... and sometimes the editor overwrites some characters instead of inserting text...

# Expected behavior

Calculate the percentage correctly, regardless which view (renedered or markdown) is shown.

# Steps to reproduce

1. start with a fresh item checklist
2. switch to rendered view
3. check some checkboxes > still correct
4. switch to markdown
5. mark some checkboxes (add x) > the percentage does not update correctly and the editor starts overwriting "]" sometimes...

# Environment

- Version : 0.11.8
- OS Version and name : OXS 10.13.5

modmod24:
I can not reproduce the problem. Did you use uppercase 'X' or had spaces in between?

-----

- Version : 0.11.9.pre
- OS Version and name: 16.04 Ubuntu

gufertum:
Yes, uppercase, but I can't reproduce it consistently either. I need to edit around multiple times before it happens...

modmod24:
I couldn't find an official source but the uppercase variant seems valid, too.
[Here](https://github.com/westurner/wiki/wiki/Checkboxes--markdown) is a reference I found (unofficial). Most references I found are using lowercase. 

Github supports both variants:
+ [x] `[x]` lowercase 'x'
- \[x] `\[x]`
- \[x\] `\[x\]`
- [X] `[X]` uppercase 'x'
- \[X] `\[X]`
- \[X\] `\[X\]`

modmod24:
has been merged, can be closed.

modmod24:
@ZeroX-DG fixed in 0.11.9. Issue can be closed.

PR #2307, opened 2018-08-18
update todo percentage correctly #2267

modmod24:
this branch treats tasklist markdown as case insensitive


modmod24:
fix Todo percentage #2267

PR #2312, opened 2018-08-20
add smart table editor

daiyam:
This change allows the use of the smart table editor of [@susisu/mte-kernel](https://github.com/susisu/mte-kernel).
It can be enabled/disabled in the Preferences.


ZeroX-DG:
Hey @daiyam , thanks for the great feature, but, I notice that the feature will not run if the user don't restart the app, can you fix this?

ZeroX-DG:
LGTM :tada:

daiyam:
@Rokt33r I've fixed the conflicts

ZeroX-DG:
We should add a blog post on this :smile: too many keyboard shortcut :smile:

PR #2314, opened 2018-08-20
tag enhancements

daiyam:
This change add 2 new options:
- `showTagsAlphabetically`: show tags of a note in alphabetical order.
- `enableLiveNoteCounts`: enable live count of notes in the Side Bar. (tags without any related notes have no displayed count)


modmod24:
hi, would it make sense to sort the tags only once when they are created (and store in this order in file), instead of every time the  note is displayed?

sorry if this suggestion doesn't make sense. I am not exactly sure what your changes does.

daiyam:
@modmod24 no, it does make sense. But what about the existing tags?

Maybe the option `saveTagsAlphabetically` should be added or should replace `showTagsAlphabetically `.
That option is shown with a warning that by selecting it, it would sort the existing tags.
There might be a performance issue since we will have to re-save all the notes; same issue when we will be able to rename a tag...

modmod24:
@daiyam I was thinking about ordering only when saving tags, to prevent frequently ordering what's already in order. existing order would remain until new tag added.

modmod24:
yes, of course. thanks.

Rokt33r:
> enableLiveNoteCounts: enable live count of notes in the Side Bar. (tags without any related notes have no displayed count)

Could you tell me more what it is?

daiyam:
@Rokt33r here some screenshots to explain it better:

When `enableLiveNoteCounts` is **off** (by default):
screenshot

When `enableLiveNoteCounts` is **on**:
screenshot

The count on `bug` isn't shown because there is no notes with `feature` **and** `bug` tags.
The count on `table` is `1` instead of `4` because only one note has both `feature` **and** `table` tags.

Rokt33r:
I confirmed it works fine. Thanks for being late!

PR #2322, opened 2018-08-22
Fix drop image rotate wrong

youngyou:
Fix drag-drop image rotate wrong, as Jordan Thornquest mentioned in slack

### DIFF
--- a/browser/main/lib/dataApi/attachmentManagement.js
+++ b/browser/main/lib/dataApi/attachmentManagement.js
@@ -12,6 +12,82 @@ const STORAGE_FOLDER_PLACEHOLDER = ':storage'
 const DESTINATION_FOLDER = 'attachments'
 const PATH_SEPARATORS = escapeStringRegexp(path.posix.sep) + escapeStringRegexp(path.win32.sep)
 
+function getImage (file) {
+  return new Promise((resolve) => {
+    const reader = new FileReader()
+    const img = new Image()
+    img.onload = () => resolve(img)
+    reader.onload = e => {
+      img.src = e.target.result
+    }
+    reader.readAsDataURL(file)
+  })
+}
+
+function getOrientation (file) {
+  const getData = arrayBuffer => {
+    const view = new DataView(arrayBuffer)
+    if (view.getUint16(0, false) !== 0xFFD8) return -2
+    const length = view.byteLength
+    let offset = 2
+    while (offset < length) {
+      const marker = view.getUint16(offset, false)
+      offset += 2
+      if (marker === 0xFFE1) {
+        if (view.getUint32(offset += 2, false) !== 0x45786966) {
+          return -1
+        }
+        const little = view.getUint16(offset += 6, false) === 0x4949
+        offset += view.getUint32(offset + 4, little)
+        const tags = view.getUint16(offset, little)
+        offset += 2
+        for (let i = 0; i < tags; i++) {
+          if (view.getUint16(offset + (i * 12), little) === 0x0112) {
+            return view.getUint16(offset + (i * 12) + 8, little)
+          }
+        }
+      } else if ((marker & 0xFF00) !== 0xFF00) {
+        break
+      } else {
+        offset += view.getUint16(offset, false)
+      }
+    }
+    return -1
+  }
+  return new Promise((resolve) => {
+    const reader = new FileReader()
+    reader.onload = event => resolve(getData(event.target.result))
+    reader.readAsArrayBuffer(file.slice(0, 64 * 1024))
+  })
+}
+
+function fixRotate (file) {
+  return Promise.all([getImage(file), getOrientation(file)])
+  .then(([img, orientation]) => {
+    const canvas = document.createElement('canvas')
+    const ctx = canvas.getContext('2d')
+    if (orientation > 4 && orientation < 9) {
+      canvas.width = img.height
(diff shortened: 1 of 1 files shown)


browser/main/lib/dataApi/attachmentManagement.js
```diff
@@ -38,26 +114,34 @@ function copyAttachment (sourceFilePath, storageKey, noteKey, useRandomName = tr
     }
 
     try {
-      if (!fs.existsSync(sourceFilePath)) {
-        reject('source file does not exist')
+      const isBase64 = typeof sourceFilePath === 'object' && sourceFilePath.type === 'base64'
+      if (!fs.existsSync(sourceFilePath) && !isBase64) {
```
ehhc: Why do you reject images that are not base64? e.g. i might drag&drop a pdf. Are all files always base64?!

browser/main/lib/dataApi/attachmentManagement.js
```diff
@@ -12,6 +12,82 @@ const STORAGE_FOLDER_PLACEHOLDER = ':storage'
 const DESTINATION_FOLDER = 'attachments'
 const PATH_SEPARATORS = escapeStringRegexp(path.posix.sep) + escapeStringRegexp(path.win32.sep)
 
+function getImage (file) {
```
ehhc: There is no documentation at all.
Nor are there any tests i see..

browser/main/lib/dataApi/attachmentManagement.js
```diff
@@ -12,6 +12,82 @@ const STORAGE_FOLDER_PLACEHOLDER = ':storage'
...
+    const img = new Image()
+    img.onload = () => resolve(img)
+    reader.onload = e => {
+      img.src = e.target.result
+    }
+    reader.readAsDataURL(file)
+  })
+}
+
+function getOrientation (file) {
```
ehhc: For me it seems like this method does a lot of black magic.. what are all the constants about? I have no idea how this method works, what it does and what it returns..

browser/main/lib/dataApi/attachmentManagement.js
```diff
@@ -12,6 +12,82 @@ const STORAGE_FOLDER_PLACEHOLDER = ':storage'
...
+    return -1
+  }
+  return new Promise((resolve) => {
+    const reader = new FileReader()
+    reader.onload = event => resolve(getData(event.target.result))
+    reader.readAsArrayBuffer(file.slice(0, 64 * 1024))
+  })
+}
+
+function fixRotate (file) {
```
ehhc: How does it "fixes" the "wrong" rotation? I don't understand what might be wrong with it since i cannot reproduce the problem..

browser/main/lib/dataApi/attachmentManagement.js
youngyou: Only if the file not exists and the file is not a base64, reject.
In additional, I think you should add `return`  before `reject`, the code should not run through if the file not exists.

browser/main/lib/dataApi/attachmentManagement.js
youngyou: Ok, I will add it later

browser/main/lib/dataApi/attachmentManagement.js
youngyou: I will add comments later.

browser/main/lib/dataApi/attachmentManagement.js
ehhc: yes, but i mean what happens if you drop a PDF for example. Is that still possible with your changes? Or will it be rejected because it is no pdf?

browser/main/lib/dataApi/attachmentManagement.js
youngyou: Just like [this](https://stackoverflow.com/questions/10600613/ios-image-orientation-has-strange-behavior), https://stackoverflow.com/questions/10600613/ios-image-orientation-has-strange-behavior
default
landscape_5
This is a demo image.

browser/main/lib/dataApi/attachmentManagement.js
youngyou: This patch does nothing for other file. It will work same as before

youngyou:
@ehhc comments added.

modmod24:
I cannot replicate the issue.  I am on 0.11.9.1 in Ubuntu 16.04

browser/main/lib/dataApi/attachmentManagement.js
youngyou: It has many dependencies, maybe we don't need to import so many code.

youngyou:
Try drop this image. @modmod24

modmod24:
ok, I see

browser/main/lib/dataApi/attachmentManagement.js
Rokt33r: I agree with @ehhc. This feature looks good to have. But I think it should not be included in the code base of boostnote. It is almost impossible to maintain...

Rokt33r: Why don't we make a npm module for this? And, I think we need some test for this too.

youngyou: Ok, I'll do it some days later.

PR #2325, opened 2018-08-22
sync local tag with wordpress tag

phantomlution:
when user publish their notes to wordpress, sync server-side tags with boostNote tags


ZeroX-DG:
@phantomlution are you still interested in this PR?

Issue #2188, opened 2018-07-05
[Bug Report] Compromized tag due to special character


modmod24:
# Current behavior
Tags containing "/" character can be created. They appear in the list of tags, displayed in the note itself and in the "Tags". However, the tag is not clickable. On clicking on the tag corresponding notes are not listed. 

![boostnote_bugtags](https://user-images.githubusercontent.com/25648628/42347161-e618822c-809c-11e8-96e4-9b85a9bd4e1d.png)

# Expected behavior
Offending characters should be truncated, I think. There may be more character causing unexpected behaviour. 

# Steps to reproduce

1. Create a new note
2. add a tag: 'this/is/a/test
3. open "Tags" list
4. find the created tag and select it (left-click)

# Environment

- Version : 0.11.6
- OS Version and name : 16.04.4 Ubuntu

IssuehuntBot:
@boostio funded this issue with $30. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/2188)

IssuehuntBot:
@daiyam has started working. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/2188)

IssuehuntBot:
@daiyam has submitted [output](https://github.com/BoostIO/Boostnote/pull/2337). [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/2188)

IssuehuntBot:
@kazup01 has rewarded. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/2188)

Issue #2321, opened 2018-08-22
11.8 Using C# as a tag


zzxxhhzxh:
# Current behavior
When using "C#" as a tag, this tag can not be clicked to show the notes.
"C##", "C#a" also can not.

# Expected behavior

# Steps to reproduce

1. 
2. 
3. 

# Environment

- Version :
- OS Version and name :

modmod24:
how were you even able to add `C#` tag? I wasn't. Same for `C##`, `C#a`

# Environment

-----

- Version : 0.11.9-pre.1
- OS Version and name : 16.04 Ubuntu

zzxxhhzxh:
My version is 0.11.8, I use windows10 64 bit.
I can input C# just like you.


![image 001](https://user-images.githubusercontent.com/42593813/44469960-31bbf080-a5ee-11e8-8433-70aa0282f9db.jpg)

modmod24:
# Current behavior

I upgraded to version 0.11.9-pre.2 and am now able to create the tags `C#`, `C##`, `C#a`. Tags are now clickable but it doesn't work right. 

# Steps to reproduce

1. create note, tag with `C`
2. create note, tag with `C#`
3. observe: both tags exist, each containing 1 note
4. click on tag `C#`, observe: notes tagged with `C` are listed, notes tagged with `C#` are not listed

@zzxxhhzxh is this what you where experiencing on your system? The behaviour is different to your description. 

![master_branch](https://user-images.githubusercontent.com/25648628/44981822-a2a6c500-af6b-11e8-937a-d182a8cfded9.gif)


# Expected behaviour - branch "fixing bug #2321"

![branch2321](https://user-images.githubusercontent.com/25648628/44981833-ac302d00-af6b-11e8-8a27-936ac840b7cf.gif)


tags should correlate with notes. I checked out branch from @daiyam and everything works as shown above. The .gif doesn't show it but `this/is/a/test` works as well. 

# Environment 

Version : 0.11.9-pre.2
Version and OS : 16.04 Ubuntu

modmod24:
Update: I reverted to 0.11.8 the behaviour is the same as 0.11.9-pre.2

PR #2337, opened 2018-08-24
fixing bug #2321 

daiyam:
This change is fixing the bug #2321 due to the unescaped characters `#` or `?` in the route.


ZeroX-DG:
LGTM :tada:

hsiehjack:
@daiyam This pull request also resolves #2188

daiyam:
@hsiehjack you are right. Thx

PR #2338, opened 2018-08-25
Spellchecker

ehhc:
another try to add spellcheck to boostnote.
The live spell check is debounced now..
Unfortunately the context menu is buggy if it is invoked from the keyboard and not with the mouse.. I have absolutely no idea why that happens.. Maybe it's enough to be merged anyway @Rokt33r ?


nac2791:
Is this still active?

ehhc:
@nac2791 if i understand it right, the mainteiner (@Rokt33r ) decided that my solution is too complex and will not be merged therefore.
I haven't closed the PR because i don't agree and will not close it. If they want to, they have to do it themself

Rokt33r:
Alright. I changed my mind. Let's give a shot.

@ehhc Could you fix the conflicts?

ehhc:
@Rokt33r the conflicts are resolved :)

Issue #1992, opened 2018-05-28
Tag <kbd> broke


bravo321:
![key-tag](https://user-images.githubusercontent.com/26696732/40613729-69796b5a-6288-11e8-8e20-3f859bd645d8.png)

- Version : Boostnote 0.11.4

- OS Version and name : Ubuntu 16.04 Linux 4.4.0-127-generic x86_64

IssuehuntBot:
@boostio funded this issue with $10. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/1992)

IssuehuntBot:
@daiyam has started working. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/1992)

IssuehuntBot:
@kazup01 has rewarded. [See it on IssueHunt](https://issuehunt.io/repos/53266139/issues/1992)

PR #2342, opened 2018-08-25
fixing inline html

daiyam:
This change fixes the issue #1992.

The issue was due to an incompatibility between `markdown-it` and `sanitize-html`.
`markdown-it` is splitting `<kbd>Ctrl</kbd>` in 3 parts `<kbd>`, `Ctrl` and `</kbd>`.
The opening and closing parts were then given to `sanitize-html` which corrects each string to valid html.

`<kbd>` became `<kbd></kbd>`.
`<kbd>` became empty.
So `<kbd>Ctrl</kbd>` became `<kbd></kbd>Ctrl`


daiyam:
@Rokt33r I've added a test and fixed a bug found due to it.

PR #2347, opened 2018-08-27
add support to front-matter

daiyam:
This change adds support of YAML front-matter by:
- hiding the front-matter in preview
- skipping the front-matter when looking for note's title

### DIFF
--- a/browser/lib/findNoteTitle.js
+++ b/browser/lib/findNoteTitle.js
@@ -1,13 +1,11 @@
-const frontMatterRegex = /^\-{3,}/
-
 export function findNoteTitle (value) {
   const splitted = value.split('\n')
   let title = null
   let isInsideCodeBlock = false
 
-  if (frontMatterRegex.exec(splitted[0])) {
+  if (splitted[0] === '---') {
     let index = 0
-    while (++index < splitted.length && !frontMatterRegex.exec(splitted[index])) {
+    while (++index < splitted.length && splitted[index] !== '---') {
     }
 
     splitted.splice(0, index + 1)
--- a/browser/lib/findNoteTitle.js
+++ b/browser/lib/findNoteTitle.js
@@ -4,11 +4,14 @@ export function findNoteTitle (value) {
   let isInsideCodeBlock = false
 
   if (splitted[0] === '---') {
-    let index = 0
-    while (++index < splitted.length && splitted[index] !== '---') {
-    }
+    let line = 0
+    while (++line < splitted.length) {
+      if (splitted[line] === '---') {
+        splitted.splice(0, line + 1)
 
-    splitted.splice(0, index + 1)
+        break
+      }
+    }
   }
 
   splitted.some((line, index) => {
--- a/browser/lib/findNoteTitle.js
+++ b/browser/lib/findNoteTitle.js
@@ -1,7 +1,17 @@
+const frontMatterRegex = /^\-{3,}/
+
 export function findNoteTitle (value) {
   const splitted = value.split('\n')
   let title = null
   let isInsideCodeBlock = false
+
+  if (frontMatterRegex.exec(splitted[0])) {
+    let index = 0
+    while (++index < splitted.length && !frontMatterRegex.exec(splitted[index])) {
+    }
+
+    splitted.splice(0, index + 1)
+  }
 
   splitted.some((line, index) => {
     const trimmedLine = line.trim()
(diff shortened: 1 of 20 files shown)


daiyam:
@sosukesuzuki I've added the test.

sosukesuzuki:
@daiyam 
I am not familiar with front-matter. :bow:
is the below you expected? or bug?
- When you start writing front-matter from the second line, it is not hidden.

daiyam:
@sosukesuzuki It's because you have an empty line before it.

> The front matter must be the first thing in the file and must take the form of valid YAML set between triple-dashed lines.
[Jekyll Doc](https://jekyllrb.com/docs/frontmatter/)

[Here some examples of use](http://assemble.io/docs/YAML-front-matter.html)

browser/lib/markdown.js
```diff
@@ -149,6 +149,7 @@ class Markdown {
     })
     this.md.use(require('markdown-it-kbd'))
     this.md.use(require('markdown-it-admonition'))
+    this.md.use(require('markdown-it-front-matter'), fm => {})
```
ZeroX-DG: Why do you need an empty callback? Please remove it if you accidentally put it there.

browser/lib/findNoteTitle.js
```diff
@@ -1,8 +1,18 @@
+const frontMatterRegex = /^\-{3,}/
```
ZeroX-DG: > The front matter must be the first thing in the file and must take the form of valid YAML set between triple-dashed lines. 

This regex will also valid for line that started with 4 dashes. You should remove the `,` there.

ZeroX-DG:
I also think if a front matter is invalid then the markdown should display it out. However, with the current behavior, if I type the code bellow, all the content of the note will be hidden.
```
----
The line bellow has 3 dashes but the line above has 4 dashes and that will hide all the note content.
---
```

browser/lib/findNoteTitle.js
daiyam: Yes, you are right.

browser/lib/markdown.js
daiyam: No accident, the plugin needs it or it won't work... It used so we can parse the front-matter to extract the properties. But since we don't need to do that, the code block is empty.

ZeroX-DG: Well that's annoying :smile: maybe because the plugin did check if the callback function is provided or not and assume that's there :smile:

browser/lib/findNoteTitle.js
ZeroX-DG: I think we should stick to the traditional `---` :smile:

daiyam:
@ZeroX-DG I've replaced the plugin with my own plugin so that delimiter must be 3 dashes.

browser/lib/markdown-it-frontmatter.js
```diff
@@ -0,0 +1,21 @@
+'use strict'
+
+module.exports = function frontMatterPlugin (md) {
+  function frontmatter (state, startLine, endLine, silent) {
+    if (startLine !== 0 || state.src.substr(startLine, state.eMarks[0]) !== '---') {
+      return false
+    }
+
+    let line = 0
+    while (++line < state.lineMax && state.src.substring(state.bMarks[line], state.eMarks[line]) !== '---') {
```
ZeroX-DG: I still think we should display the markdown if the front-matter syntax is invalid. For example:
```
---
layout: default
-----
sdasdasdasdasd
```
Should display:
image
With the current behavior (hide all content) the user will get confuse.

Here is my suggestion for the plugin:
```js
module.exports = function frontMatterPlugin (md) {
  function frontmatter (state, startLine, endLine, silent) {
    if (startLine !== 0 || state.src.substr(startLine, state.eMarks[0]) !== '---') {
      return false
    }

    let line = 0
    let isFrontMatterClosed = false
    while (line < state.lineMax) {
      line++
      if (state.src.substring(state.bMarks[line], state.eMarks[line]) === '---') {
        isFrontMatterClosed = true
        break
      }
    }

    state.line = line + 1

    return isFrontMatterClosed
  }

  md.block.ruler.before('table', 'frontmatter', frontmatter, {
    alt: [ 'paragraph', 'reference', 'blockquote', 'list' ]
  })
}
```
The the modification above, the front-matter will be displayed out if it's invalid and will be hide only when it's valid.

PR #2371, opened 2018-09-05
Tags over the editor are clickable

daiyam:
This change allows to filter the notes by clicking on a tag in the tag list over the editor.


daiyam:
@Rokt33r I've resolved the conflicts

PR #2381, opened 2018-09-07
Pipe character not escaped when inserting a URL into a table [WIP]

### LINKED ISSUE #2362: Pipe character not escaped when inserting a URL into a table
# Current behavior

Boostnote gets the title of websites if an URL is inserted. However, if the title contains a pipe character it is not escaped when the URL is in a table. 
boostnote_issue_1

boostnote_issue_2

# Expected behavior
The title extraction should not mess up the table. As far as I can understand, it is also possible to escape the pipe character even outside of the table, so that seems like an easy fix. If this breaks something, it is also possible to escape the character only if the URL is inserted in a table.

# Steps to reproduce

1. Create a table with two columns
2. Insert some text into the second column
3. Insert a URL with a pipe in its title into the first column
4. Observe that the second column text is gone from the preview

# Environment

- Version : 0.11.8
- OS Version and name : Ubuntu 16.04

Antogin:
Closes #2362

Before:
screen recording 2018-09-07 at 02 50 19 58 pm
After:
screen recording 2018-09-07 at 02 53 12 75 pm

Currently escapes <kbd>|</kbd> in titles at all times but the escape symbol will note be in the preview


Antogin:
@daiyam @OndragKlaus. Will this do?

OndragKlaus:
I actually like that this will influence the titles generated outside of tables too because the code is simpler and it makes it easier to copy/paste.

PR #2382, opened 2018-09-08
Links are clickable

daiyam:
This changes allow to click on the links in the editor (like in VS Code).

link-clickable

### DIFF
--- a/extra_scripts/codemirror/addon/hyperlink/hyperlink.js
+++ b/extra_scripts/codemirror/addon/hyperlink/hyperlink.js
@@ -10,11 +10,10 @@
   'use strict'
 
   const shell = require('electron').shell
+  const yOffset = 2
 
   class HyperLink {
     constructor(cm) {
-      this.xOffset = 10
-      this.yOffset = 2
       this.cm = cm
       this.lineDiv = cm.display.lineDiv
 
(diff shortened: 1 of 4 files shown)


extra_scripts/codemirror/addon/hyperlink/hyperlink.js
```diff
@@ -0,0 +1,124 @@
...
+      if (className.indexOf('cm-url') !== -1) {
+        const match = /^\((.*)\)|\[(.*)\]|(.*)$/.exec(el.textContent)
+        return match[1] || match[2] || match[3]
+      }
+
+      return null
+    }
+    onMouseDown(e) {
+      const { target, metaKey } = e
+      if (!metaKey) {
```
ZeroX-DG: > On Macintosh keyboards, this is the ⌘ Command key. On Windows keyboards, this is the Windows key (⊞ Windows).

I don't know how it looks on MacOS but on Windows and Linux, the meta key is used to open the main menu therefore when you hit meta key + click link, the link does open in the browser but the menu is also displayed which is quite annoying.

I think you will have to change this key to something else. Also you can't set it to <kbd>ctrl</kbd> on windows and linux as it's the key for multi-cursor in codemirror. `Shift perhaps?`

daiyam:
@ZeroX-DG oops, my bad. I've fixed it.

PR #2389, opened 2018-09-11
fix admonitions -> use the "legacy" types and pass them as options to…

### LINKED ISSUE #2379: 11.9: admonitions (except 'note') not rendered  on Linux (regression)
# Current behavior

# Expected behavior
Admonitions should be rendered (works in 11.8)

# Steps to reproduce

1.  Upgrade to 11.9
2.  Show note that uses not 'note' admonition
3. 

# Environment

- Version :  11.9
- OS Version and name : Linux Mint 19

ehhc:
… the plugin. should fix #2379


ehhc:
thank you @kazup01 :)

PR #2340, opened 2018-08-25
fixing Mermaid diagram's height

daiyam:
This change allows to specify the desired height of the diagram (in percentage in `vh`).
It should be able to fix #2335.

For example:
~~~
```mermaid:50
graph TB
	c1-->c2
	c2-->c3
	c3-->c4
	c4-->c5
	c5-->c6
```
~~~

Issue #2335, opened 2018-08-24
Mermaid diagrams don't resize in height


ChrisK91:
[<img alt="Issuehunt badges" src="https://issuehunt.io/r/BoostIo/Boostnote/issues/2335/badge.svg" />](https://issuehunt.io/r/BoostIo/Boostnote/issues/2335)

# Current behavior

When creating "tall" mermaid graphs, the graph area is not resized in height, resulting in unreadable text, both in preview and in output

# Expected behavior

The graph area should be resizable, to allow creation of readable graphs. One option could be, to allow user specified dimensions for the created graph area.

# Steps to reproduce

1. create a new note
2. Insert: 
```
\`\`\`mermaid
graph TB
    c1-->c2
    c2-->c3
    c3-->c4
    c4-->c5
    c5-->c6
```
3. Observe the resulting output: 
![image](https://user-images.githubusercontent.com/1904175/44575163-ae380600-a78b-11e8-9151-baa20d4b3295.png)

# Environment

- Version : 0.11.8
- OS Version and name : Windows 10, display resolution 1080p

> There is a $40.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/BoostIo/Boostnote/issues/2335).
> - Checkout the [Issuehunt explorer](https://issuehunt.io/r/BoostIo/Boostnote/) to discover more funded issues.
> - Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on Issuehunt to raise funds.

daiyam:
@ChrisK91 here what I get with the PR #2585:
![screenshot](https://user-images.githubusercontent.com/587742/50418133-373e7700-082a-11e9-8240-aae988bcb2b1.png)

ttbadr:
i had same issue in version 0.11.12, hope to fix it ,thanks

dayfuaim:
Same _still_ in 0.11.15 (OSX 10.10).
Needs to be fixed.

qiuker521:
same still 0.11.15 WIN10
Needs to be fixed.

And thanks for your mermaid support.

![image](https://user-images.githubusercontent.com/22740352/55895994-e5123d80-5bef-11e9-8a6b-31859fc99550.png)

IssuehuntBot:
@boostio has funded $40.00 to this issue.

---
- Submit pull request via [IssueHunt](https://issuehunt.io/repos/53266139/issues/2335) to receive this reward.
[...]

PR #2394, opened 2018-09-13
add attributes to fence blocks

daiyam:
This change adds attributes to fence blocks like: `mermaid(50h):Graph 1`

Its format is: `language(attributes):fileName:firstLineNumber`.
The formats for an attribute are: `name="value"` or `name='value'` or `name=value` or `name`.
The attribute `<number>h` is transformed as `height=<number>`

For example:
- `javacript:test.js:5`
- `mermaid`
- `mermaid(50h)`
- `mermaid(height=50):Graph 1`
- `mermaid(height="50"):Graph 1`
- `mermaid(height='50')`
- `chart(yaml 50h):My chart 8`

`chart`, `flowchart`, `mermaid`, `sequence` will display the `fileName` as **title**.

This is a better fix for #2335 than my previous proposed pull request #2340

screenshot


browser/lib/markdown-it-fence.js
```diff
@@ -0,0 +1,130 @@
...
+    token.map = [startLine, state.line]
+
+    return true
+  }
+
+  md.block.ruler.before('fence', '_fence', fence, {
+    alt: ['paragraph', 'reference', 'blockquote', 'list']
+  })
+
+  for (let name in renderers) {
```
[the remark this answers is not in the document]
daiyam: @ZeroX-DG thanks! I did learn something today [The for-of loop](http://exploringjs.com/es6/ch_for-of.html)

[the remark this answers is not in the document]
ZeroX-DG: Haha, amazing right? I also discorvered the `for-of` loop a few months ago :smile: I realized that I've been in stone age for years :smile:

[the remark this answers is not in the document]
daiyam: I did know the `for-of` loop but I didn't know I could use `const` for the iterator. (Since the loop do change the value/address of the variable)

browser/components/markdown.styl
```diff
@@ -213,29 +213,25 @@ pre
...
+  &.CodeMirror
+    height initial
+    flex-wrap wrap
+    &>code
+      flex 1
+      overflow-x auto
   &>span.filename
-    width 100%
-    border-radius: 5px 0px 0px 0px
     margin -8px 100% 8px -8px
```
ZeroX-DG: Suggested change to fix a small display bug:
`z-index: 10`
Before:
image
After:
image

daiyam: @ZeroX-DG I removed the errors for every diagram (chart, flowchart, mermaid and sequence).
I've also fixed a bug with the line number...

But I can't replicated the visual bug you are getting...
Here what I get:
screenshot

Here my preferences:
screenshot

ZeroX-DG: You preview font size is too small, you can set it to `24` and you will see the problem

daiyam: Here my fix:
screenshot

ZeroX-DG:
@daiyam Can you fix the conflict ?

daiyam:
@ZeroX-DG it's done

Rokt33r:
@daiyam I'm sorry for being late response. Could you fix the conflict?

daiyam:
@Rokt33r all done

Rokt33r:
Awesome!

daiyam:
@100ideas the `1200` is just a fallback if there is no width (in a `display:none`). 
But the problem is that it's using the width of the window as its width.

daiyam:
@100ideas I've found a way to fix this. You can test it with the PR #2585.

PR #2399, opened 2018-09-14
Clear search when a new note is created

### LINKED ISSUE #2241: [bug] fails to create a new note
hi, I hope this is not a duplicate:

# Current behavior

unable to create a new note
cannot create note

# Expected behavior

able to create a new note
can create new note

# Steps to reproduce

1. enter a search term !important
2. click on 'Make a note Crtl + N' icon
3. make a Markdown Note
4. press Ctrl + N
5.  make a Markdown Note
6. if satisfied, goto step 7, else goto step 4
7. done

# Environment

- Version : 0.11.8
- OS Version and name : 16.04 Ubuntu

Pudge601:
This ensures that the newly created note is visible and selected in the note list.

Fixes #2241

### DIFF
--- a/browser/main/NoteList/index.js
+++ b/browser/main/NoteList/index.js
@@ -170,16 +170,15 @@ class NoteList extends React.Component {
     }
   }
 
-  focusNote (selectedNoteKeys, noteKey) {
+  focusNote (selectedNoteKeys, noteKey, pathname) {
     const { router } = this.context
-    const { location } = this.props
 
     this.setState({
       selectedNoteKeys
     })
 
     router.push({
-      pathname: location.pathname,
+      pathname,
       query: {
         key: noteKey
       }
@@ -198,6 +197,7 @@ class NoteList extends React.Component {
     }
     let { selectedNoteKeys } = this.state
     const { shiftKeyDown } = this.state
+    const { location } = this.props
 
     let targetIndex = this.getTargetIndex()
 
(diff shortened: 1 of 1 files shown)


ZeroX-DG:
You code will not focus the correct note if there is one note that matched in the search box.
boostnote

Pudge601:
Hmm... that's true.

The more I'm looking into this issue, the more confused I'm getting. I'm starting to think that the root issue is that the code for jumping to a note by hash (using the `list:jump` event) doesn't work if the note is not visible in the note list.

That event gets fired when a new note is created (hence this issue), and also when clicking a link to a note from another note (`browser/components/MarkdownPreview.js:826`). I've tested and found the same issue when clicking a note link; if there is a search active which the linked note does not match, then it will not jump to the linked note.

bootnote-jump-search

However, just to add to my confusion, it _is_ possible to jump to a linked note which isn't in the current note list when looking at Starred notes, or notes within a particular Storage folder - in this case, the note is shown in the preview, but doesn't appear in the notes list on the left! I have now idea how this behaviour works, but I'm tempted to say that it is a bug in itself (or at least unintended behaviour), since it leads to a very weird scenario where the note which is in the preview is not the focused note in the note list, and in fact doesn't even appear in the note list.

bootnote-jump-starred
bootnote-jump-storage

This has led me to thinking that perhaps it should be switching to the `/home` route whenever we jump to a note by its hash. This provides consistent behaviour for all of these scenarios;

* Creating a note with search active
* Jumping to a linked note with search active
* Jumping to a linked note while looking at Starred notes
* Jumping to a linked note while looking at a Storage folder

Pudge601:
I've changed my fix in this PR to switch to the '/home' route when jumping to a note by hash. This should solve all of the issues described in my last comment in a consistent manner.

ZeroX-DG:
Is it because of me or the old bug is still there?

ZeroX-DG:
@Pudge601 What's the status of this?

PR #2402, opened 2018-09-15
fix a bug causing right click to copy a note -- fixes #1833

### LINKED ISSUE #1833: Right clicking on a note in the list copies its link to the clipboard
# Current behavior

Right clicking on a note in the note list copies its link to the clipboard.

# Faulty code

Looks like `copyNoteLink` is called rather than curried at https://github.com/BoostIO/Boostnote/blob/f4024f46836d1aa60e10df05ad1455047cd0bbde/browser/main/NoteList/index.js#L514

The code should probably be:

    click: this.copyNoteLink.bind(null, note)

Or (though `this` context is not required for this function):

    click: this.copyNoteLink.bind(this, note)

Or the `copyNoteLink` function could be refactored to use `this` and state to find the selected note(s) like `deleteNote` and `cloneNote` do.

jacobherrington:
Refs #1833 

This is my first PR to Boostnote, I am happy to hear feedback!

### DIFF
--- a/browser/main/NoteList/index.js
+++ b/browser/main/NoteList/index.js
@@ -519,7 +519,7 @@ class NoteList extends React.Component {
         click: this.cloneNote.bind(this)
       }, {
         label: copyNoteLink,
-        click: this.copyNoteLink(note)
+        click: this.copyNoteLink
       })
       if (note.type === 'MARKDOWN_NOTE') {
         if (note.blog && note.blog.blogLink && note.blog.blogId) {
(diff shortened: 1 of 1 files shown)


browser/main/NoteList/index.js
```diff
@@ -684,8 +684,11 @@ class NoteList extends React.Component {
       })
   }
 
-  copyNoteLink (note) {
-    const noteLink = `[${note.title}](:note:${note.key})`
+  copyNoteLink () {
+    const { selectedNoteKeys } = this.state
```
ZeroX-DG: I think you shouldn't re-find the selected note again because the selected note is already been searched at the top of the `handleNoteContextMenu` function. Please re-use the selected note in the `handleNoteContextMenu` function to improve performance for the app.

Issue #2153, opened 2018-06-29
`Error occurs!` appears with an empty hotkey


kobtea:
# Current behavior

`Show/Hide Boostnote` do not allows empty with an error.
`Toggle editor mode` allows empty.

![screenshot from 2018-06-29 23-49-20](https://user-images.githubusercontent.com/1685207/42099295-047bb030-7bf8-11e8-9fbd-da217a26bf58.png)

# Expected behavior

`Show/Hide Boostnote` allows empty like `Toggle editor mode`.

# Steps to reproduce

1. Open `Preference` > `Hotkeys`.
2. Make `Show/Hide Boostnote` empty.
3. Save.

# Environment

- Version : 0.11.6
- OS Version and name : Ubuntu 18.04

PR #2430, opened 2018-09-29
Fix errors when saving a blank hotkey

samherrington:
Saving a blank hotkey showed false error even though hotkey was
saved. This solution solves the issue as it stands, but I
would recommend looking into cleaning up the error handling
surrounding this functionality before adding any new hotkey
fields.

Refs #2153


samherrington:
This also satisfies #2087, I think.

jacobherrington:
@ZeroX-DG I am fairly sure the error messages correctly reflect what is happening now. If the success message shows up, the hotkey seems to be consistently set to what is in the text box.

ZeroX-DG:
Well, clearly that's the current behavior :smile: Or we will have to prevent users from saving blank hotkeys, however, there're requests that people want to "disable" that hotkey by leaving it blank. So do you have other ideas other than sticking to the current behavior? If not then I think showing a message to notify the user is not a bad idea.

samherrington:
@ZeroX-DG The new behavior allows users to save blank hotkeys and displays the success message when they do (because they have successfully saved a blank hotkey). Before this fix, a blank hotkey returned an error.

ZeroX-DG:
boostnote_bug_hotkey
@samherrington is this the intended behavior?

samherrington:
@ZeroX-DG No, with this PR you should see "Successfully applied!" when either or both of the hotkeys are blank.

ZeroX-DG:
Haha @jacobherrington TIL

Issue #2010, opened 2018-05-30
cannot search by Chinese


leafiy:
![default](https://user-images.githubusercontent.com/870449/40704443-b129f760-641a-11e8-858a-60705af098a5.png)

IssuehuntBot:
@boostio funded this issue with $20. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/2010)

IssuehuntBot:
@ivanifchen has submitted [output](https://github.com/BoostIO/Boostnote/pull/2433). [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/2010)

ThaddeusJiang:
Only happen in production, I have checked the develop env, it works correctly

PR #2433, opened 2018-09-29
Cannot search by Chinese

IvanIFChen:
Use "compositionend" listener to check when done typing Chinese, Japanese, or Korean words.

Result GIF:
http://recordit.co/iLSWqBhJY9

This is a PR for fixing this issue:
https://github.com/BoostIO/Boostnote/issues/2010
https://issuehunt.io/repos/53266139/issues/2010

Reference:
https://developer.mozilla.org/en-US/docs/Mozilla/IME_handling_guide
https://codepen.io/CarterTsai/pen/IFnGz
https://developer.mozilla.org/en-US/docs/Web/Events/compositionend

### DIFF
--- a/browser/main/TopBar/index.js
+++ b/browser/main/TopBar/index.js
@@ -16,12 +16,18 @@ class TopBar extends React.Component {
       searchOptions: [],
       isSearching: false,
       isAlphabet: false,
-      isIME: false,
+      isIMEDone: false,
       isConfirmTranslation: false
     }
 
     this.focusSearchHandler = () => {
       this.handleOnSearchFocus()
+    }
+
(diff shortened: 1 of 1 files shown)


amedora:
doesn't work as expected.
compositionend

AWolf81:
I think we can close this PR. @amedora can you please check if it's working on master.

`composition` events are added in PR #3037.

amedora:
> I think we can close this PR. @amedora can you please check if it's working on master.
I confirmed that current master works perfect. This PR should be closed.

Rokt33r:
@AWolf81 @amedora Thanks!

PR #2434, opened 2018-09-29
Remove console.logs

samherrington:
Using console in production is generally undesirable due to
performance loss and security concerns. Errors were changed
to console.error and console.logs were removed.

### DIFF
--- a/browser/main/lib/ConfigManager.js
+++ b/browser/main/lib/ConfigManager.js
@@ -87,7 +87,6 @@ function validate (config) {
 }
 
 function _save (config) {
-  console.log(config)
   window.localStorage.setItem('config', JSON.stringify(config))
 }
 
--- a/browser/main/lib/ConfigManager.js
+++ b/browser/main/lib/ConfigManager.js
@@ -87,6 +87,7 @@ function validate (config) {
 }
 
 function _save (config) {
+  console.log(config)
   window.localStorage.setItem('config', JSON.stringify(config))
 }
 
--- a/browser/main/lib/dataApi/resolveStorageData.js
+++ b/browser/main/lib/dataApi/resolveStorageData.js
@@ -32,12 +32,11 @@ function resolveStorageData (storageCache) {
   const version = parseInt(storage.version, 10)
   if (version >= 1) {
     if (version > 1) {
-      console.log('The repository version is newer than one of current app.')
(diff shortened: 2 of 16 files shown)


browser/main/lib/dataApi/resolveStorageData.js
```diff
@@ -32,12 +32,11 @@ function resolveStorageData (storageCache) {
   const version = parseInt(storage.version, 10)
   if (version >= 1) {
     if (version > 1) {
-      console.log('The repository version is newer than one of current app.')
+      return
```
ehhc: return will not create a promise anymore! that changes the baviour of this method!

dev-scripts/dev.js
```diff
@@ -37,7 +37,7 @@ function startServer () {
           resolve()
         } else {
           if (!firstRun) {
-            console.log(stats.compilation.errors[0])
+            return
```
ZeroX-DG: That console.log is actually there by design :smile: Please undo you change :smile:

browser/main/lib/ConfigManager.js
```diff
@@ -87,7 +87,6 @@ function validate (config) {
 }
 
 function _save (config) {
-  console.log(config)
```
ZeroX-DG: Should we leave the console.log here? Maybe it will come in handy in the future when the users hit some issue related to the config and they can capture this console.log?  @Rokt33r @sosukesuzuki

samherrington: In general, console.log won't offer much information to users unless they are devs, and if they are devs, they can easily add it back in. In particular, doing something like logging AWS credentials is pretty bad from a security perspective.

ZeroX-DG: Yes, clearly logging AWS credentials is a bad idea (maybe it was for debugging purpose have the team forgot to remove it :smile: ) I also agree that console logging the config is a bad idea, but before we can invent some kind of a button to take the current config and attach it to the new issue page, leaving the config in the console for people to take a screenshot of is not a bad idea.

samherrington: @ZeroX-DG I've made the changes you requested!

samherrington:
All good man! Glad to be able to work with you guys on this!

Issue #2361, opened 2018-09-02
Reconsiering default hotkeys in macOS


0xBADDCAFE:
# Current behavior

In macOS, "Toggle editor mode" default hotkey is assigned Cmd+M. But it conflicts macOS's "Minimize the front window to the Dock." default hotkey so do not works on Boostnote.

Refs: https://support.apple.com/en-us/HT201236

And "Show/Hide Boostnote" default hotkey is assigned "Cmd+Alt+L". If another application also use this keybinding, "Show/Hide Boostnote" hotkey works system-widely, so the app's hotkey do not work. Overriding system-wide hotkey by default keybinding confuses user. So I think this hotkey should disabled by default.

(Note: "Cmd+Alt+L" used by IntelliJ platform to call "Auto formatting code". Boostnote is programmer friendly, so not a few people uses with IntelliJ platform.)

Of course, these keybinding can change, but if clear to want to disable, causes another issue's problem (https://github.com/BoostIO/Boostnote/issues/2153)

# Expected behavior

"Toggle editor mode" default hotkey is assigned without Cmd+M.

"Show/Hide Boostnote" default hotkey is disabled.

# Steps to reproduce

1. keydown Cmd+M
2. minimize Boostnote and do not toggled editor mode.

- - -

1. Open another application that has "Cmd+Alt+L" keybinding
2. keydown "Cmd+Alt+L"
3. Boostnote is focused and another app do not have effects.

# Environment

- Version : 0.11.8
- OS Version and name : macOS High Sierra 10.13.6

Rokt33r:
Close this in favor of https://github.com/BoostIO/Boostnote/pull/2437

PR #2437, opened 2018-09-30
Change Toggle Editor Mode hotkey on OS X

jacobherrington:
By default the Toggle Editor Mode hotkey was set to Command + M,
which is a great and mnemonic hotkey. Unfortunately that hotkey
is also the hotkey to minimize an application.

This commit changes the default Toggle Editor Mode hotkey to
Command + Option + M on OS X, which will not minimize the window
and maintains the mnemonic pattern.


jacobherrington:
This PR resolves #2361.

I disagree with #2361 on the default Show/Hide hotkey setting because I think that setting a hotkey so it doesn't conflict with other programs is on the user.

ZeroX-DG:
I also agree with you about the show/hide Boostnote hotkey.

PR #2440, opened 2018-09-30
Clone notes to the same folder

jacobherrington:
Currently notes are cloned to the default folder if the clone
note link is clicked from "All Notes" or "Starred." They are only
cloned to the same folder as the original note if the clone note
link is clicked inside of that folder.

This commit changes this behavior such that a clone is always in
the same folder as the note it was cloned from.

### DIFF
--- a/browser/main/NoteList/index.js
+++ b/browser/main/NoteList/index.js
@@ -659,7 +659,7 @@ class NoteList extends React.Component {
     dataApi
       .createNote(storage.key, {
         type: firstNote.type,
-        folder: folder.key,
+        folder: selectedNotes[0].folder,
         title: firstNote.title + ' ' + i18n.__('copy'),
         content: firstNote.content
       })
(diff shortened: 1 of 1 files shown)


browser/main/NoteList/index.js
```diff
@@ -659,7 +659,7 @@ class NoteList extends React.Component {
     dataApi
       .createNote(storage.key, {
         type: firstNote.type,
-        folder: folder.key,
+        folder: firstNote.folder,
```
ehhc: I think, after you'ver changed the location of the note, the attachments  will not work anymore because they have been copied to the wrong location (line 667 of this file). Can you have a look at that?

jacobherrington: Good call out - I will test this.

jacobherrington: @ehhc it seems to work for me, are you able to break this?

ehhc: @jacobherrington to be honest, it was only a guess that it might break. I haven't tried. Maybe it works fine :)
What might be difficult: if you clone a note to a folder in a different storage (is that even possible??).. But also only a guess.. :) maybe everything is fine :)

Issue #2108, opened 2018-06-20
[Feature request] Add "Copy" button


NotFounds:
It would be cool to add a button which to copy a text in snippet page.

IssuehuntBot:
@boostio funded this issue with $20. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/2108)

IssuehuntBot:
@jacobherrington has started working. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/2108)

IssuehuntBot:
@jacobherrington has submitted [output](https://github.com/BoostIO/Boostnote/pull/2445). [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/2108)

IssuehuntBot:
@kazup01 has rewarded. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/2108)

PR #2445, opened 2018-09-30
Add a Copy button the Snippets tab

jacobherrington:
This adds a button to copy snippets.

Solves #2108


ZeroX-DG:
Can you show some message when the snippet is copied? For example:
https://github.com/BoostIO/Boostnote/blob/2ccd00a378cdc1511ac17e4b40470976d1a1e1a4/browser/components/MarkdownPreview.js#L694-L697

PR #2448, opened 2018-09-30
Fix bug caused by poor regex

jacobherrington:
I created this bug in #2435. This should fix it.

screen shot 2018-09-30 at 6 18 12 pm 1


jacobherrington:
@Rokt33r this should get merged in before the next release, there is a bug that will happen to all hotkeys in the next release.

PR #2456, opened 2018-10-01
lint before commit

daiyam:
This change is running the `lint` command before a commit.


jacobherrington:
I personally think forcing lint on commits is a bit heavy handed - our CI already runs lint so I don't think it's totally necessary.

ZeroX-DG:
I think with pre-commit hook we can detect the problem right away without waiting for the CI to fail and prevent some unnecessary commit like "fixed lint" :smile:

ZeroX-DG:
LGTM but can you discard `yarn.lock`? Also, I think running test in the pre-commit hook is a bit heavy for small changes (There's still some chance that small changes can cause failed tests, but I think leaving it to the CI is the best). @daiyam

daiyam:
@ZeroX-DG I agree with you since the tests are heavy.

Rokt33r:
@ZeroX-DG @daiyam I think tests should be executed before pushing. So how do you think add pre-push hook for testing?

PR #2461, opened 2018-10-02
add collapsible block

daiyam:
The change adds a collapsible block.

When it's closed (by default):
screenshot

When opened:
screenshot


ZeroX-DG:
@daiyam why don't we use:
<details>
  <summary>Click to expand</summary>
  whatever
</details>
```
<details>
  <summary>Click to expand</summary>
  whatever
</details>
```

daiyam:
Because
```
?> ## Click to expand
?>
?> whatever
```
is more readable than
```
<details>
  <summary>## Click to expand</summary>
  whatever
</details>
```

nagledb:
The fact that there isn't already a markdown-it plugin for this suggests that it's not something widely wanted. I'm skeptical of adding a completely new and unique markdown syntax for a very niche edge case.

Putting that aside, I'd suggest making your own markdown-it plugin for it and then having Boostnote use the plugin, rather than embedding the whole thing into Boostnote. That way whatever syntax it ends up with has a chance to catch on outside of just Boostnote.

Regarding the Markdown syntax:
- I don't find the syntax intuitive.
- I think that the summary line should be marked differently from the content.
- There should be a way to denote whether it starts in the open or closed state.

Regarding the Rendered content:
- I really don't like the eyeball icon.
- The default styling that you get if you just use `<details>` and `<summary>` directly looks good enough to me. I don't see any need to change it.

daiyam:
I've removed the custom styling.

And for ***why I'm adding a new block*** is that I find that html code is not reader friendly. So I always try to avoid html code.

ZeroX-DG:
@Rokt33r what do you think about this?

Rokt33r:
I agree with @nagledb 's idea. It should be presented as a plugin of markdown-it first. I'm not going to merge this until the new syntax becomes popular enough.

PR #2462, opened 2018-10-02
fix small visual bug in tag autocomplete

daiyam:
This change fixes a small visual bug in the tag's autocomplete.

Before:
screenshot

After:
screenshot


ZeroX-DG:
LGTM :tada:

PR #2471, opened 2018-10-06
add automatic collapsible code blocks & headings

daiyam:
This change allows code blocks and headings to be automatically collapsed.

How to configure:
screenshot

---

screenshot

The result:
screenshot


ReK42:
1) The icons for btn-clipboard and btn-expand/btn-collapse do not render properly:
image

Same note in 0.11.9:
image

2) The btn-clipboard icon moves when the code block is expanded:
image

3) The expand/collapse button is at the bottom of the code block. Convention for expand/collapse icons is to have them at the top and to the left of the container (e.g. \<details\> tag, code folding, etc.).

Here are the display settings I used:
image

nagledb:
Please use the default styling for `<summary>` and `<details>` rather than the custom eyeball thing. The default styling is more consistent with what's used elsewhere and looks more professional.

End users can always customize their CSS to add the eyeballs if they want.

daiyam:
I have update the styling to use the standard one:
screenshot

PR #2473, opened 2018-10-08
handle all dropped images

daiyam:
This change allows to drop multiple images.

And for better memory usage, use `webp` instead of `png`


daiyam:
The `webp` conversion is no good. It will cause some issues when exporting (bad extension, image not recognized by the system)

PR #2474, opened 2018-10-08
disable hyperlink for attachments

daiyam:
This change disables the following popover on attached images:
screenshot

### DIFF
--- a/extra_scripts/codemirror/addon/hyperlink/hyperlink.js
+++ b/extra_scripts/codemirror/addon/hyperlink/hyperlink.js
@@ -52,7 +52,8 @@
 
       if (className.indexOf('cm-url') !== -1) {
         const match = /^\((.*)\)|\[(.*)\]|(.*)$/.exec(el.textContent)
-        return match[1] || match[2] || match[3]
+        const url = match[1] || match[2] || match[3]
+        return /^:storage\//.test(url) ? null : url
       }
 
       return null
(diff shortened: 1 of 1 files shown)


extra_scripts/codemirror/addon/hyperlink/hyperlink.js
```diff
@@ -52,7 +52,8 @@
 
       if (className.indexOf('cm-url') !== -1) {
         const match = /^\((.*)\)|\[(.*)\]|(.*)$/.exec(el.textContent)
-        return match[1] || match[2] || match[3]
+        const url = match[1] || match[2] || match[3]
+        return /^:storage\//.test(url) ? null : url
```
ehhc: do you mind to use attachmentManagement.STORAGE_FOLDER_PLACEHOLDER instead of hard-coding `:storage` here?

daiyam: @ehhc it's not possible
`require('../browser/main/lib/dataApi/attachmentManagement')` throws the error `Cannot find module 'browser/lib/findStorage'`

`hyperlink.js` is loaded by `Chromium`
`attachmentManagement.js` is written to be loaded by `node`

ehhc: if it is really not possible (i am not a nodejs/electron pro, so i'm not sure about that) could you at least add an comment pointing to that file? That would help a lot in case somebody will refactor it again (e.g. in a few years..)

daiyam: @ehhc sure, I can do that

PR #2476, opened 2018-10-08
alternate background color for table rows

daiyam:
This change alternates the background color of the table rows.

screenshot


ZeroX-DG:
Should we append css to all CodeMirror themes? Or can we just include another separate css file for that?

daiyam:
@ZeroX-DG I've updated how to calculate the alternative color. Now, it's used a mix of the background color with a little bit of the text color.

ZeroX-DG:
LGTM :+1:

Issue #2472, opened 2018-10-08
Part of InfoPanel is hidden when EDITOR_PREVIEW mode


masaodev:
# Current behavior

InfoPanel hides only when in EDITOR_PREVIEW mode.
Probably this behavior is from #2320

![cap](https://user-images.githubusercontent.com/41692876/46589005-db443c00-cadf-11e8-9cd9-2a550823a2e7.gif)

# Environment

- Version : Latest
- OS Version and name : Windows 10

PR #2477, opened 2018-10-09
fix visual bug with preview

daiyam:
This change fixes hidden the note's infos or the tag autocomplete behind the note's preview.

It's fixing #2472


ZeroX-DG:
LGTM :tada:

Issue #2483, opened 2018-10-10
Ubuntu 18.10 segmentation fault


crakoucas:
# Current behavior
In Ubuntu 18.10 when i launch Boostnote i have this error: [1]    7816 segmentation fault (core dumped)  boostnote.
It s an Electon problem from the last [glibc version 2.28](https://github.com/electron/electron/issues/13972). 
VsCode have the same error and fixed this on the last version 1.28.0.

Thanks 

# Expected behavior

# Steps to reproduce

1. Install with deb package
2. Launch 
3. Error : [1]    7816 segmentation fault (core dumped)  boostnote

# Environment

- Version : last 0.11.9
- OS Version and name : Ubuntu 18.10

Ok Fix when you Update Electron to version 2.0.9 => [Pull request](https://github.com/BoostIO/Boostnote/pull/2484)

PR #2484, opened 2018-10-10
fix segmentation fault with glibc 2.28

crakoucas:
This is fix :bug: 
ERRROR : [1] 7816 segmentation fault (core dumped) in Ubuntu 18.10 and i think all linux distro with glibc 2.28

[Issue](https://github.com/BoostIO/Boostnote/issues/2483)

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "dev": "node dev-scripts/dev.js"
   },
   "config": {
-    "electron-version": "2.0.9"
+    "electron-version": "3.0.3"
   },
   "repository": {
     "type": "git",
@@ -56,7 +56,7 @@
     "codemirror": "^5.39.0",
     "codemirror-mode-elixir": "^1.1.1",
     "electron-config": "^1.0.0",
-    "electron-gh-releases": "^2.0.2",
+    "electron-gh-releases": "^2.0.4",
     "escape-string-regexp": "^1.0.5",
     "file-uri-to-path": "^1.0.0",
     "file-url": "^2.0.2",
@@ -121,7 +121,7 @@
     "css-loader": "^0.19.0",
     "devtron": "^1.1.0",
     "dom-storage": "^2.0.2",
-    "electron": "2.0.9",
+    "electron": "3.0.3",
     "electron-packager": "^12.0.0",
     "eslint": "^3.13.1",
     "eslint-config-standard": "^6.2.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -121,7 +121,7 @@
     "css-loader": "^0.19.0",
     "devtron": "^1.1.0",
     "dom-storage": "^2.0.2",
-    "electron": "2.0.7",
+    "electron": "2.0.9",
```
daiyam: @crakoucas can you update it to the lastest one. Thx

crakoucas: No problem

daiyam:
@crakoucas I was more thinking about the v2.0.11 but v3.0.3 is even better.

But sadly, there is a small bug due to it (when the window is restored, its size is wrong).
Here the patch to fix it:

```diff
diff --git a/dev-scripts/dev.js b/dev-scripts/dev.js
index 000a1bfd..9698a2fe 100644
--- a/dev-scripts/dev.js
+++ b/dev-scripts/dev.js
@@ -49,7 +49,7 @@ function startServer () {
 }
 
 function startElectron () {
-  spawn(electron, ['--hot', './index.js'])
+  spawn(electron, ['--hot', './index.js'], { stdio: 'inherit' })
     .on('close', () => {
       server.close()
     })
diff --git a/lib/main-window.js b/lib/main-window.js
index fa54d5ce..86b70bea 100644
--- a/lib/main-window.js
+++ b/lib/main-window.js
@@ -14,6 +14,7 @@ const mainWindow = new BrowserWindow({
   y: windowSize.y,
   width: windowSize.width,
   height: windowSize.height,
+  useContentSize: true,
   minWidth: 500,
   minHeight: 320,
   autoHideMenuBar: showMenu,
```

crakoucas:
@daiyam Done, but i can't reproduce the bug at home. Can you test it ?

daiyam:
@crakoucas it's good for me (on macOS).

Rokt33r:
I'll check this on Windows and Linux now.

Rokt33r:
I confirmed it works well on Windows

KaiNissen:
I can confirm that this patch fixes the startup issue on Ubuntu 18.10.

matthew-nm:
I am having the same issue on Kubuntu 18.10. Is this fix going to be merged?

Rokt33r:
Btw sorry for being late... I'd just forgot about it... If you want something really badly, please ping me on our slack!

PR #2492, opened 2018-10-12
Remove emacs keymap Ctrl-V binding to allow pasting

### LINKED ISSUE #1406: I can not paste in emacs mode
I am using boostnote on windows.
I change keymap to emacs and use it.
I want to copy and paste text information on a web page onto boostnote, but if I copy it with Ctrl + c and paste it with Ctrl + y on boostnote, I can not paste anything.
The text on boostnote can be copied with Alt + w and pasted with Ctlr + y.

amireldor:
Fixes #1406 

Hey friends, this fix is borrowed from a CodeMirror issue: https://github.com/codemirror/CodeMirror/issues/5321#issuecomment-375232302

It's a one-liner that should be run only once. I wasn't sure where to put this code so I tried a main entry that is executed once: the mounting of the main window.

**If you think of a better place to put this please let me know so I can fix**. I'm not familiar too much with boostnote code.

### DIFF
--- a/browser/main/Main.js
+++ b/browser/main/Main.js
@@ -168,6 +168,8 @@ class Main extends React.Component {
       }
     })
 
+    delete CodeMirror.keyMap.emacs["Ctrl-V"]
+
     eventEmitter.on('editor:fullscreen', this.toggleFullScreen)
   }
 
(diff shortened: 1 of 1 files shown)


Rokt33r:
I don't use emacs but, it seems some of emacs users want to have this badly. Yeah, let's try it for now.

PR #2497, opened 2018-10-15
delete tag through context menu

daiyam:
This change adds the possibility to delete a tag through its context menu.

screenshot

### DIFF
--- a/browser/main/Detail/MarkdownNoteDetail.js
+++ b/browser/main/Detail/MarkdownNoteDetail.js
@@ -65,7 +65,9 @@ class MarkdownNoteDetail extends React.Component {
   }
 
   componentWillReceiveProps (nextProps) {
-    if (!this.state.isMovingNote && (nextProps.note.key !== this.props.note.key || nextProps.note.tags.length < this.props.note.tags.length)) {
+    const isNewNote = nextProps.note.key !== this.props.note.key
+    const hasDeletedTags = nextProps.note.tags.length < this.props.note.tags.length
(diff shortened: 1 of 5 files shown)


browser/main/Detail/MarkdownNoteDetail.js
```diff
@@ -65,7 +65,7 @@ class MarkdownNoteDetail extends React.Component {
   }
 
   componentWillReceiveProps (nextProps) {
-    if (nextProps.note.key !== this.props.note.key && !this.state.isMovingNote) {
+    if (!this.state.isMovingNote && (nextProps.note.key !== this.props.note.key || nextProps.note.tags.length < this.props.note.tags.length)) {
```
ZeroX-DG: This condition is really long and may confuse new contributors, can you assign some parts of the condition into variables with meaningful names?

daiyam: So that's is the reason for temporary variables I've seen in the code.
It's to increase the readability of the code  👍 

```
if (
  !this.state.isMovingNote &&
  (
    // if it's a new note
    nextProps.note.key !== this.props.note.key ||
    // if there is deleted tags
    nextProps.note.tags.length < this.props.note.tags.length
  )
) {
```

```
const isNewNote = nextProps.note.key !== this.props.note.key
const hasDeletedTags = nextProps.note.tags.length < this.props.note.tags.length
if (!this.state.isMovingNote && (isNewNote || hasDeletedTags)) {
```

I can see why it is more readable 😉

PR #2498, opened 2018-10-15
create note with selected tags

daiyam:
This change creates a note with the tags selected in the SideNav.

### DIFF
--- a/browser/lib/newNote.js
+++ b/browser/lib/newNote.js
@@ -3,14 +3,18 @@ import dataApi from 'browser/main/lib/dataApi'
 import ee from 'browser/main/lib/eventEmitter'
 import AwsMobileAnalyticsConfig from 'browser/main/lib/AwsMobileAnalyticsConfig'
 
-export function createMarkdownNote (storage, folder, dispatch, location) {
+export function createMarkdownNote (storage, folder, dispatch, location, params) {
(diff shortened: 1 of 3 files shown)


ZeroX-DG:
Should we add a checkbox to notify and allow users to not create a new note with selected tags?

daiyam:
I could add an option so that new notes use the selected tags or not. Would it be ok?

ZeroX-DG:
Yeah @daiyam that sounds good!

daiyam:
here we go 😉

ZeroX-DG:
Hey @daiyam is it me or your code causing this problem?
bb

daiyam:
@ZeroX-DG sadly, no. The issue is on master...

If you add `console.log(location.pathname)` at the top of the method `tagListComponent`(line 209) in the file `/Boostnote/browser/main/SideNav/index.js`, when you on the second tag, you will see in the console:
```
/tags/new hello
/tags/new%20hello
/tags/new%20hello
/tags/new%20hello
```

Beside the fact that the list is updated 4 times for 1 click, you can see that the spaces have been replaced by `%20`, so the code can't get individual tags (it splits on space...).

It happens since the merge `Merge pull request #2484 from crakoucas/fix-electron-glibc-segmentation-fault` (106caf2b3e960acf3f7d8ba808c802e3e52eb9f9).
Maybe due to the newer electron...

PR #2499, opened 2018-10-15
store editor view into config

daiyam:
When the editor's option `switchPreview` is on `RIGHTCLICK`, this change stores the editor view (`CODE` or `PREVIEW `) in the config.


browser/components/MarkdownEditor.js
```diff
@@ -84,6 +83,10 @@ class MarkdownEditor extends React.Component {
           this.refs.preview.focus()
         }
         eventEmitter.emit('topbar:togglelockbutton', this.state.status)
+
+        const newConfig = Object.assign({}, this.props.config)
```
ZeroX-DG: Please change `this.props.config` to `config`

daiyam: Done. Thx

Issue #975, opened 2017-10-13
Un-check all checkboxes


UkiahSmith:
For notes with checkboxes it would be a nice new feature to have a menu item, and key command, to uncheck all checkboxes.

My use case: I have a daily checklist of things I need to do every day. I check each item as I do them throughout the day. At the end of the day I have to manually uncheck all the checkboxes to prepare the list for the next morning. As I grow and refine my list it becomes more time consuming to uncheck all the boxes. A new command to uncheck all boxes at one time would make this rather quick, and painless.

IssuehuntBot:
@jsmwoolf has started working. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/975)

IssuehuntBot:
@makibishi0212 has submitted [output](https://github.com/BoostIO/Boostnote/pull/2503). [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/975)

IssuehuntBot:
@kazup01 has rewarded. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/975)

PR #2503, opened 2018-10-17
Add button clear all checkboxes

makibishi0212:
This change adds button un-check all checkboxes (issue #975)

uncheck2

### DIFF
--- a/browser/components/TodoListPercentage.js
+++ b/browser/components/TodoListPercentage.js
@@ -12,7 +12,7 @@ import styles from './TodoListPercentage.styl'
  */
 
 const TodoListPercentage = ({
-  percentageOfTodo, onClick
+  percentageOfTodo, onClearCheckboxClick
 }) => (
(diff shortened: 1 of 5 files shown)


browser/components/TodoListPercentage.js
```diff
@@ -12,19 +12,23 @@ import styles from './TodoListPercentage.styl'
...
+    <div styleName='todoClear'>
+      <p styleName='todoClearText' onClick={(e) => onClick(e)}>clear</p>
+    </div>
   </div>
 )
 
 TodoListPercentage.propTypes = {
-  percentageOfTodo: PropTypes.number.isRequired
+  percentageOfTodo: PropTypes.number.isRequired,
+  onClick: PropTypes.func.isRequired
```
ZeroX-DG: I don't think `onClick` is a good name, I think you should rename it to something like `onClearCheckboxClick`?

makibishi0212: Thanks for your review.
I fix it.

PR #2511, opened 2018-10-19
Fixed "Save as HTML" to respect blockquotes.

### LINKED ISSUE #1922: Indentation tag [ > ]  is not exported right format to HTML
# Current behavior
alt text
# Expected behavior
alt text
# Steps to reproduce

1.  Export the markdown as HTML file
P.S. I don't need that green Indentation bar, but the actual indentation spaces
# Environment

- Version : Boostnote 0.11.4
- OS Version and name : WIn10

### LINKED ISSUE #2514: Inline code renders incorrectly in HTML export
# Current behavior
Inline code containing angle brackets (such as HTML or equalities), or ampersands render correctly in the BoostNote editor but when exported to HTML, they are 'over-escaped' and do not appear correctly:

In BoostNote:

image

image

When exported as HTML and opening in a browser it does not display correctly:

image

Examining the HTML you can see that the ampersands have been escaped one too many times:

image

# Expected behavior

Inline code containing special characters should be escaped to the correct level in HTML export, like so:

image
 
which renders correctly in a browser.

# Steps to reproduce

1. Create a new Markdown note 
2. Enter some inline code containing angle brackets or ampersands (such as the example above)
3. File -> Export to HTML
4. View the exported HTML in a browser and see the code does not display correctly
5. Open the exported HTML in a code editor and note that the escaping is not correct

# Environment

- Version : 0.11.9
- OS Version and name : Windows 10 x64

nagledb:
When saving as HTML, the raw text was being HTML escaped prior to
running it through markdown. This was preventing blockquotes from
rendering because greater-than signs were escaped. The HTML escaping
seems unneeded since the markdown renderer already takes care of it,
and it also wasn't consistent with how markdown was being rendered
within the application.

Fixes #1922
Fixes #2514

### DIFF
--- a/browser/components/MarkdownPreview.js
+++ b/browser/components/MarkdownPreview.js
@@ -16,7 +16,6 @@ import convertModeName from 'browser/lib/convertModeName'
 import copy from 'copy-to-clipboard'
 import mdurl from 'mdurl'
 import exportNote from 'browser/main/lib/dataApi/exportNote'
-import { escapeHtmlCharacters } from 'browser/lib/utils'
 import context from 'browser/lib/context'
 import i18n from 'browser/lib/i18n'
 import fs from 'fs'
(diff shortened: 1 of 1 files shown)


nagledb:
Rebased against current [head](f92da165445bcad56a2f0188846eb795c07e2d84) (without any changes).

Verified that this also fixes #2514

ZeroX-DG:
LGTM but I still need @Rokt33r to review this.

Rokt33r:
@nagledb 
> The HTML escaping seems unneeded since the markdown renderer already takes care of it,
I don't think it is true. Markdownit doesn't sanitize anything. But, I'm okay with discarding this because most of users open the exported html on the latest browser like Chrome. So, it should be fine.

PR #2532, opened 2018-10-24
update dependencies

btoo:
update grunt so its binary can be run without installing it globally

update electron

update jest

yarn fix


btoo:
i forked Boostnote so i could add my own mods and i had to update a bunch of dependencies in doing so. im submitting this PR in case you guys might want to use some of the newer features of webpack, babel, etc.

Rokt33r:
@ZeroX-DG Is this pr still valid? Its more than a year ago. I'm quite skeptical

ZeroX-DG:
@Rokt33r I was going go tell the PR owner to resolve the conflict to remove the invalid parts and maybe review what's still valid because this PR contain an update for webpack to webpack 4 and other dependencies. But if you think that kind of upgrade needed to be done by you or you have different plan on upgrading dependencies then feel free to close this PR :+1:

Issue #1811, opened 2018-04-15
Scrolling not sync between Editor and Preview


blinker57:
# Current behavior
I have been experiencing cases where preview does not sync its scroll position to where the editor view is. It seems to act up when I use several small code block mixed with regular text and markdown.

Love the editor, thanks!

# Expected behavior
I expect the preview to always show where you are typing in the editor.

# Steps to reproduce
1. Write a lot of text mixed in with code block samples now and then 
2. After a while, you will find that where you are typing in the editor is not in view in the preview.

# Environment
- Version : 0.11.4
- OS Version and name : Windows 7 64-bit, Intel i5, 32GB memory

nagledb:
This was bugging me today so I spent some time investigating. It looks like syncing is not content-aware. Instead, it simply syncs up the two windows based on where you are in terms of pixel distance in the rendered version of the scrolled window.

So for example, let's say your markdown window's content has a total rendered height of 10,000 pixels and your preview window's content has a total rendered height of 12,000 pixels. You scroll your markdown window so that the top of the window is at pixel height 2,500. That means the top of your markdown window is now 0.25 (25%) of the way through the document. The rendered window then gets scrolled so that its top is also at 0.25 (pixel height 3,000).

This doesn't necessarily work well because the rendered height of a given markdown element can vary considerably. Some examples: Some lines of markdown don't get rendered at all. Headings are much taller than paragraph text. Word-wrapping can cause the number of lines to change.

As a real-world example for how bad that can be: I'm currently editing a markdown document that has 418 lines of content. I scrolled my markdown pane so that lines 271 to 297 were in view; the preview window showed me the content for lines 334 to 351. Those ranges don't even overlap.

A better solution might be to sync the windows based on the content. For example, look at the line number for what's at the top and then scroll the other window so that the same line number is at the top. I spent a bit of time trying to implement that today but couldn't get it to work right.

IssuehuntBot:
@adlersantos funded this issue with $10. [See it on IssueHunt](https://issuehunt.io/repos/53266139/issues/1811)

ytee:
It would be good to disable / enable scroll sync with a button

PR #2542, opened 2018-10-26
Added option to disable scroll sync.

nagledb:
While this doesn't solve the issues in #1811 it does at least allow us to regain some manual control over scrolling when sync isn't working well.

It may also be a useful feature in its own right. Sometimes I find myself wanting to scroll up to look at something else as reference while I'm editing. This makes that easier.

### DIFF
--- a/browser/components/MarkdownSplitEditor.js
+++ b/browser/components/MarkdownSplitEditor.js
@@ -30,6 +30,8 @@ class MarkdownSplitEditor extends React.Component {
   }
 
   handleScroll (e) {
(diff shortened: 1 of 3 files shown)


nagledb:
Good idea on the hotkey. I'll see if I can figure how to to do that later.

ZeroX-DG:
@nagledb OK, we will merge it later. Hope that you can implement hotkey for this in the next PR.
Here is my PR for the hotkey system:
https://github.com/BoostIO/Boostnote/pull/1929

PR #2197, opened 2018-07-09
Improve bracket autoclosing

chang:
Fixes #1428.

boost_brackets

Improves bracket editing by autoclosing markdown patterns for: LaTeX, italic, bold, code snippets, and Python docstrings.

Added pairs:
- `$$` (latex)
- `**` (markdown italic and bold)

Added triplets:
- ` ``` ` (code snippets)
- `"""` and `'''` (python docstrings)

PR #2545, opened 2018-10-27
Add ~ and _ to autoclosing brackets

Danmou:
#2197 didn't add the tilde, used for strikethrough, and the alternative italic/bold underscore. Adding the underscore might be a bit annoying to some, but it's only a problem if you want to start a word with underscore (and you can always delete the extra underscore of course).


ZeroX-DG:
@Danmou very sorry for the late review, can you resolve the conflict?

Danmou:
Done.

Issue #2561, opened 2018-10-31
render pane and editor pane have dislocation


hzhangxyz:
# Current behavior

when there is many image in a single note,
the editor pane and render pane have some dislocation,
as the screenshop shows

![image](https://user-images.githubusercontent.com/11623447/47763890-8c17b280-dcfd-11e8-8d73-f43cf2ae2a76.png)

# Expected behavior

what's expected is when scrolling editor pane to the bottom, the render pane also scrolls to the bottom

# Steps to reproduce
just pasting many many image into a single note can reproduce this issue

# Environment

- Version : 0.11.9
- OS Version and name : archlinux

# something else to say

maybe it is difficult to locate every line of markdown, but please let right pane scroll to bottom if left pane scolled to bottom

hzhangxyz:
OK, it is not the image's fault, it seems in browser/components/MarkdownSplitEditor.js 
`const distance = (targetHeight * srcTop / srcHeight) - targetTop` is wrong when left pane height is smaller then right pane

IssuehuntBot:
@boostio funded this issue with $10. [See it on IssueHunt](https://issuehunt.io/repos/53266139/issues/2561)

IssuehuntBot:
@hzhangxyz has started working. [See it on IssueHunt](https://issuehunt.io/repos/53266139/issues/2561)

Flexo013:
This is solved by #2405 which is part of Boostnote Legacy version 0.16.0.

Thanks for your contribution @daiyam ❤️ (you will receive the bounty soon)

PR #2405, opened 2018-09-17
Better scroll sync between the editor and the preview in the SplitEditor

daiyam:
## Description

This change implements better positioning between the editor and the preview in the SplitEditor (particularly with lot of images).

When scrolling, it calculates the position of the shown block (at 1/3 below the top) based on the previous blocks' heights.
When the cursor is active, the preview is moved at a position so we can see what we are typing.

## Issue fixed

- #2734
- #2561

## Type of changes

- :radio_button: Bug fix (Change that fixed an issue)
- :white_circle: Breaking change (Change that can cause existing functionality to change)
- :radio_button: Improvement (Change that improves the code. Maybe performance or development improvement)
- :white_circle: Feature (Change that adds new functionality)
- :white_circle: Documentation change (Change that modifies documentation. Maybe typo fixes)

## Checklist:

- :radio_button: My code follows [the project code style](docs/code_style.md)
- :white_circle: I have written test for my code and it has been tested
- :radio_button: All existing tests have been passed
- :white_circle: I have attached a screenshot/video to visualize my change if possible

PR #2562, opened 2018-10-31
scroll to bottom if another pane scrolled

hzhangxyz:
solve (#2561)


hzhangxyz:
close because (#2405) is better solution

Issue #2530, opened 2018-10-24
[BUG] Click on task changes the text of the checkbox


kyan001:
It's on macOS, version 0.11.10.

I have a task like this:
```markdown
- [ ] `- [ ] Todo`  # Before clicks
```
If you click on that check box, it will be like this:
```markdown
- [x] `- [ ] Todo`  # After the 1st click
```
As far as now, it's ok, but when you wanna cancel it by clicking it again, it will be like this:
```markdown
- [x] `- [x] Todo`  # After the 2nd click
```
instead of ``- [ ] `- [ ] Todo` ``. And if I clicked it again, it will be like this:
```markdown
- [ ] `- [x] Todo`  # After the 3rd click
```
Screenshots:
![image](https://user-images.githubusercontent.com/3616478/47411078-38a3e280-d79a-11e8-91fe-6adaa5144640.png)
![image](https://user-images.githubusercontent.com/3616478/47411093-422d4a80-d79a-11e8-9070-fad38e96f124.png)
![image](https://user-images.githubusercontent.com/3616478/47411100-48bbc200-d79a-11e8-8c26-2820c493a1e0.png)
![image](https://user-images.githubusercontent.com/3616478/47411150-7143bc00-d79a-11e8-888b-622b33b01486.png)

IssuehuntBot:
@boostio funded this issue with $20. [See it on IssueHunt](https://issuehunt.io/repos/53266139/issues/2530)

kyan001:
> @kyan001 have you finished this PR? I can still reproduce the bug.

I thought this place is a place to report bugs... Should I have to fix it by myself? If so, I'd better to start learning electron now.

ZeroX-DG:
@kyan001 Oops, wrong page :smile: sorry.

IssuehuntBot:
@kazup01 has rewarded. [See it on IssueHunt](https://issuehunt.io/repos/53266139/issues/2530)

PR #2569, opened 2018-11-03
fix for #2530

masaodev:


ZeroX-DG:
@masaodev Have you finished the PR? I can still reproduce the old bug.

ZeroX-DG:
Fixed in https://github.com/BoostIO/Boostnote/pull/2589

PR #2583, opened 2018-11-08
fix GUI style element for default theme

arcturus140:
fix white text not visible on white background for default theme and white theme.

bn_fixguistyle


arcturus140:
oh boy, I did not meant to delete this branch :roll_eyes:

Issue #2588, opened 2018-11-08
Confirmation is needed for the Empty Trash operation.


kanonji:
# Current behavior

No confirmation for the Empty Trash operation.

![boostnote 2018-11-09 02 28 28](https://user-images.githubusercontent.com/134377/48216314-b6393680-e3c7-11e8-9b9b-98ec568ce6f2.png)

# Expected behavior

The Empty Trash operation affects much more notes than clicking trash icon on  top-right corner that have a confirmation.
I think it is better to have confirmation for the Empty Trash operation too.

![boostnote 2018-11-09 02 26 15](https://user-images.githubusercontent.com/134377/48216313-b6393680-e3c7-11e8-8f47-de2c60bcfd69.png)

# Steps to reproduce

1. Right-click on Trash
2. Click on Empty Trash

# Environment

- Version : 0.11.10
- OS Version and name : Windows 10

ZeroX-DG:
Thank you for reporting. We'll look into this.

PR #2590, opened 2018-11-09
enhancement for #2588

arcturus140:
# enhancement for #2588 

empty trash requires user confirmation. This is mandatory as it is a destructive operation.


arcturus140:
as a side note, there could be 2 different messages for the non-destructive 'move to trash' operation and the destructive 'empty trash'. Currently, both messages display: 
> This will permanently remove this note.

PR #2603, opened 2018-11-12
Update electron framework to version 3.0.8

jhit:
On Ubuntu 18.10 there is a issue with glibc versions below 2.28 see
[Prebuilt Electron binaries segfault at startup on Arch Linux with glibc 2.28 #13972](https://github.com/electron/electron/issues/13972)

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "dev": "node dev-scripts/dev.js"
   },
   "config": {
-    "electron-version": "3.0.3"
+    "electron-version": "3.0.8"
   },
   "repository": {
     "type": "git",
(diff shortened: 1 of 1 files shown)


daiyam:
@jhit the glibc issue is already fixed by #2484 but an update to the Electron 3.0.8 is welcome 😄

jhit:
@daiyam Currently you are using electron 3.0.3, the "Fix" you linked to was for the 2.0.x branch of electron. As soon as you switched to 3.0.3 you reverted this "Fix". 3.0.3 has the glibc issue too. I do not know which version of the 3.0.x branch fixed the glibc issue but 3.0.8 is working here on my Ubuntu 18.10.

Redsandro:
Any version of Boostnote after this update crashes AMD Picasso computers. 0.11.10 is the last working version.

https://github.com/BoostIO/Boostnote/issues/3122

Issue #2628, opened 2018-11-20
Paste process error, added '\' and space


weaming:
# Current behavior

![image](https://user-images.githubusercontent.com/10275711/48759609-9a466680-ecde-11e8-8739-4f3f67cbbbf6.png)

# Expected behavior

* No `\` before `_`
* No more space after the end of line

# Steps to reproduce

Copy and paste below into it

```
import time
import pytz
import arrow
import datetime
from django.utils import timezone as djtz
from django.utils.dateparse import parse_datetime

# ISO 8601
# https://www.w3.org/TR/NOTE-datetime
# http://crsmithdev.com/arrow/#tokens
TIME_FORMAT = 'YYYY-MM-DDTHH:mm:ss.SSSZZ'


def get_current_timestamp():
    """
    :return: current timestamp in ms (13 bit)
    """
    return int(time.time() * 1000)


def get_timezone_now():
    """
    need set USE_TZ = True in settings
    :return: datetime now determined by timezone in django
    """
    return djtz.now()


def get_timezone_after(now, **kwargs):
    """
    :param now: datetime by timezone in django
    :param kwargs: timedelta kwargs
    :return: datetime by timezone in django
    """
    if isinstance(now, str):
        now = parse_datetime(now)
    return now + djtz.timedelta(**kwargs)


def convert_timedelta(duration):
    days, seconds = duration.days, duration.seconds
    hours = days * 24 + seconds // 3600
    minutes = (seconds % 3600) // 60
    seconds = (seconds % 60)
    return hours, minutes, seconds


def convert_timedelta_to_minutes(duration):
    return round(duration.total_seconds()/60)


def format_time(o, fmt=TIME_FORMAT, timezone='America/New_York'):
    return arrow.get(o).to(timezone).format(fmt)


def make_aware_utc(d):
    return djtz.make_aware(d, timezone=pytz.utc)


def datetime_to_date_utc(dt):
    return datetime.datetime(dt.year, dt.month, dt.day, tzinfo=pytz.utc)


def iter_dates_between(start_date, end_date):
    d = start_date
    delta = datetime.timedelta(days=1)
    while d <= end_date:
        yield d
        d += delta


def get_aware_datetime(date_str):
    ret = parse_datetime(date_str)
    if not djtz.is_aware(ret):
        ret = djtz.make_aware(ret)
    return ret
```

# Environment

- Version : 0.11.11
- OS Version and name : OSX 10.13.3

daiyam:
[...]

PR #2641, opened 2018-11-25
fix pasting into fenced code block

daiyam:
## Description

This change fixes the pasting in code block.
It avoid the pasted code to be parsed as HTML and so avoid any `\` or new lines to be added to the code.

## Issue fixed

- #2628

## Type of changes

## Checklist:

### DIFF
--- a/browser/components/CodeEditor.js
+++ b/browser/components/CodeEditor.js
@@ -615,7 +615,6 @@ export default class CodeEditor extends React.Component {
           } else {
             return true
           }
-          return true
         } else {
           return false
         }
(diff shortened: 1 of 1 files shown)


daiyam:
I've added the option to enable/disable to paste as HTML
screenshot

and the shortcut to do it even if it's disabled
screenshot

But the major change is that instead of using the browser's clipboard, it uses the electron's clipboard (so the code can be shared between the new shortcut and `Ctrl+V`).
So, when pasting smartly, it can also recognize images in Data URI.

ZeroX-DG:
@daiyam can you resolve the conflict?

richardtks:
@daiyam what is feature of smart paste? Just curious.

daiyam:
@richardtks if the copied data is some html code (from navigator or styled terminal), it convert the it as markdown code (h1 becomes an heading). But it won't convert if it's pasted in a code block.

Issue #2534, opened 2018-10-25
Scheduled themes or auto day/night mode


lebeshev:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%2430%20Rewarded-%237E24E3.svg" />](https://issuehunt.io/r/BoostIo/Boostnote/issues/2534)

Would be cool if users were able to make some kind of a theme schedule, so they could use light theme during the day, but dark theme during the night, by manually setting the time when themes get changed (kinda like Vivaldi's theme schedule featue).
Did't find this in settings or an issue with this feature request.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $30.00)

- [<img src='https://avatars3.githubusercontent.com/u/13612118?v=4' alt='boostio' width=24 height=24> boostio](https://issuehunt.io/u/boostio) ($30.00)

### Submitted pull Requests
- [#2658 Fix issue2534](https://issuehunt.io/r/BoostIO/Boostnote/pull/2658)
---

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/BoostIo/Boostnote/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
---
IssueHunt has been backed by the following sponsors. [Become a sponsor](https://issuehunt.io/membership/members)
</details>

ZeroX-DG:
This is definitely a great feature to have, thank you for suggesting.

IssuehuntBot:
@boostio funded this issue with $30. [See it on IssueHunt](https://issuehunt.io/repos/53266139/issues/2534)

gregueiras:
Hi, me and @susanalima would like to work on this issue. Anything that we should know in advance?

Thank you 😊

ZeroX-DG:
Hey @gregueiras and @susanalima, there's a bounty on this issue so if you want to work on this issue and earn money from it. Then you can sign up with IssueHunt: https://issuehunt.io/ then submit your link to the pull request at this [link](https://issuehunt.io/repos/53266139/issues/2534). After your PR is merged you will be rewarded. Thank you for contributing to Boostnote

IssuehuntBot:
@gregueiras has submitted a pull request. [See it on IssueHunt](https://issuehunt.io/repos/53266139/issues/2534)

Flexo013:
[...]

PR #2658, opened 2018-11-29
Fix issue2534

gregueiras:
## Description
Users can now schedule a theme to be applied
image
In this example, from 07:20 to 17:50 the theme to be used is _Monokai_, otherwise the Default theme is used

## Issue fixed

Issue #2534 Scheduled themes or auto day/night mode

## Type of changes

## Checklist:

Note: I couldn't run the tests


gregueiras:
@ZeroX-DG Thank you 😄

ZeroX-DG:
@gregueiras can you resolve all the conflicts?

gregueiras:
> @gregueiras can you resolve all the conflicts?

I've resolved all the conflicts and changed the scheduled theme dropdown options to be the same as the default theme one.

browser/main/modals/PreferencesModal/UiTab.js
```diff
@@ -207,6 +215,43 @@ class UiTab extends React.Component {
...
+    const sliderID = 'rs-range-line-' + number
+    const bulletID = 'rs-bullet-' + number
+
+    const rangeSlider = document.getElementById(sliderID)
+    const rangeBullet = document.getElementById(bulletID)
+
+    if (rangeSlider && rangeBullet) {
+      const bulletPosition = rangeSlider.value / rangeSlider.max
+      rangeBullet.style.left =
+        bulletPosition * 574 + 6 * (1 - bulletPosition) + 'px'
```
ZeroX-DG: Can you explain to me where the value `574` come from?

browser/main/modals/PreferencesModal/UiTab.js
```diff
@@ -207,6 +215,43 @@ class UiTab extends React.Component {
     }, 2000)()
   }
 
+  /**
+   * Gets the total number of minutes and returns a string in the HH:MM format
```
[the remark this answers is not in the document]
gregueiras: I've removed that comment

browser/main/modals/PreferencesModal/UiTab.js
gregueiras: We thought that we needed to set the slider position, but I tried removing it and it is working as expected, so I've removed the handleSlider function and we can just use the handleUIChange

gregueiras:
I've fixed the problem. It was just the way I was comparing the dates. If you scheduled the change to happen at 15:30, it would change only at 15:31.

browser/main/lib/ThemeManager.js
```diff
@@ -10,8 +10,8 @@ const chooseTheme = ui => {
   const minutes = now.getHours() * 60 + now.getMinutes()
 
   const isEndAfterStart = end > start
-  const isBetweenStartAndEnd = minutes >= start && minutes <= end
-  const isBetweenEndAndStart = minutes >= start || minutes <= end
+  const isBetweenStartAndEnd = minutes >= start && minutes < end
+  const isBetweenEndAndStart = minutes >= start || minutes < end
```
gregueiras: I've fixed the problem. It was just the way I was comparing the dates. If you scheduled the change to happen at 15:30, it would change only at 15:31.

ZeroX-DG:
@gregueiras it's still not working for me. I'm not quite sure what is causing it because at first the `setInterval` doesn't work so I removed `this.refreshTheme=` and it start running but it still doesn't change the theme. I was on setting page the theme didn't change, not sure if it change on main page.

gregueiras:
Hey, sorry about the delay.

I couldn't reproduce your problem... I thought it could be some bug regarding timezone, but even after changing it, it's still working

Peek 2020-03-24 10-20


Do you any other idea?

ZeroX-DG:
@gregueiras The problem that I'm having is the incorrect theme from config. I added this line in your refresh theme loop.
image
And this is what been logging to the terminal
Peek 2020-03-27 11-25
Do you have any idea for what is causing this issue?

gregueiras:
I believe I've fixed it!
Screenshot from 2020-03-27 00-01-01


Previously, I was changing the instance returned from ConfigManager, but wasn't really setting it anywhere. I've changed ThemeManager to return the new configuration if it changes, and then we can set the new config on the loop. I've tought about setting it directly on ThemeManager, but as ConfigManager imports it, it would create a circular dependency. 

What do you think?

Issue #2675, opened 2018-12-03
The output HTML has no codeblock color


blackredscarf:
# Current behavior

My note like this:
![](https://i.loli.net/2018/12/03/5c0542cd22eee.png)

But the codeblock of output HTML has no color:
![](https://i.loli.net/2018/12/03/5c05430b4e6ac.png)

# Expected behavior

the codeblock of output HTML has color and same to the note.

# Steps to reproduce

# Environment

- Version : windows 0.11.11
- OS Version and name :

pmscosta:
Working on it! :+1:

IssuehuntBot:
@issuehuntfest has funded $20.00 to this issue. [See it on IssueHunt](https://issuehunt.io/repos/53266139/issues/2675)

daiyam:
@blackredscarf I've improve my PR #2612 to fix the issue. Can you test it?

IssuehuntBot:
@daiyam has submitted a pull request. [See it on IssueHunt](https://issuehunt.io/repos/53266139/issues/2675)

PR #2686, opened 2018-12-07
Bug - The output HTML has no codeblock color #2675

pmscosta:
## Description
This PR solves the #2675 issue. Upon looking at the code I noticed that the default configuration (sanitize = 'STRICT'), did not allow style tags in the HTML parsing module used. The best solution would be to change the default configuration, present in ConfigManager.js, to sanitize='ALLOW_STYLES'. Meanwhile, in the first run of the program a default configuration is stored in a different directory at your home folder (~./config). For most users, this default configuration already has the 'STRICT' tag. In the startup this saved configuration is assigned with the one present in ConfigManager, so the newly defined tag won't be used. Finally, for this solution to take effect it would require for all previous users to delete the .config folder and run the program. 
The solution I implemented, that requires no deletions of folders, was simply to allow style tags in the STRICT sanitize method (that renders the ALLOW_STYLES sanitize useless).

Output:
screenshot from 2018-12-07 14-22-13

## Issue fixed
The output HTML has no codeblock color #2675

## Type of changes

- :radio_button: Breaking change (Change that can cause existing functionality to change)
- :white_circle: : Improvement (Change that improves the code. Maybe performance or development improvement)

## Checklist:

### DIFF
--- a/browser/lib/markdown.js
+++ b/browser/lib/markdown.js
@@ -38,7 +38,7 @@ class Markdown {
       const allowedTags = ['iframe', 'input', 'b',
         'h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'h7', 'h8', 'br', 'b', 'i', 'strong', 'em', 'a', 'pre', 'code', 'img', 'tt',
         'div', 'ins', 'del', 'sup', 'sub', 'p', 'ol', 'ul', 'table', 'thead', 'tbody', 'tfoot', 'blockquote',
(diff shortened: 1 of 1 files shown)


ZeroX-DG:
I think you get the wrong idea here. I think what the issue creator means was the codeblock displayed in app has a different background than when it's exported to an html file.

pmscosta:
Yes, you are right.
But, from what I've seen both of the cases have some issues. If we can stylize the markdown in the app, it would make sense to see that styling output in the markdown viewer (which is not currently possible). My PR aimed to fix that. The output HTML function also ignores all the style tags when done, but I'll look into that now.

ZeroX-DG:
I believe this issue is fixed in the latest version

Issue #2645, opened 2018-11-25
Request: Customize All Keyboard Shortcuts


incace:
# Current behavior

Currently, it looks like you can only customize three of the Hotkeys/Keyboard Shortcuts. Are there any plans in the future to add the ability to customize all hotkeys? I find focusing the search box with Cmd + Shift + L to be rather annoying. Please tell me I am missing a setting somewhere where I can customize them all.

![screenshot](https://user-images.githubusercontent.com/61343/48983962-f6055b00-f0f5-11e8-95b3-2289f64f39ce.png)

# Expected behavior

Customize all Hotkeys/Shortcuts

# Steps to reproduce

N/A

# Environment

- Version : 0.11.11
- OS Version and name : macOS High Sierra (10.13.6)

ZeroX-DG:
Our shortcut system is pretty messy, currently, you can't customize other hotkeys because they're hardcoded.

incace:
Thanks for taking the time to reply. Is this a possibility in the future?

ZeroX-DG:
Not in the near future, I think. It's pretty messy in the code right now and the project maintainers are very busy and working on a re-write version of this app so it very depends on the community right now.

IssuehuntBot:
@kazup01 funded this issue with $30. [See it on IssueHunt](https://issuehunt.io/repos/53266139/issues/2645)

IssuehuntBot:
@kazup01 has cancelled funding for this issue.(Cancelled amount: $30.00) [See it on IssueHunt](https://issuehunt.io/repos/53266139/issues/2645)

IssuehuntBot:
@issuehuntfest has funded $30.00 to this issue. [See it on IssueHunt](https://issuehunt.io/repos/53266139/issues/2645)

elfman:
I want to work on this issue, what keyboard shortcuts should be customizable? all options in `views`?

jvgrootveld:
@elfman If you are trying to add more customisable shortcuts, could you check if the shortcut still exist for exiting _Editor mode_? The [wiki - Keyboard Shortcuts](https://github.com/BoostIO/Boostnote/wiki/Keyboard-Shortcuts) describes `Ctrl + W` as _ Exit Editor mode_ but on OSX `CMD + W` is bound to _Exit Window_.

daiyam:
To be able to customize CodeMirror shortcuts would be great.
[...]

PR #2751, opened 2018-12-26
More hotkeys can be customized in settings page

elfman:
for #2645

### DIFF
--- a/browser/main/lib/ConfigManager.js
+++ b/browser/main/lib/ConfigManager.js
@@ -26,7 +26,17 @@ export const DEFAULT_CONFIG = {
     toggleMain: OSX ? 'Command + Alt + L' : 'Super + Alt + E',
     toggleMode: OSX ? 'Command + Alt + M' : 'Ctrl + M',
     deleteNote: OSX ? 'Command + Shift + Backspace' : 'Ctrl + Shift + Backspace',
-    pasteSmartly: OSX ? 'Command + Shift + V' : 'Ctrl + Shift + V'
+    pasteSmartly: OSX ? 'Command + Shift + V' : 'Ctrl + Shift + V',
+    reload: OSX ? 'Command + R' : 'Ctrl + R',
+    devTool: OSX ? 'Command + Alt + I' : 'Control + Shift + I',
+    nextNode: OSX ? 'Command + ]' : 'Ctrl + ]',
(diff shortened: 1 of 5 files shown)


browser/main/lib/ConfigManager.js
```diff
@@ -26,7 +26,17 @@ export const DEFAULT_CONFIG = {
     toggleMain: OSX ? 'Command + Alt + L' : 'Super + Alt + E',
     toggleMode: OSX ? 'Command + Alt + M' : 'Ctrl + M',
     deleteNote: OSX ? 'Command + Shift + Backspace' : 'Ctrl + Shift + Backspace',
-    pasteSmartly: OSX ? 'Command + Shift + V' : 'Ctrl + Shift + V'
+    pasteSmartly: OSX ? 'Command + Shift + V' : 'Ctrl + Shift + V',
+    reload: OSX ? 'Command + R' : 'Ctrl + R',
+    devTool: OSX ? 'Command + Alt + I' : 'Control + Shift + I',
```
ZeroX-DG: Should we allow users to customize `devTool`?

elfman: why not?

ZeroX-DG: Well maybe a user will change it (without knowing what it does) because it conflict with the hotkey that they want and next time they can’t open the dev tool using that hotkey anymore. But I think we can help the user in that case so I guess we can leave it :smile:

jvgrootveld: Personally I think shortcuts only need to be available for frequently used features. Maybe an advanced settings part could be added to optionally set shortcuts to advanced features like the `devTool`.

ZeroX-DG: I'm thinking of a scenario where a user wants to use the shortcut that was assigned to `devTool` so the customizable `devTool` shortcut in this case is useful

ZeroX-DG:
@elfman sorry for the wait, can you resolve the conflict please?

ZeroX-DG:
@elfman ping

ZeroX-DG:
@elfman this is the last ping, I'll take over this if you not interested in it anymore :+1:

ZeroX-DG:
Take over here https://github.com/BoostIO/Boostnote/pull/3601. Thank you for this PR :+1:

Issue #2718, opened 2018-12-18
Update button


empeje:
# Current behavior

I'm looking for update button, now I need to download binary manually.

# Expected behavior

Update button not available.

# Steps to reproduce

1. Open Boostnote
2. Look for the update button in the menu bar
3. Nothing there

# Environment

- Version: Mac OSX High Sierra
- OS Version and name: 0.11.11

ZeroX-DG:
Usually, there will be a message box saying `Update available!`.

empeje:
I got the message box prompt when there's an update but I can't trigger it manually.

Rokt33r:
Closed because https://github.com/BoostIO/Boostnote/issues/2827 covers this issue.

PR #2784, opened 2019-01-09
Fix Issue 2718

MiloTodt:
Added "Check for Updates" to the Help menu, which manually calls the built in update function as requested.

Fix > #2718 

image

### DIFF
--- a/browser/components/MarkdownPreview.js
+++ b/browser/components/MarkdownPreview.js
@@ -491,6 +491,7 @@ export default class MarkdownPreview extends React.Component {
     eventEmitter.on('export:save-md', this.saveAsMdHandler)
     eventEmitter.on('export:save-html', this.saveAsHtmlHandler)
     eventEmitter.on('print', this.printHandler)
+    eventEmitter.on('check-updates', this.checkUpdate)
   }
 
   componentWillUnmount () {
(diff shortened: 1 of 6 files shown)


browser/components/MarkdownPreview.js
```diff
@@ -491,6 +491,7 @@ export default class MarkdownPreview extends React.Component {
     eventEmitter.on('export:save-md', this.saveAsMdHandler)
     eventEmitter.on('export:save-html', this.saveAsHtmlHandler)
     eventEmitter.on('print', this.printHandler)
+    eventEmitter.on('check-updates', this.checkUpdate)
```
ZeroX-DG: I don't see the `checkUpdate` method in this file, is this a mistake? If it not can you explain this to me?

MiloTodt:
I suppose it may help if I actually submit the code, eh?

browser/components/MarkdownPreview.js
MiloTodt: Actually pushing the written code would have been a great first step on my end.

MiloTodt:
##Diagram of event messages sent. 

The reason it goes menu->index->app->index is that I was unable to find a way to ssend from menu->app directly.

image

ZeroX-DG:
I think it looks OK but I can't really test it because I'm on Linux. I'll need @Rokt33r to review this.

Rokt33r:
I made a new spec, https://github.com/BoostIO/Boostnote/issues/2827, to resolve the issue. So I think I should close this pr. 😭 

Thanks for your contribution!

PR #2798, opened 2019-01-13
Disable updates in dev

MiloTodt:
App will no longer attempt to update whilst in development mode.
Follows guidelines from https://electronjs.org/docs/tutorial/updates

### DIFF
--- a/lib/main-app.js
+++ b/lib/main-app.js
@@ -2,7 +2,8 @@ const electron = require('electron')
 const app = electron.app
 const Menu = electron.Menu
 const ipc = electron.ipcMain
-const GhReleases = require('electron-gh-releases')
+const GhReleases = require('electron-gh-releases') 
+const isDev = require('electron-is-dev') //isDev is True if running in development mode. Overide by setting ELECTRON_IS_DEV environment variable to 1
 // electron.crashReporter.start()
 var ipcServer = null
 
@@ -35,6 +36,10 @@ const updater = new GhReleases(ghReleasesOpts)
 // Check for updates
 // `status` returns true if there is a new update available
 function checkUpdate () {
+  if (isDev){ //Prevents app from attempting to update when in dev mode.
+    console.log("Updates are disabled in Development mode, see main-app.js")
+    return true
+  }
   if (process.platform === 'linux' || isUpdateReady) {
     return true
   }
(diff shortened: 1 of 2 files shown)


lib/main-app.js
```diff
@@ -3,6 +3,7 @@ const app = electron.app
 const Menu = electron.Menu
 const ipc = electron.ipcMain
 const GhReleases = require('electron-gh-releases')
+const isDev = require('electron-is-dev') // isDev is True if running in development mode. Overide by setting ELECTRON_IS_DEV environment variable to 1
```
ZeroX-DG: Do we really need a package just for checking the app is in development? Can we just use the old fashion way like this?
```
const isDev = process.env.NODE_ENV !== 'production'
```

lib/main-app.js
```diff
@@ -94,12 +99,12 @@ app.on('ready', function () {
 
   // Check update every day
   setInterval(function () {
-    checkUpdate()
+    if (!isDev) checkUpdate()
```
ZeroX-DG: Why do we have to re-check if the app is in development if we already check it in the `checkUpdate` function?

lib/main-app.js
MiloTodt: The merits between those two choices I'm not sure of, this is just the recommended method from the official Electron guides.

MiloTodt: electron-is-dev is pretty lightweight 
image

lib/main-app.js
MiloTodt: This was a consideration I had but felt there may be some value in making it explicit for functions that call checkUpdate that it does not run in dev mode. Computationally it’s a lightweight conditional check but the concern was for future dodgers to have more clarity why the update check is not running in development mode. Feel free to remove.

lib/main-app.js
ZeroX-DG: I think the old way is enough for this task because `electron-is-dev` does almost the same thing:
image

lib/main-app.js
ZeroX-DG: Well, I think you're right. Let's just leave it then.

lib/main-app.js
```diff
@@ -3,6 +3,7 @@ const app = electron.app
 const Menu = electron.Menu
 const ipc = electron.ipcMain
 const GhReleases = require('electron-gh-releases')
+const isDev = 'ELECTRON_IS_DEV' in process.env
```
ZeroX-DG: I'm still a little bit concern about this. Can you change to this just for safe:
```
'ELECTRON_IS_DEV' in process.env && process.env.NODE_ENV !== 'production'
```

lib/main-app.js
```diff
@@ -3,6 +3,7 @@ const app = electron.app
 const Menu = electron.Menu
 const ipc = electron.ipcMain
 const GhReleases = require('electron-gh-releases')
+const isDev = 'ELECTRON_IS_DEV' in process.env && process.env.NODE_ENV !== 'production'
```
Rokt33r: I don't think we need to check `'ELECTRON_IS_DEV' in process.env`. Just checking `process.env.NODE_ENV !== 'production'` is definitely enough. 😄

MiloTodt: Changed as requested

Issue #1448, opened 2018-01-22
Open image like url


ww7:
Sometimes need to zoom image, request to make images/attachments clickable, with saved drag'n'drop ability

ItsPugle:
What about having something like [the Medium Image Zoom feature](https://medium.design/image-zoom-on-medium-24d146fc0c20)?

IssuehuntBot:
@kazup01 has rewarded $13.50 to @roottool. [See it on IssueHunt](https://issuehunt.io/repos/53266139/issues/1448)

- :moneybag: Total deposit: $15.00
- :tada: Repository reward(0%): $0.00
- :wrench: Service fee(10%): $1.50

PR #2809, opened 2019-01-16
Zoom in and out image #1448

roottool:
## Description

I added a new feature like [the Medium Image Zoom feature](https://medium.design/image-zoom-on-medium-24d146fc0c20).
zoominoutimage

## Issue fixed

#1448 

## Type of changes

## Checklist:

### Note
I was able to test all existing tests by Windows subsystem for linux(Ubuntu 18.04).

### DIFF
--- a/browser/components/MarkdownPreview.js
+++ b/browser/components/MarkdownPreview.js
@@ -845,8 +845,8 @@ export default class MarkdownPreview extends React.Component {
 
           const zoomImgWidth = img.width * magnification
           const zoomImgHeight = img.height * magnification
-          const zoomImgTop = document.body.clientHeight / 2 - zoomImgHeight / 2
-          const zoomImgLeft = document.body.clientWidth / 2 - zoomImgWidth / 2
+        const zoomImgTop = (document.body.clientHeight - zoomImgHeight) / 2
+        const zoomImgLeft = (document.body.clientWidth - zoomImgWidth) / 2
           const originalImgTop = img.y + rect.top
           const originalImgLeft = img.x + rect.left
           const originalImgRect = {
--- a/browser/components/MarkdownPreview.js
+++ b/browser/components/MarkdownPreview.js
@@ -838,30 +838,39 @@ export default class MarkdownPreview extends React.Component {
       const imgList = item.contentWindow.document.body.querySelectorAll('img')
       for (const img of imgList) {
         img.onclick = () => {
-          const magnification = document.body.clientWidth / img.width
-          const top = document.body.clientHeight / 2 - img.height * magnification / 2
-          const animationSpeed = 300
+          const widthMagnification = document.body.clientWidth / img.width
+          const heightMagnification = document.body.clientHeight / img.height
+          const baseOnWidth = widthMagnification < heightMagnification
+          const magnification = baseOnWidth ? widthMagnification : heightMagnification
+
+          const zoomImgWidth = img.width * magnification
+          const zoomImgHeight = img.height * magnification
+          const zoomInImgTop = document.body.clientHeight / 2 - zoomImgHeight / 2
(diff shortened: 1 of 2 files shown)


ZeroX-DG:
Nice feature!

MiloTodt:
Great feature!

browser/components/MarkdownPreview.js
```diff
@@ -832,6 +832,82 @@ export default class MarkdownPreview extends React.Component {
         )
       }
     )
+
+    document.querySelectorAll('iframe').forEach(item => {
```
ZeroX-DG: Why do you need to select all iframes like this? The markdown preview only contains 1 iframe, then why did you have to find all iframes and loop through them? Can you explain this to me?

browser/components/MarkdownPreview.js
```diff
@@ -832,6 +832,82 @@ export default class MarkdownPreview extends React.Component {
...
+      for (const img of imgList) {
+        img.onclick = () => {
+          const widthMagnification = document.body.clientWidth / img.width
+          const heightMagnification = document.body.clientHeight / img.height
+          const baseOnWidth = widthMagnification < heightMagnification
+          const magnification = baseOnWidth ? widthMagnification : heightMagnification
+
+          const zoomImgWidth = img.width * magnification
+          const zoomImgHeight = img.height * magnification
+          const zoomImgTop = document.body.clientHeight / 2 - zoomImgHeight / 2
```
ZeroX-DG: Little suggestion, you can transform this code into:
```js
const zoomImgTop = (document.body.clientHeight - zoomImgHeight) / 2
```
This way, we only need to do 2 calculation instead of 3

roottool: I didn't notice it.
I am going to fix it tomorrow. 
Thank you for teaching it!

browser/components/MarkdownPreview.js
roottool: # Explanation
The code is my mistake.
Because when I tried to catch image elements into iframe, I used [the code](https://stackoverflow.com/a/42907951) as it is without thinking.

# Suggestion
I have a suggestion to resolve it.
suggestion
`document.querySelector()` searches for an element from entire HTML.
I considered the possibility of adding new iframes in Boostnote.
So I use a classname of a markdown preview iframe.

It has a pros and cons.
**Pros**
- Specifying an element is easy because `MarkdownPreview` used only here.

**Cons**
- If a classname changes from `MarkdownPreview`, a classname using `document.querySelector()` will need to change too.

## Before
```js
    document.querySelectorAll('iframe').forEach(item => {
      const rect = item.getBoundingClientRect()
      const imgList = item.contentWindow.document.body.querySelectorAll('img')
```

## After
```js
    const markdownPreviewIframe = document.querySelector('.MarkdownPreview')
    const rect = markdownPreviewIframe.getBoundingClientRect()
    const imgList = markdownPreviewIframe.contentWindow.document.body.querySelectorAll('img')
```

ZeroX-DG: We can just go for the 2nd method, I don't think there's any reason to change the `MarkdownPreview` class name. In case we were going to change it, a simple search & replace for the `MarkdownPreview` class wouldn't be hard.

browser/components/MarkdownPreview.js
ZeroX-DG: No problem :+1:

browser/components/MarkdownPreview.js
roottool: I was worried about changing only in HTML.
I want to modify from before code to after code because I feel relieved by your reply.

ZeroX-DG: Yes, you can just go with the after code :+1:

roottool: OK, I rewrote to after code.
Please check the code again.

PR #2820, opened 2019-01-21
fix #2816 Application window bigger after every restart

### LINKED ISSUE #2816: Application window bigger after every restart
It's a little bit annoying, everytime I start Boostnote the window is bigger than last time. The position doesnt change, but width (a little bit) and especially height.
I never use Boostnote in full-window-mode, and I would like it, if the application starts in same size I setted.
I found the closed issue["Keep resized window size"](https://github.com/BoostIO/Boostnote/issues/461), but I think the problem remains.

Win10 Pro 1809, 1920 x 1200 Resolution, Boostnote 0.11.12

MiloTodt:
Removed option in new BrowserWindow:

useContentSize Boolean (optional) - The width and height would be used as web page's size, which means the actual window's size will include window frame's size and be slightly larger. Default is false.

Which was set to true.

Boostnote now opens to the same size as previous isntead of slightly larger each time. 

## Issue fixed
fix #2816 

## Type of changes
 
- ✔️ Bug fix (Change that fixed an issue)

## Checklist:

- ✔️ My code follows [the project code style](docs/code_style.md)
- ✔️ All existing tests have been passed

### DIFF
--- a/lib/main-window.js
+++ b/lib/main-window.js
@@ -18,7 +18,6 @@ const mainWindow = new BrowserWindow({
   y: windowSize.y,
   width: windowSize.width,
   height: windowSize.height,
-  useContentSize: true,
   minWidth: 500,
   minHeight: 320,
   webPreferences: {


Rokt33r:
I've tried this but the window size is still different after I reopen the app again

alehmann76:
I tried it on a Windows 7 machine and it runs like expected.

ZeroX-DG:
@Flexo013 can you test this on window see if the bug is still there? if not then we can close this PR.

PR #2897, opened 2019-02-22
Replace invalid regex for storage placeholder the export functions

dredav:
This pull request will fix the issue #2884. The match to `("|])` doesn't make sense for me. The or match `]` should match `)`. I also removed the group match at the replacement with `DESTINATION_FOLDER`. The group is not needed.

## Issue fixed
Fix wrong test
#2884

## Type of changes

- :large_blue_circle: Bug fix (Change that fixed an issue)
- :white_circle: Breaking change (Change that can cause existing functionality to change)
- :white_circle: Improvement (Change that improves the code. Maybe performance or development improvement)
- :white_circle: Feature (Change that adds new functionality)
- :white_circle: Documentation change (Change that modifies documentation. Maybe typo fixes)

## Checklist:

- :large_blue_circle: My code follows [the project code style](docs/code_style.md)
- :white_circle: I have written test for my code and it has been tested
- :large_blue_circle: All existing tests have been passed
- :white_circle: I have attached a screenshot/video to visualize my change if possible

PR #2901, opened 2019-02-24
Export attachments as html should be okay again --> fixes #2815

### LINKED ISSUE #2815: Exported image source path is incorrect
# Current behavior

Images source path incorrectly placed by Boostnote. It inserts the '%5C' instead of the character '/'.
Therefore source path looks like:
`src="attachments%5C9fad04b1.PNG"`
when it should be:
`src="attachments/fad04b1.PNG"`

Result is that the HTML has broken images links:
untitled

Replacing the wrong characters in the HTML file solves the problem.

# Steps to reproduce

1. Insert any image with Drag&Drop
2. Export in HTML

# Environment

- Version : 0.11.12
- OS Version and name : Windows 7

ehhc:


ZeroX-DG:
This PR will be moved to https://github.com/BoostIO/Boostnote/pull/2897

PR #2910, opened 2019-03-02
Sorting notes alphabetically now sort starting numbers in titles right

### LINKED ISSUE #2894: Sort alphabetically not considering numbers higher than 9
# Current behavior

The sort is set to "Alphabetically".

When I create notes starting with sections numbers, e.g. 1.0 / 2.0 / 2.1 / ... they are all sorted one after another. Now, however, when I create let us say a note starting with 17.0, then it's sorted before 2.0 instead after. So i guess the sorting machine only considers the first character.

# Expected behavior
If sort "Alphabetically" is set and the notes start with numbers, consider the whole first number that is seperated with a comma, dot or any other kind of non-numeric character.

dredav:
## Description
Sorting titles with numbers tries to match float values of left and right titles at comparison. When both numbers are floats and not equal we can do a direct compare. In all other cases the full title will compare.

## Screenshot before
screen shot 2019-03-02 at 14 22 26

## Screenshot after
screen shot 2019-03-02 at 14 22 58

## Issue fixed
#2894

## Type of changes

- :large_blue_circle: Bug fix (Change that fixed an issue)

## Checklist:

- :large_blue_circle: My code follows [the project code style](docs/code_style.md)
- :large_blue_circle: All existing tests have been passed
- :large_blue_circle: I have attached a screenshot/video to visualize my change if possible

### DIFF
--- a/browser/main/NoteList/index.js
+++ b/browser/main/NoteList/index.js
@@ -41,8 +41,9 @@ function sortByAlphabetical (a, b) {
     const floatA = parseFloat(matchA[1])
     const floatB = parseFloat(matchB[1])
 
-    if (Math.abs(floatA - floatB) > 0.01) {
-      return floatA - floatB
+    const diff = floatA - floatB
+    if (diff !== 0) {
+      return diff
     }
 
     // The float values are equal. We will compare the full title.
--- a/browser/main/NoteList/index.js
+++ b/browser/main/NoteList/index.js
@@ -26,15 +26,15 @@ const { remote } = require('electron')
 const { dialog } = remote
 const WP_POST_PATH = '/wp/v2/posts'
 
-const matchStartingTitleNumber = new RegExp('^([0-9]*\.?[0-9]+).*$')
+const regexMatchStartingTitleNumber = new RegExp('^([0-9]*\.?[0-9]+).*$')
 
 function sortByCreatedAt (a, b) {
   return new Date(b.createdAt) - new Date(a.createdAt)
 }
 
 function sortByAlphabetical (a, b) {
-  const matchA = matchStartingTitleNumber.exec(a.title)
-  const matchB = matchStartingTitleNumber.exec(b.title)
+  const matchA = regexMatchStartingTitleNumber.exec(a.title)
+  const matchB = regexMatchStartingTitleNumber.exec(b.title)
 
   if (matchA && matchA.length === 2 && matchB && matchB.length === 2) {
     // Both note titles are starting with a float. We will compare it now.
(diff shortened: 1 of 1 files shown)


browser/main/NoteList/index.js
```diff
@@ -26,11 +26,30 @@ const { remote } = require('electron')
...
 function sortByAlphabetical (a, b) {
+  const matchA = regexMatchStartingTitleNumber.exec(a.title)
+  const matchB = regexMatchStartingTitleNumber.exec(b.title)
+
+  if (matchA && matchA.length === 2 && matchB && matchB.length === 2) {
+    // Both note titles are starting with a float. We will compare it now.
+    const floatA = parseFloat(matchA[1])
+    const floatB = parseFloat(matchB[1])
+
+    if (floatA < floatB) {
```
Rokt33r: Rather than using if statement, `return floatA - floatB` should be enough.

Rokt33r:
@dredav I've submitted a change request. I'll merge this as soon as you resolve it.

dredav:
> @dredav I've submitted a change request. I'll merge this as soon as you resolve it.

Thanks for your review @Rokt33r. I've removed the `if` / `else if ` to return`-1` or `1`, but I must add a Math.Abs call to check if the floats are equal or not. For the equal case we need to compare the full note title in case for title example `1.0 Note 2` and `1.0 Note 1`.

dredav:
@Rokt33r What do you think?

> > @dredav I've submitted a change request. I'll merge this as soon as you resolve it.
> 
> Thanks for your review @Rokt33r. I've removed the `if` / `else if ` to return`-1` or `1`, but I must add a Math.Abs call to check if the floats are equal or not. For the equal case we need to compare the full note title in case for title example `1.0 Note 2` and `1.0 Note 1`.

browser/main/NoteList/index.js
```diff
@@ -26,11 +26,28 @@ const { remote } = require('electron')
...
 function sortByAlphabetical (a, b) {
+  const matchA = regexMatchStartingTitleNumber.exec(a.title)
+  const matchB = regexMatchStartingTitleNumber.exec(b.title)
+
+  if (matchA && matchA.length === 2 && matchB && matchB.length === 2) {
+    // Both note titles are starting with a float. We will compare it now.
+    const floatA = parseFloat(matchA[1])
+    const floatB = parseFloat(matchB[1])
+
+    if (Math.abs(floatA - floatB) > 0.01) {
```
Rokt33r:
```js
const diff = floatA - floatB
if (diff !== 0) {
  return diff
}
```

dredav: Thanks for for suggestion. I'll implement it.

PR #2952, opened 2019-03-27
Issue 2950 - clicking on elements when in full page preview toggles the mode back to split.

### LINKED ISSUE #2950: Switching editor takes priority over (un)fold button
# Current behavior

On a markdown note clicking the fold or unfold arrow causes a switch to editing mode rather than just folding or unfolding.
image

# Expected behavior

The editor mode should not be switched to when clicking the (un)fold arrow.

Small side note: Performing a long click on the arrow does perform the expected behavior.

# Steps to reproduce

1. Have a markdown note, with a fold HTML block.
```md
<details><summary>Test summary.</summary>
Test body
</details>
```
2. Make sure to use the toggle mode view (not the side by side view).
3. Click the (un)fold arrow.
4. It switches to the editor mode.

# Environment

- Version : 0.11.15
- OS Version and name : Windows 10 and Ubuntu 16.04

MiloTodt:
## Description

Refactoring of the handleMouseDown event such that the full page preview pane is not toggled off whenever any element -that is not the `<body>`  of the page- is clicked. Anything like a Button, Video, link etc should no longer toggle the view back to split.

While I was at it, I cleaned up this function greatly, which should help the efforts of future developers. 

## Issue fixed
fix #2950 

## Type of changes

- ✅  Bug fix (Change that fixed an issue)
- ✅ Improvement (Change that improves the code. Maybe performance or development improvement)

## Checklist:

- ✅ My code follows [the project code style](docs/code_style.md)
- ✅ All existing tests have been passed
- ✅ I have attached a screenshot/video to visualize my change if possible


MiloTodt:
I'm not really sure what the purpose of this is:
```js
    if (e.target != null) {
      switch (e.target.tagName) {
        case 'A':
        case 'INPUT':
          return null
      }
    }
```

but I've left it included in refactored to a more readable form that is logically the same as 
```js
    if (clickElement === null && targetTag !== 'INPUT') {
      return null
    }
```

I'm not sure for in what case the clicked element would be null, but still have a tag at all.

MiloTodt:
That code was added here: 

https://github.com/boostio/Boostnote/commit/9cd6d6d4c14d30a4a346b0fd56f031b6337f13c7

But these two statements aren't the same. as case 'A' would no longer return null.

MiloTodt:
I noticed that this file has quite a bit of CSS at the top. Is there reasons for it to be directly inside this file as opposed to a .styl file as all other files have in this folder?

PR #2846, opened 2019-01-29
Introduce markdownlint #864

roottool:
## Description

I introduced [markdownlint](https://github.com/DavidAnson/markdownlint) in Boostnote.
markdownlint

### Note
This feature checks [all rules and aliases](https://github.com/DavidAnson/markdownlint#rules--aliases).
I didn't implement ignore settings on purpose.
Because I think that it will exceed the range dealt with in #864.

## Issue fixed

#864 

## Type of changes

- :white_circle: Bug fix (Change that fixed an issue)
- :white_circle: Breaking change (Change that can cause existing functionality to change)
- :white_circle: Improvement (Change that improves the code. Maybe performance or development improvement)
- :radio_button: Feature (Change that adds new functionality)
- :white_circle: Documentation change (Change that modifies documentation. Maybe typo fixes)

## Checklist:

- :radio_button: My code follows [the project code style](docs/code_style.md)
- :white_circle: I have written test for my code and it has been tested
- :radio_button: All existing tests have been passed
- :radio_button: I have attached a screenshot/video to visualize my change if possible

PR #2983, opened 2019-04-29
Add markdownlint rules form

roottool:
## Caution

This PR relates to #2846.
Please merge this PR after you merge #2846.

## Description

This PR adds a config editor for markdownlint in `Preferences > Interface > Editor`.
I referred to [this line](https://github.com/codemirror/CodeMirror/blob/858f5dcc2cf19d42393c360721e5c3a7dc547c43/demo/lint.html#L13) and [this definition](https://github.com/codemirror/CodeMirror/blob/858f5dcc2cf19d42393c360721e5c3a7dc547c43/demo/lint.html#L156-L161) in codemirror demo.

A value in this editor is linted by [jsonlint-mod](https://www.npmjs.com/package/jsonlint-mod). 
Because markdownlint is broken by broken markdownlint rules.
And `../node_modules/codemirror/addon/lint/json-lint.js` depends on `../node_modules/jsonlint-mod/lib/jsonlint.js`.

## Edited

I improved not to need to reload Boostnote when markdownlint rules is changed by this editor.

## Screenshot

AddMarkdownlintRules

## Type of changes

## Checklist:

### DIFF
--- a/browser/components/CodeEditor.js
+++ b/browser/components/CodeEditor.js
@@ -620,9 +620,9 @@ export default class CodeEditor extends React.Component {
 
   setCodeEditorLintConfig () {
     const { mode } = this.props
-    const checkMarkdownNoteIsOpening = mode === 'Boost Flavored Markdown'
+    const checkMarkdownNoteIsOpen = mode === 'Boost Flavored Markdown'
 
-    return checkMarkdownNoteIsOpening ? {
+    return checkMarkdownNoteIsOpen ? {
       'getAnnotations': this.validatorOfMarkdown,
       'async': true
     } : false
--- a/browser/components/CodeEditor.js
+++ b/browser/components/CodeEditor.js
@@ -86,7 +86,7 @@ export default class CodeEditor extends React.Component {
     this.searchHandler = (e, msg) => this.handleSearch(msg)
     this.searchState = null
     this.scrollToLineHandeler = this.scrollToLine.bind(this)
-    this.setCodeEditorLintConfig = this.getCodeEditorLintConfig.bind(this)
+    this.getCodeEditorLintConfig = this.getCodeEditorLintConfig.bind(this)
     this.validatorOfMarkdown = this.validatorOfMarkdown.bind(this)
 
     this.formatTable = () => this.handleFormatTable()
--- a/browser/components/CodeEditor.js
+++ b/browser/components/CodeEditor.js
@@ -651,11 +651,7 @@ export default class CodeEditor extends React.Component {
           let ruleNames = ''
           item.ruleNames.map((ruleName, index) => {
             ruleNames += ruleName
-            if (index === item.ruleNames.length - 1) {
-              ruleNames += ': '
-            } else {
-              ruleNames += '/'
-            }
+            ruleNames += (index === item.ruleNames.length - 1) ? ': ' : '/'
           })
           foundIssues.push({
             from: CodeMirror.Pos(item.lineNumber, 0),
--- a/browser/components/CodeEditor.js
+++ b/browser/components/CodeEditor.js
@@ -26,6 +26,7 @@ import {languageMaps} from '../lib/CMLanguageList'
 import snippetManager from '../lib/SnippetManager'
 import {generateInEditor, tocExistsInEditor} from 'browser/lib/markdown-toc-generator'
 import markdownlint from 'markdownlint'
+import ConfigManager from '../main/lib/ConfigManager'
 
 CodeMirror.modeURL = '../node_modules/codemirror/mode/%N/%N.js'
 
@@ -39,11 +40,14 @@ function translateHotkey (hotkey) {
 }
 
 const validatorOfMarkdown = (text, updateLinting) => {
+  const config = ConfigManager.get()
   const lintOptions = {
     'strings': {
       'content': text
-    }
+    },
+    'config': JSON.parse(config.editor.customMarkdownLintConfig)
   }
+  console.log(config.editor.customMarkdownLintConfig)
 
   return markdownlint(lintOptions, (err, result) => {
     if (!err) {
--- a/browser/components/CodeEditor.js
+++ b/browser/components/CodeEditor.js
@@ -533,8 +533,8 @@ export default class CodeEditor extends React.Component {
     if (prevProps.keyMap !== this.props.keyMap) {
       needRefresh = true
     }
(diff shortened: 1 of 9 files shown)


browser/components/CodeEditor.js
```diff
@@ -644,6 +618,57 @@ export default class CodeEditor extends React.Component {
     }
   }
 
+  setCodeEditorLintConfig () {
+    const { mode } = this.props
+    const checkMarkdownNoteIsOpening = mode === 'Boost Flavored Markdown'
```
AWolf81: Nit: I think renaming the const would be good. As it's not opening - it is already open. Or is it about to open?
```suggestion
    const checkMarkdownNoteIsOpen = mode === 'Boost Flavored Markdown'
```

roottool: Thnak you for review!
It's a false by my wrong English.

browser/components/CodeEditor.js
```diff
@@ -644,6 +618,57 @@ export default class CodeEditor extends React.Component {
...
+    }
+
+    return markdownlint(lintOptions, (err, result) => {
+      if (!err) {
+        const foundIssues = []
+        result.content.map(item => {
+          let ruleNames = ''
+          item.ruleNames.map((ruleName, index) => {
+            ruleNames += ruleName
+            if (index === item.ruleNames.length - 1) {
```
AWolf81: Nit: IMO it would be more readable if you'd do an inline if like:

```suggestion
            ruleNames += (index === item.ruleNames.length - 1) ? ': ' : '/'
```

roottool: I didn't notice it. Thank you for an advice!

browser/main/lib/ConfigManager.js
```diff
@@ -59,7 +59,10 @@ export const DEFAULT_CONFIG = {
     enableFrontMatterTitle: true,
     frontMatterTitleField: 'title',
     spellcheck: false,
-    enableSmartPaste: false
+    enableSmartPaste: false,
+    customMarkdownLintConfig: `{
```
AWolf81: Please add a constant like following to the top of the file:
```js
const DEFAULT_MARKDOWN_LINT_CONFIG = `{
  "default": true
}`
```

And use it here like:
```suggestion
    customMarkdownLintConfig: DEFAULT_MARKDOWN_LINT_CONFIG 
```

The constant is used so you're not messing with the indentation. The indent added a weird looking output like in the following screenshot:
grafik

After the change it will look like:
grafik

AWolf81:
I like the linting & the configuration field. Well done. 👍 

I think it would be good if we'd add some help docs to the linting. No need to add in this PR but maybe we can create an issue for it.

I think it would be great if we could have the [RULES.md](https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md) as cheatsheet link or should we add the link to the [boostnote cheatsheet](https://github.com/TobseF/boostnote-markdown-cheatsheet/blob/master/BOOSTNOTE_MARKDOWN_CHEAT_SHEET.md)?

Also how can the `custom Markdown lint rules` be used? 
Are there docs about the keys that can be added? 
Custom linting rules can't be added here, right? Or is it possible? If it is possible, an example would be great. I haven't found one.

roottool:
> Custom linting rules can't be added here, right?

Yes, we can't add custom linting rules using the configuration field.
We need to create a PR to [markdownlint](https://github.com/markdownlint/markdownlint) referring to [creating_rules.md](https://github.com/markdownlint/markdownlint/blob/master/docs/creating_rules.md).

Thanks.

AWolf81:
Ah OK, thanks, I see. So for now it's just for disabling or enabling exisisting rules. e.g. the following will disable two rules (heading order & no multiple spaces rule) and enable all not excluded rules with default:
```
{
  "default": true,
  "MD012": false,
  "MD001": false
}
```

I think we need a documentaion to descibe what's possible with that config section.

roottool:
## Description
[This commit](https://github.com/BoostIO/Boostnote/pull/2983/commits/c82dbddc748624cc40758e7f8725a5e2dfabd734#diff-4771352a7e4e4f9d82f0be3e914757adR657) fixes that markdownlint result desplay works properly.
This issue was caused by specifying an incorrect value.

- `item.lineNumber` isn't sequence number. Therefore, I had to adjust to sequence number.
- I had to add a logic to fetch the number of characters until tailing the line.


## Issue number
#2996

browser/components/CodeEditor.js
```diff
@@ -644,6 +618,55 @@ export default class CodeEditor extends React.Component {
     }
   }
 
+  setCodeEditorLintConfig () {
```
daiyam: Hi @roottool, this should be named `getCodeEditorLintConfig` since it returns the config.

browser/main/modals/PreferencesModal/UiTab.js
```diff
@@ -637,6 +640,28 @@ class UiTab extends React.Component {
...
           </div>
+          <div styleName='group-section'>
+            <div styleName='group-section-label'>
+              {i18n.__('Custom MarkdownLint Rules')}
+            </div>
+            <div styleName='group-section-control'>
+              <div style={{fontFamily}}>
+                <ReactCodeMirror
+                  width='400px'
+                  height='400px'
```
daiyam: Can you make it smaller?

daiyam:
Can you remove the gutter `CodeMirror-lint-markers` when `markdownlint` is disabled?

AWolf81:
@daiyam I've added the gutter toggling to PR #3034. It's almost working just the linting errors are not properly displayed after re-enabling the linting.
I need to do more debugging to fix that.

browser/main/modals/PreferencesModal/UiTab.js
roottool: I made height of the form smaller from `400px` to `200px`.
capture

daiyam: Thx

Rokt33r:
I'm going to merge https://github.com/BoostIO/Boostnote/pull/3034! Thanks!

Rokt33r:
@roottool Submit pr to here. https://issuehunt.io/r/BoostIo/Boostnote/issues/3017 :)

roottool:
@Rokt33r, Thank you for the suggestion!
But I decline your suggestion.
Because I don't think that I have the right to receive the reward.
So I'm glad I will use the reward for other issues.

Issue #1706, opened 2018-03-20
Option to rename tags


cheoAlejo:
It would be great if we could rename a tag so the change gets reflected to the notes belonging to that tag.

IssuehuntBot:
@rohimchou has funded $10.00 to this issue.

---
- Submit pull request via [IssueHunt](https://issuehunt.io/repos/53266139/issues/1706) to receive this reward.
- Want to contribute? Chip in to this issue via [IssueHunt](https://issuehunt.io/repos/53266139/issues/1706).
- Checkout the [IssueHunt Issue Explorer](https://issuehunt.io/issues) to see more funded issues.
- Need help from developers? [Add your repository](https://issuehunt.io/r/new) on IssueHunt to raise funds.

PR #2989, opened 2019-05-03
Issue 1706: Option to rename tags

khaliqgant:
## Description
Tags at present moment cannot be renamed. This PR reuses the "RenameFolderModal" to allow a user to rename a tag that is selected. 

video

This update also renames the "RenameFolderModal.styl" to "RenameModal.styl" since the RenameTagModal.js uses that style as well and I felt it was clearer to use a more generic name. This is my first PR to the project (be kind! 😬) so I wasn't sure if that is something the maintainers would agree with or not.

## Issue fixed
Fixes https://github.com/BoostIO/Boostnote/issues/1706
## Type of changes

## Checklist:


browser/main/modals/RenameTagModal.js
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  handleInputKeyDown (e) {
+    switch (e.keyCode) {
+      case 13:
+        this.confirm()
+    }
+  }
+
+  handleConfirmButtonClick (e) {
```
ZeroX-DG: Why don't you call `this.confirm` directly instead of making a whole new method for it?

khaliqgant: Thanks for looking it over! Updated.

browser/main/modals/RenameTagModal.js
```diff
@@ -0,0 +1,124 @@
...
+  render () {
+    return (
+      <div styleName='root'
+        tabIndex='-1'
+        onKeyDown={(e) => this.handleKeyDown(e)}
+      >
+        <div styleName='header'>
+          <div styleName='title'>{i18n.__('Rename Tag')}</div>
+        </div>
+        <ModalEscButton handleEscButtonClick={(e) => this.handleCloseButtonClick(e)} />
```
AWolf81: Nit. I think `handleCloseButtonClick` method is not needed here.
You could destructure prop `close` with `const { close } = this.props` and do
```suggestion
        <ModalEscButton handleEscButtonClick={close} />
```

browser/main/modals/RenameTagModal.js
```diff
@@ -0,0 +1,124 @@
...
+
+        <div styleName='control'>
+          <input styleName='control-input'
+            placeholder={i18n.__('Tag Name')}
+            ref='name'
```
AWolf81: I think it would be better to use callback refs as it's likely that this will be deprecated in a future React version. Please use `ref={ref => this.nameInput=ref}` & add `this.nameInput = null` to the constructor & update the references from `this.refs.name` to the new property.

[Docs about string refs leagcy api](https://reactjs.org/docs/refs-and-the-dom.html#legacy-api-string-refs)

browser/main/modals/RenameTagModal.js
```diff
@@ -0,0 +1,124 @@
...
+            value={this.state.name}
+            onChange={(e) => this.handleChange(e)}
```
AWolf81: Nit. Please use `.bind(this)` for `onChange`, `onKeyDown` and `onClick` so you can write e.g. `onChange={this.handleChange}`.
But this requires `this.handleChange = this.handleChange.bind(this)` in the constructor - we could probably also use class properties but I think that's not used in the code base. 
I'm also not sure if it's supported by the Babel setup - I haven't checked.

AWolf81:
Looks great. Just some smaller inline comments & one point would be awesome if you could add. 

If you're selecting a tag and then rename it - the location is wrong and no notes are displayed.

By adding the following code to your `renameTag` method before closing the modal you can redirect to the selected tag to correctly update:
```js
import { hashHistory } from 'react-router' // add to imports

renameTag (tag, updatedTag) {
  Promise
    .all(notes.map(note => dataApi.updateNote(note.storage, note.key, note)))
    .then(() => { /* ... */ })
    .then(() => {
      if (window.location.hash.includes(tag)) {
        hashHistory.replace(`/tags/${updatedTag}`)
      }
      this.props.close()
    })
}
```

browser/main/modals/RenameTagModal.js
khaliqgant: Fixed in e985f1226a342b7a77190ec9c5a6832eae38ab25

browser/main/modals/RenameTagModal.js
khaliqgant: Fixed in b1d62c8bea08e4257fa5ee918c55ba17e197463c

browser/main/modals/RenameTagModal.js
khaliqgant: Fixed in ce5d4acc11f3c436a5e915510bfb5720ae67ac54

khaliqgant:
Thanks for the detailed feedback @AWolf81!

 Great tip with your comment [above](https://github.com/BoostIO/Boostnote/pull/2989#issuecomment-490606222) as it exposed an issue with the topbar tag 
image
not updating when the tag was renamed. I added in an event (not sure if is the optimal way, but the best I could come up with at the time) to keep that portion up to date. Result below:

updated-video

khaliqgant:
Branch has been brought up to date with the latest master

khaliqgant:
Just a friendly ping on this @ZeroX-DG 😀

ZeroX-DG:
Hi @khaliqgant , sorry for the late response.

Can you change the color of the modal so that it will match the theme? For example, here is how it look on monokai theme

image

Also a note can't have duplicated tags but if you do like this, the bug will appear:
1. Add a tag called `tag1` to a note
2. Add a tag called `tag2` to that same note
3. Rename `tag2` to `tag1`
4. The note will now have duplicated tags.

khaliqgant:
Thanks for the feedback @ZeroX-DG.

> Can you change the color of the modal so that it will match the theme?
Updated with https://github.com/BoostIO/Boostnote/pull/2989/commits/9a2dcbc54cd66bcbc529833a5a002cd2ddcb18e4

Duplicated tags issue when renaming updated in https://github.com/BoostIO/Boostnote/pull/2989/commits/4efbe0832233b479342da5acd7400a45e33113bb and added in an error message to assist the user as seen in the video

video

ZeroX-DG:
@khaliqgant This is not quite what I wanted, I think it's better to merge tag into the existing tag if there's any. If the user want to rename tag `computer` to `computers` which already existed, the app should display a message showing the tag `computers` already existed, renaming this will make note with tag `computer` become `computers` and still allow user to rename it if it's what the user want.

ZeroX-DG:
@khaliqgant ping

ZeroX-DG:
@khaliqgant ping 2

ZeroX-DG:
Close due to inactivity

PR #2990, opened 2019-05-03
Upgrade React & React-DOM & React-Router + adding DevTools

AWolf81:
Great app. I really like the app and this is my first contribution to the code base. I think this is ready for a first review but I'll do a review tomorrow (e.g. to remove development console.logs if I've missed any etc.)

## Description
I've updated React, React-DOM & React-Router to the latest version. There was no issue for this open but I think it's good to update & React-carousel complaint during installation that React was not matching the library version.

I've first thought this is easier todo but it was quite some work. The main problem was React-Router as the newest version v5 is totally different compared to v2 - biggest issue was the removed query prop from history. There is a good [post](https://codeburst.io/react-router-v4-unofficial-migration-guide-5a370b8905a) about how to migrate & what's different compared to v2/v3
First I wanted to migrate to v3 but this was also not working out-of the box - so I decided to migrate the routing to the latest version.

So please have a close look at every route. I think I've tested everything but maybe I have missed an edge case.

**Overview**
- Upgraded React, Redux, React-Router & some more package that were required to the latest version
- Added React & Redux-Devtools (see screenshot below) - helped during development
- Changed routing to component based routing as required by React-Router

**Issues in the code**
- Sometimes during fast typing in search field the input will lose focus and move to markdown text area. Need to check what's causing this and if it's introduced by the routing. It's also happening on master so it's not introduced here. It happens during backspace deleting letters.
- There are many warnings about styling issues in the console - we could turn the output off by ignoring them but I think it's better to fix them and keep the setting to logging. I haven't looked into this and I think we could improve this later.
- There are warnings from React about missing props. I think I haven't introduced these but they're probably easy to fix either by removing `isRequired` or by not rendering if an important part is missing.
- Initial rendering is not smooth but it is as in master. I think we can improve this later by delaying display until everything is ready & keep the loading spinner active. So the app loading is looking way nicer.

**Possible improvements (later, no need to do it in this PR)**
- I think the redux part of the app could be refactored (separate reducer files, action creators & mapState & mapDispatchToProps with connected components) - if you like I can create an issue for this. This will improve readability / code quality.
- Code style improvement: I think it would be good to use class properties instead of the binding in constructor. Less code & better readability. Like mentioned in this [post](https://medium.com/@oleg008/arrow-functions-in-react-f782d11460b4).
- Auto-scroll preformance (here & on master): There is a pretty large lag on Windows 10 during scrolling around 0.5 - 1s before scrolling starts. Not sure why & I need to check if I can see something in the profiling.

## Issue fixed

-

## Type of changes

- :radio_button: Breaking change (Change that can cause existing functionality to change)
- :radio_button: Improvement (Change that improves the code. Maybe performance or development improvement)

## Checklist:

- :white_circle: All existing tests have been passed - `yarn jest` passed but `yarn test` failed (see screenshot below) - need to check why it fails  

**Failed `yarn test`**
grafik

**DevTools**
grafik

**Todos**
- [x] Enable & fix auto-scrolling (commented because of the query to search change)


Rokt33r:
@AWolf81 Submit this issue to here. https://issuehunt.io/r/BoostIo/Boostnote/issues/2993 I'm going to pay for this pr gladly.

browser/lib/newNote.js
```diff
@@ -1,7 +1,9 @@
-import { hashHistory } from 'react-router'
+// import { hashHistory } from 'react-router'
```
Rokt33r: Discard comments

browser/main/Detail/MarkdownNoteDetail.js
```diff
@@ -9,7 +9,8 @@ import StarButton from './StarButton'
 import TagSelect from './TagSelect'
 import FolderSelect from './FolderSelect'
 import dataApi from 'browser/main/lib/dataApi'
-import { hashHistory } from 'react-router'
+// import { hashHistory } from 'react-router'
```
Rokt33r: Discard comments

browser/main/Detail/SnippetNoteDetail.js
```diff
@@ -8,7 +8,8 @@ import StarButton from './StarButton'
 import TagSelect from './TagSelect'
 import FolderSelect from './FolderSelect'
 import dataApi from 'browser/main/lib/dataApi'
-import {hashHistory} from 'react-router'
+// import {hashHistory} from 'react-router'
```
Rokt33r: Discard comments

browser/main/Main.js
```diff
@@ -12,8 +12,8 @@ import _ from 'lodash'
 import ConfigManager from 'browser/main/lib/ConfigManager'
 import mobileAnalytics from 'browser/main/lib/AwsMobileAnalyticsConfig'
 import eventEmitter from 'browser/main/lib/eventEmitter'
-import { hashHistory } from 'react-router'
-import store from 'browser/main/store'
+// import { hashHistory } from 'react-router'
```
Rokt33r: Discard comments

browser/main/NewNoteButton/index.js
```diff
@@ -21,6 +21,9 @@ class NewNoteButton extends React.Component {
     this.state = {
     }
 
+    this.handleNewNoteButtonClick = this.handleNewNoteButtonClick.bind(this)
+
+    // why is there a newNoteHandler?!
```
Rokt33r: I think we don't need it because we have `this.handleNewNoteButtonClick`. But we have to fix `componentDidMount` and `componentWillUnmount ` also so they can use `this.handleNewNoteButtonClick` too.

browser/main/store.js
```diff
@@ -465,13 +467,21 @@ function getOrInitItem (target, key) {
   return results
 }
 
+const history = createHistory()
+
+// Query not available in HashHistory with ReactRouter >= v4
```
Rokt33r: I see. But let's discard comments.

lib/main.html
```diff
@@ -132,8 +132,8 @@
   </script>
 
   <script src="../node_modules/@rokt33r/js-sequence-diagrams/dist/sequence-diagram-min.js"></script>
-  <script src="../node_modules/react/dist/react.min.js"></script>
-  <script src="../node_modules/react-dom/dist/react-dom.min.js"></script>
+  <script src="../node_modules/react/umd/react.development.js"></script>
```
Rokt33r: I assume we need to use another html file for production env because I don't want to use development script in the compiled app.

So what I want to ask are:
1. Rename this file to `lib/main.development.html`
2. Create `lib/main.production.html` by copying from `lib/main.development.html` and update the url so we can use production script.
3. Update declaration statements for `const url` in `lib/main-window.js` so electron can choose the html files by environment.

AWolf81: Good catch. I've added this.
I've also updated the link regex or was that not needed?
What do you think should we create an issue to bundle the electron scripts and we don't have to manually import each script? Minifiying could be done by Webpack production build too.

browser/main/store.js
AWolf81: OK, removed.

browser/main/NewNoteButton/index.js
AWolf81: OK, removed the not needed `newNoteHandler` and updated the mentioned locations.

browser/lib/newNote.js
```diff
@@ -28,9 +29,9 @@ export function createMarkdownNote (storage, folder, dispatch, location, params,
         note: note
       })
 
-      hashHistory.push({
+      history.push({
```
AWolf81: Nit: Should I change this to `dispatch(push(..))` so it's consistent to the other locations? Then we could also remove the import of history.

Rokt33r: I think so. It should be better.

browser/main/NoteList/index.js
```diff
@@ -758,14 +758,16 @@ class NoteList extends React.Component {
     return copy(noteLink)
   }
 
+  // Navigate seems unused
```
Rokt33r: It is executed when a global shortcut event is dispatched. See `componentDidMount`.

lib/main.html
Rokt33r: > I've also updated the link regex or was that not needed?

I cannot understand could you point where the regex is in?

> What do you think should we create an issue to bundle the electron scripts and we don't have to manually import each script? Minifiying could be done by Webpack production build too.

I absolutely want to do. But let's do things one by one. It should be done in another pr.

browser/components/MarkdownPreview.js
```diff
@@ -1023,7 +1023,8 @@ export default class MarkdownPreview extends React.Component {
 
     if (!href) return
 
-    const regexNoteInternalLink = /main.html#(.+)/
+    const regexNoteInternalLink = process.env.NODE_ENV === 'production' ? /main.production.html#(.+)/ : /main.development.html#(.+)/
```
AWolf81: This is the regex I've metioned.
After having a second look on this I think it's important to update. I tested the regex [here](https://regex101.com/r/zRLTIR/1).

browser/main/NoteList/index.js
AWolf81: ah, OK, thanks. I missed that.
I'll remove the comment and also the commented line on line 763.

AWolf81:
@Rokt33r I've replaced the direct usage of `history` with dispatch to `connected-react-router`.

I think it would be good to add that we recommend to use `dipatch(push(...))` over `history.push(...)` where ever possible. Maybe we could add this to [code style guide doc](https://github.com/BoostIO/Boostnote/blob/master/docs/code_style.md). If you like I can create a PR for this.

In the style guide I've seen that it's recommended to use class property over class methods. What do you think should we clean our code base to be consitent with this? During my last reviews I commented to have `.bind(this)` as this is used almost everywhere.

lib/main.html
AWolf81: OK, I'll create an issue for the Webpack update. Then we can also discuss the details there.

browser/components/MarkdownPreview.js
Rokt33r: Hmm.. `/main.[A-Za-z0-9]+.html#/` might be better to have rather than two regex.

Rokt33r: Hmm.. /main.[A-Za-z0-9]+.html#/ might be better to have rather than two regex.

Rokt33r:
> In the style guide I've seen that it's recommended to use class property over class methods. What do you think should we clean our code base to be consitent with this? During my last reviews I commented to have .bind(this) as this is used almost everywhere.

> I think we should also add some info to React hooks and how we'd like to use them.

Both of idea sounds great. But let's do it in other prs. And, I'm going to merge this soon.

browser/components/MarkdownPreview.js
AWolf81: Yes, good idea. Maybe also `/main[.\w]*.html#/` will work - also supports `main.html`. I'll update this later today.

AWolf81: https://regex101.com/r/zRLTIR/2

Rokt33r: @AWolf81 Yeah I think your idea is better. Could you fix it too? Then I think we could merge this.

AWolf81: @Rokt33r I don't get it to work. Seems like my previous regex was also not working.
Here is my code inside `MarkdownPreview.js`:
grafik

And I'm getting the following output on local link click:
grafik

Seems like I'm not getting into the if for local file links.

AWolf81: I think it's because there is no `main.development.html` in the `linkHash` string. Maybe new router changed here.
I have to check how to detect the local link.

AWolf81: Maybe I have the link in markdown not right. Do I need one slash in the url? I tested with `[test link](#Test)`. I was a bit in a hurry before going to work. I'll check it later today after work.

Rokt33r: I'll try it too later.

AWolf81: OK, I've fixed it by modifying parsing of the href string. Inspired by the following [post](https://www.abeautifulsite.net/parsing-urls-in-javascript).

Tested all link types (empty link, line link `:line:<number>`, tag link `#title` & note link `:note:noteKey`). I just couldn't verify old link style `storage.key-note.key` - not sure how to test it. But it should still work.

**Note:** Adding a slash to a tag link e.g. `/#title` won't work but I think that's OK - same behavior as on master.

browser/components/MarkdownPreview.js
```diff
@@ -1018,15 +1018,19 @@ export default class MarkdownPreview extends React.Component {
...
+    const linkHash = hash === '' ? rawHref : hash // needed because we're having special link formats that are removed by parser e.g. :line:10
 
-    if (!href) return
+    if (!rawHref) return // not checked href because parser will create file://... string for [empty link]()
 
-    const regexNoteInternalLink = /main.html#(.+)/
-    if (regexNoteInternalLink.test(linkHash)) {
-      const targetId = mdurl.encode(linkHash.match(regexNoteInternalLink)[1])
-      const targetElement = this.refs.root.contentWindow.document.getElementById(
+    const regexNoteInternalLink = /.*[main.\w]*.html#/
```
AWolf81: **Note**
Added `.*` to the regex in front of `[main.\w]` to match everything before `main.development.html`. The string to test contains file protocol + path e.g. `file:///c:/path/to/project/main.development.html`

https://regex101.com/r/zRLTIR/4

browser/components/MarkdownPreview.js
```diff
@@ -1018,15 +1018,19 @@ export default class MarkdownPreview extends React.Component {
...
+
+    if (regexNoteInternalLink.test(href)) {
+      const targetId = mdurl.encode(linkHash)
+      const targetElement = this.refs.root.contentWindow.document.querySelector(
```
AWolf81: Changed to querySelector because `targetId` contains `#` e.g. `#test`

browser/components/MarkdownPreview.js
AWolf81: I think `ctrl+left mouse` handling in the editor is not working for every link type, right? Tested on master. I think it's using just the open external Url link - it would be good to share the link handling code for this location so the same functionality is available there.

Should we create an issue for this so it can be improved later?

package.json
```diff
@@ -157,11 +164,12 @@
...
     "mock-require": "^3.0.1",
     "nib": "^1.1.0",
-    "react-css-modules": "^3.7.6",
+    "react-css-modules": "^4.7.9",
     "react-input-autosize": "^1.1.0",
-    "react-router": "^2.4.0",
-    "react-router-redux": "^4.0.4",
-    "react-test-renderer": "^15.6.2",
+    "react-test-renderer": "^16.8.6",
+    "redux-devtools": "^3.5.0",
```
AWolf81: There is an issue with `redux-devtools`. I'm not sure if we need to fix this before merging but I'm getting the error message from the screenshot pretty often especially if I'm selecting a dom element.

During install of `redux-devtools` I'm getting the following `warning " > redux-devtools@3.5.0" has incorrect peer dependency "react-redux@^4.0.0 || ^5.0.0 || ^6.0.0".`
There is an issue to it [here](https://github.com/reduxjs/redux-devtools/issues/473) but with-out a comment. Maybe we have to change version to `React-redux` 6.x until this is fixed.

grafik

AWolf81: I can not reproduce the behaviour with the warning message at the moment. Is this still a problem? If not I think we can stay ar react-redux 7.x and ignore the warning from `redux-devtools`.

AWolf81:
@Rokt33r I've merged master into this branch & checked the issue with devtools but couldn't reproduce it.

If it occurs later I think changing React-redux to 6.x version could improve/fix the problem. But we should avoid using the old version as they're working on hooks support improvements.

So I think this is ready for merging or is anything open that I should check?

Rokt33r:
> So I think this is ready for merging or is anything open that I should check?

You don't need to. I'm going to merge this after I merge https://github.com/BoostIO/Boostnote/pull/2983

Seems lots of people don't like this feature so I'm going to merge tomorrow and deploy again.

Rokt33r:
@AWolf81 Could you fix the conflicts? I'm going to merge this as soon as they are fixed.

AWolf81:
@Rokt33r Sure, I've merged master, did a quick test and everything is working.

PR #2992, opened 2019-05-05
Standardx setup for linting & code formatting

AWolf81:
## Description

Similar to closed PR #924 I'd like to add code formatting with pre-commit hook. It will improve development experience as you don't have to care about semicolons, spacings etc. because it will be automatically formatted on save to a consistent code style.

I first wanted to go with prettier but it was problematic to have function paren formatting working.
Then I decided to use [standard](https://www.npmjs.com/package/standard) because it's exactly doing what we need with an exception - it can't be customized.
So in the end I'm using [standardx](https://www.npmjs.com/package/standardx) - it is using standard lib under the hood but can be customized.

The only drawback of **standardx** is that vscode can only use **standard** and that's why I have added the configuration into `package.json` as well. So vscode is using `package.json` and `yarn lint` is using `.eslintrc` file.
Not perfect but I think OK to start with. Once the [PR](https://github.com/standard/vscode-standardjs/pull/71) is merged / released we can use Standardx in VS code and remove the config from `package.json`.

Styles we're having (basically according to standard style - just to mention some of the styles):
- 2 space indent
- no semis
- space before parens
- no-unused vars
- no-prop warnings JSX (disabled by custom rule)
- prefer-const

**Summary**
- Added vscode `settings` & `recommendations`
- Pre-commit prevents commiting with linting errors (see screenrecording below) (could be by-passed during dev. by `--no-verify` command line arg)
- No code changes - I've added some notes & `eslint-disable-line` to remove some warnings. We could fix these later:
  - The no-useless escapes are disabled inline because in VS code these are not ignored because the ignore setting is not applied - so I've added them
  - Not changed `ComponentWillReceiveProps` - not sure why the unsafe use warning from linting is not present
- Some new lint disables because they cause a lot of linting errors & I think it's OK to ignore them:
  - "no-loop-func"
   - "default-case"
   - "no-useless-concat"
   - "jsx-a11y/alt-text" Added because `React-App` is adding accessibility rules like this.

**Points I've noticed during setup**
1. JSX indentation is a bit problematic if not every prop is on it's own line - couldn't get the indentation right if the tag starts with `<div someProps={ } ` and then the next prop in the next line couldn't be indented correctly. (I'll add an inline comment to this later.)
2. Sometimes an additional semicolon wasn't removed by VS code but it was correctly formatted in command line - I think that's no problem but that was a bit weird.

## Type of changes

- :radio_button: Improvement (Change that improves the code. Maybe performance or **development** improvement)

## Checklist:

**Demo editing with Standardx formatting & linting**
Note: The prop typo is not detected in VS code because it's using `standard` instead of `standardx`. Formatting is happening on save (ctrl+s).
Screenrecording_Linting_formatting_with_standardx

**Demo - Failing pre-commit hook**
Note: Coloring of the error is not correctly displayed in the recording. But I've fixed that by forcing the coloring.
Screenrecording_precommit_hook

[Here](https://gist.github.com/AWolf81/5f049310e07f9e52e6eab42bc0357b5d) is a test file where I've checked the linting rules.

### DIFF
--- a/browser/components/CodeEditor.js
+++ b/browser/components/CodeEditor.js
@@ -17,14 +17,14 @@ import iconv from 'iconv-lite'
 
 import { isMarkdownTitleURL } from 'browser/lib/utils'
 import styles from '../components/CodeEditor.styl'
-const { ipcRenderer, remote, clipboard } = require('electron')
 import normalizeEditorFontFamily from 'browser/lib/normalizeEditorFontFamily'
+import TurndownService from 'turndown'
+import { languageMaps } from '../lib/CMLanguageList'
+import snippetManager from '../lib/SnippetManager'
+import { generateInEditor, tocExistsInEditor } from 'browser/lib/markdown-toc-generator'
+const { ipcRenderer, remote, clipboard } = require('electron')
 const spellcheck = require('browser/lib/spellcheck')
 const buildEditorContextMenu = require('browser/lib/contextMenuBuilder')
-import TurndownService from 'turndown'
-import {languageMaps} from '../lib/CMLanguageList'
-import snippetManager from '../lib/SnippetManager'
-import {generateInEditor, tocExistsInEditor} from 'browser/lib/markdown-toc-generator'
 
 CodeMirror.modeURL = '../node_modules/codemirror/mode/%N/%N.js'
 
@@ -113,7 +113,7 @@ export default class CodeEditor extends React.Component {
 
       function makeOverlay (query, style) {
         query = new RegExp(
-          query.replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, '\\$&'),
+          query.replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, '\\$&'), // eslint-disable-line
           'gi'
         )
         return {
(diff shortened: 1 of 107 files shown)


browser/components/CodeEditor.js
```diff
@@ -1061,23 +1062,21 @@ export default class CodeEditor extends React.Component {
     } = this.props
     const fontFamily = normalizeEditorFontFamily(this.props.fontFamily)
     const width = this.props.width
-    return (<
-      div className={
+    return (<div
```
AWolf81: This is what I've meant in the PR comment. 
```
<div className={...}
  ref='root' ...>
```
Was a problem for `standardx` formatting. But after adding the new line before `className` the formating is OK.
I think that's OK but it was a bit weird during setup because the formatter introduced a linting error.

package.json
```diff
@@ -196,9 +205,27 @@
       "jest-localstorage-mock"
     ]
   },
+  "lint-staged": {
+    "linters": {
+      "*.{js,jsx,mjs}": [
+        "npm run lint:fix",
```
AWolf81: I'm not sure with this. I think I'd change to `lint` with-out fix to avoid introducing a problem during committing. 

What do you think? Or should we use `lint:fix` and `lint` afterwards to check that the fix doesn't introduce an issue.

Rokt33r:
It sounds quite reasonable to use standardx although it cause tons of conflicts in other prs. @ZeroX-DG How do you think?

package.json
Rokt33r: Yeah it is wrong. We should do lint without fixing.

AWolf81:
We can close this as we're using prettier now (already on master).

Issue #2920, opened 2019-03-12
[Bug]TOC hasn't id attribute when the title is all Chinese characters


Aaron-Bird:
[<img alt="Issuehunt badges" src="https://issuehunt.io/r/BoostIo/Boostnote/issues/2920/badge.svg" />](https://issuehunt.io/r/BoostIo/Boostnote/issues/2920)

# Current behavior
The second title hasn't  an id attribute
So toc can't jump to '菠萝'

![1](https://user-images.githubusercontent.com/25508292/54218355-62329000-4528-11e9-99d0-fbfe1c283da5.png)
![2](https://user-images.githubusercontent.com/25508292/54218364-6494ea00-4528-11e9-92c5-9f5e083236cc.PNG)

test text:
```
# test
# 菠萝
# 芒果
```

I guess 'slugify.js' is causing trouble(./browser/lib/slugify.js)  
It deleted all non-English characters.  
```
14 slug = slug.replace(/[^\w\s-]/g, '').replace(/\s+/g, '-')
```

> There is a $30.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/BoostIo/Boostnote/issues/2920).
> - Checkout the [Issuehunt explorer](https://issuehunt.io/r/BoostIo/Boostnote/) to discover more funded issues.
> - Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on Issuehunt to raise funds.

IssuehuntBot:
@boostio has funded $30.00 to this issue.

---
- Submit pull request via [IssueHunt](https://issuehunt.io/repos/53266139/issues/2920) to receive this reward.
- Want to contribute? Chip in to this issue via [IssueHunt](https://issuehunt.io/repos/53266139/issues/2920).
- Checkout the [IssueHunt Issue Explorer](https://issuehunt.io/issues) to see more funded issues.
- Need help from developers? [Add your repository](https://issuehunt.io/r/new) on IssueHunt to raise funds.

Flexo013:
The fix for this is part of the latest release 0.12.0.

PR #2994, opened 2019-05-08
Fixes that TOC hasn't id attribute when the title is all 2-byte characters

roottool:
## Description

This PR fixes that TOC hasn't id attribute when the title is all 2-byte characters.
I referred to `browser/lib/slugify.js` in 4f9b374 and [a logic of Slug in Visual Studio Code](https://github.com/microsoft/vscode/blob/90e06350be0c25fb6c4e9a4d3ee0124dfd496a4d/extensions/markdown-language-features/src/slugify.ts#L22).

According to [RFC3986](https://tools.ietf.org/html/rfc3986#section-2.3), the underscore can be used in URI.

> unreserved  = ALPHA / DIGIT / "-" / "." / "_" / "~"

So, I don't replace the underscore.

`"."` and `"~"` is used by file path.
e.g. `~/example.js`
I think that the file path which is written in `href` of `<a>` is a security risk.
Therefore, I exclude `"."` and `"~"`.

### Screenshot

FixedSlug

## Issue fixed

#2920 

## Type of changes

## Checklist:


ZeroX-DG:
The TOC is broken, can you fix it? It tries to open the TOC link as an attachment I guess.

roottool:
> href
> A URL fragment is a name preceded by a hash mark (`#`), which specifies an internal target location (an `id` of an HTML element) within the current document.
> [\<a\>: The Anchor element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#Attributes)

When the anchor element in `<iframe>` was clicked, it can't jump the heading in `<iframe>` which is created by script.
Because the heading element doesn't exist in current HTML file.
e.g.
href

Therefore, I added `onclick` to the ancher element that the hash link exists in `href`.
jumpToHeading

browser/components/MarkdownPreview.js
```diff
@@ -895,6 +895,22 @@ export default class MarkdownPreview extends React.Component {
...
+      // Removes a hash
+      const hashLink = a.hash.substr(1)
+      const fileRegExp = new RegExp('^file://')
+      const mainHtmlRegExp = new RegExp('/lib/main.html')
+      const isHeadingInMarkdownPreview = fileRegExp.test(a.href) && mainHtmlRegExp.test(a.href)
+      const targetEl = markdownPreviewIframe.contentWindow.document.getElementById(hashLink)
+      if (hashLink !== '' && isHeadingInMarkdownPreview && targetEl) {
+        const dataLine = targetEl.dataset.line
+        a.onclick = () => {
+          this.scrollTo(dataLine)
```
ZeroX-DG: Can you move this code into this function? Because we already have an existing link click listener.
https://github.com/BoostIO/Boostnote/blob/9a6ee9d2ef2d853b623d386c928088356ba4dbce/browser/components/MarkdownPreview.js#L1014-L1024

roottool: I understood. I fixed it.

Since the link hash this time did not match the regular expression, the regular expression has been changed.
https://github.com/BoostIO/Boostnote/blob/857367efba4312b842b5ece6fd5e6912702090ff/browser/components/MarkdownPreview.js#L1029-L1030
e.g.
FixRegExp

browser/components/MarkdownPreview.js
```diff
@@ -1036,9 +1026,10 @@ export default class MarkdownPreview extends React.Component {
 
     if (!href) return
 
-    const regexNoteInternalLink = /main.html#(.+)/
+    const extractId = /(main.html)*#/
```
ZeroX-DG: Shouldn't it be `/(main.html)?#/` ?

roottool: No, I think I need this line.
First, the `id` of the target element don't have a hash.
e.g.
FixedSlug
Therefore, I had to remove a hash in `linkHash`.

Second, the past `regexNoteInternalLink` didn't match `linkHash`.
`linkHash` doesn't include `main.html`.
e.g.
FixRegExp
So, I had to change the regex rule.

ZeroX-DG: I understand that, it's just I want to replace the regex from `(main.html)*#` to `(main.html)?#`. But I think it doesn't matter :smile: We can go by your way then.

roottool: I'm sorry. I misunderstood that this line was not necessary.
Thank you for a suggestion. I think `(main.html)?#` is better than `(main.html)*#` too.

ZeroX-DG:
@roottool  Can you resolve the conflict?

roottool:
I solved the conflict by adopting the master branch code.
https://github.com/BoostIO/Boostnote/pull/2990#discussion_r283061567

Because I don't want to change the code of the current master branch.
Operation check result is like the following gif.

### Screenshot

AdjustedMasterBranch

Rokt33r:
@roottool Could you submit the pr to https://issuehunt.io/r/BoostIo/Boostnote/issues/3020?

Flexo013:
This might be related to some recent issues with 0.13.0
- #3289 
- #3287 
- #3283

PR #3001, opened 2019-05-10
Fix regular expression related checkbox logic

### LINKED ISSUE #2829: Checkbox does not work correctly in quotation
[Issuehunt badges](https://issuehunt.io/r/BoostIo/Boostnote/issues/2829)

# Current behavior
When creating a checkbox inside a quotation then the checkbox becomes unclickable.

# Expected behavior
The checkbox should be clickable in a quotation.

# Steps to reproduce

Markdown:
```
# Tasks
- [ ] `Do something`
> - [_] `Style for more lines`
> Some more lines  
```
reproduced_bug

Boostnote Editor automatically adds an underscore into the brackets when using quotation
`> - [_]`

# Environment

- Version :  Boostnote 0.11.12
- OS Version and name : macOs Sierra 10.12.6

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Sponsors (Total: $20.00)

- [boostio boostio](https://issuehunt.io/u/boostio) ($20.00)

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/BoostIo/Boostnote/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

mpls104:
evidence

## Description
As described in issue, checkbox was unclickable in quotation.
Fixed regular expression on checkbox logic and added some test patterns.

This is my first contribution to OSS in my life. :smile: 
Please kindly point out if there is something wrong 🙏 

## Issue fixed
#2829

## Type of changes

## Checklist:

### DIFF
--- a/browser/components/MarkdownEditor.js
+++ b/browser/components/MarkdownEditor.js
@@ -159,8 +159,8 @@ class MarkdownEditor extends React.Component {
     e.preventDefault()
     e.stopPropagation()
     const idMatch = /checkbox-([0-9]+)/
-    const checkedMatch = /^\s*>?\s*[+\-*] \[x]/i
-    const uncheckedMatch = /^\s*>?\s*[+\-*] \[ ]/
+    const checkedMatch = /^(\s*>?)*\s*[+\-*] \[x]/i
+    const uncheckedMatch = /^(\s*>?)*\s*[+\-*] \[ ]/
     const checkReplace = /\[x]/i
     const uncheckReplace = /\[ ]/
     if (idMatch.test(e.target.getAttribute('id'))) {
--- a/browser/components/MarkdownEditor.js
+++ b/browser/components/MarkdownEditor.js
@@ -159,8 +159,8 @@ class MarkdownEditor extends React.Component {
     e.preventDefault()
     e.stopPropagation()
     const idMatch = /checkbox-([0-9]+)/
-    const checkedMatch = /^\s*[\+\-\*] \[x\]/i
-    const uncheckedMatch = /^\s*[\+\-\*] \[ \]/
+    const checkedMatch = /^\s*>?\s*[\+\-\*] \[x\]/i
+    const uncheckedMatch = /^\s*>?\s*[\+\-\*] \[ \]/
     const checkReplace = /\[x\]/i
     const uncheckReplace = /\[ \]/
     if (idMatch.test(e.target.getAttribute('id'))) {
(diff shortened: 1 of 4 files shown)


ZeroX-DG:
I'm gonna be criticizing as hard as I can :rofl:

ZeroX-DG:
It currently doesn't work for a nested quote like this. I don't know if we should support it.
image

mpls104:
Thank you for checking precisely  😄 
I think it's better to support this and done some changes.
What do you think @ZeroX-DG ?

ZeroX-DG:
Well, I'll need confirmation from the "higher" power @Rokt33r

AWolf81:
LGTM.

Just one thing I've noticed is that sub-elements are striked through if the top element is checked. I think you haven't introduced the issue but it's a bit weird and it would be probably better to not have them striked as they're not handled in the action item done count. See screenshot below, everything seems done but the counter says 60%.
But I think we could improve that in a separate PR if it's difficult to fix.

grafik

I'll have a look at your regex updates in a minute and leave an inline comment if there is something to improve.

browser/components/MarkdownEditor.js
```diff
@@ -159,10 +159,10 @@ class MarkdownEditor extends React.Component {
     e.preventDefault()
     e.stopPropagation()
     const idMatch = /checkbox-([0-9]+)/
-    const checkedMatch = /^\s*[\+\-\*] \[x\]/i
-    const uncheckedMatch = /^\s*[\+\-\*] \[ \]/
-    const checkReplace = /\[x\]/i
-    const uncheckReplace = /\[ \]/
+    const checkedMatch = /^(\s*>?)*\s*[+\-*] \[x]/i
+    const uncheckedMatch = /^(\s*>?)*\s*[+\-*] \[ ]/
```
AWolf81: I've first thought there are missing escape sequences in the regex but I think it's OK to omit them.
I did a quick test in the following [regex101](https://regex101.com/r/LzQcdc/2) - just wondering why the nested checkboxes are not matching there.
Is it because in Boostnote it is running in a loop or do I have a mistake in the regex101 test patterns?

ZeroX-DG:
@AWolf81 The nested checkbox is strikethrough is because of the CSS property
```
text-decoration: line-through;
```

If that CSS property is applied to the parent element, the children element will be affected too which is a bit weird. You're right, it should be solved in a separated issue, it may be difficult to solve.

arcturus140:
i think this strike through behaviour is intended, but I also think it is weird 😣

browser/components/MarkdownEditor.js
arcturus140: it's a line pattern and only matched the first occurrence

Rokt33r:
LGTM. And I'm also thinking it is a bit weird. Let's fix it in other pr.

Issue #1683, opened 2018-03-15
#tags within body of markdowns?


ChetCopperpot:
[<img alt="Issuehunt badges" src="https://issuehunt.io/r/BoostIo/Boostnote/issues/1683/badge.svg" />](https://issuehunt.io/r/BoostIo/Boostnote/issues/1683)

New to the product and so far love it!

Hopefully this is a non-issue, but I'd be surprised if this functionality was not already available, because it seems like an incredibly valuable feature. 

I'm (unsuccessfully) attempting to insert #tags within the body of markdowns, eg:

Markdown To-Do List:
- [ ] Item 1
- [ ] Item 2 #tag1
- [ ] Item 3 #tag2

As a user, I would expect to be be able to click the tag associated with "Item 2" and pull up a list of all markdowns / snippets with that tag; however, I can only figure out how to add tags to a markdown as a whole as opposed to tagging individual parts within a single markdown.

Please tell me this functionality already exists...if not, it should!

> There is a $45.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/BoostIo/Boostnote/issues/1683).
> - Checkout the [Issuehunt explorer](https://issuehunt.io/r/BoostIo/Boostnote/) to discover more funded issues.
> - Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on Issuehunt to raise funds.

IssuehuntBot:
@kazup01 funded this issue with $1. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/1683)

IssuehuntBot:
@anuragsachdeva28 has started working. [Visit this issue on Issuehunt](https://issuehunt.io/repos/53266139/issues/1683)

IssuehuntBot:
@kazup01 cancelled funding, $1, of this issue. [See it on IssueHunt](https://issuehunt.io/repos/53266139/issues/1683)

IssuehuntBot:
@kazup01 has cancelled funding for this issue.(Cancelled amount: $40.00) [See it on IssueHunt](https://issuehunt.io/repos/53266139/issues/1683)

IssuehuntBot:
@issuehuntfest has funded $40.00 to this issue. [See it on IssueHunt](https://issuehunt.io/repos/53266139/issues/1683)

aashutoshrathi:
Hi, @ChetCopperpot.
I have a fix for this.
[...]

PR #3002, opened 2019-05-11
Add tag link handling with :tag:#tag syntax

AWolf81:
## Description
I've added the tag links by using `:tag:#tag` link syntax. 
It's working but not ready for merge because I'd like to wait for React upgrade because I'm using hashHistory directly. This will be changed later to `dispatch(push(...))`

## Issue fixed

#1683

## Type of changes

## Checklist:

- :radio_button: I have attached a screenshot/video to visualize my change if possible --> see issue

## Note
The tag links are not handled in editor just in `MarkdownPreview` and I think we should open a new issue for this as I think the other link types (:line) are also not handled - or is line & tag handling only required in preview? I think they should be handled as there is a tooltip indicating how to click but the click is not working.


AWolf81:
@ZeroX-DG sure, I've resolved the conflict and checked the feature.

I'll open an issue (if it's not already in the issues) for the Crtl+Link click that's not working for tag and line links as mentioned in my note above.

For the failing Travis CI I think that's caused by linting errors. I'll create a separate PR for it as that's more work do to (e.g. check prettier with standard.js usage).

Rokt33r:
@AWolf81 @ZeroX-DG I don't think we need to keep `#`. `:tag:tag` should be enough. How do you think?

ZeroX-DG:
That could be a good idea. Keep the consistency with other type of links like `:note:` or `:line:`. I'm OK with either way so @AWolf81 what do you think about this?

AWolf81:
I would keep it with the hash as it's easier to remember and I think it would be confusing to users. As the tag is always displayed with the # in front.
The consistency would be better and it's an easy change - just removing the hash from the Regex.

Maybe if there would be a context menu to add tag links we could do it with-out #. But we should create an issue for this and discuss it there e.g. right click on tag shows `Insert link to tag`.

Rokt33r:
@AWolf81 

I think prefixing `:tag:` is already enough to prevent the confusing. So yeah, I want to keep consistency so users can expect same behavior of other internal links. So please fix it. :pray: 

>  Maybe if there would be a context menu to add tag links we could do it with-out #. But we should create an issue for this and discuss it there e.g. right click on tag shows Insert link to tag.

Sounds good to me but should be better to presented in a separated PR.

AWolf81:
@Rokt33r sure. I've fixed it. 👍 

For the context menu I'll create an issue and start working on it later today.

Issue #3007, opened 2019-05-14
A xss on the newest version


fr4nk404:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%2440%20Funded-%2300A156.svg" />](https://issuehunt.io/r/BoostIo/Boostnote/issues/3007)

# XSS 
There is a xss in the newest version via a label named `mermaid`
When we insert codes like this:

```mermaid
graph LR
id1["<iframe src=javascript:alert('xss')></iframe>"]
```
we can see there is a xss in the latest version.
![image](https://user-images.githubusercontent.com/16134876/57669403-6a4ca000-763d-11e9-9bb6-cdb487a7a185.png)

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Sponsors (Total: $40.00)

- [<img src='https://avatars3.githubusercontent.com/u/13612118?v=4' alt='boostio' width=24 height=24> boostio](https://issuehunt.io/u/boostio) ($40.00)

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/BoostIo/Boostnote/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

ZeroX-DG:
Good catch! I'll check it when I get home.

permutate:
https://nvd.nist.gov/vuln/detail/CVE-2019-12184

IssueHuntBot:
@boostio has funded $40.00 to this issue.

---
- Submit pull request via [IssueHunt](https://issuehunt.io/repos/53266139/issues/3007) to receive this reward.
- Want to contribute? Chip in to this issue via [IssueHunt](https://issuehunt.io/repos/53266139/issues/3007).
- Checkout the [IssueHunt Issue Explorer](https://issuehunt.io/issues) to see more funded issues.
- Need help from developers? [Add your repository](https://issuehunt.io/r/new) on IssueHunt to raise funds.

amedora:
@Flexo013, could you close this? (resolved by #3028)

Flexo013:
Thanks for the ping @amedora!

PR #3028, opened 2019-05-24
Fix 3007

amedora:
## Description

Not only for mermaid, other code block like 'sequence' as below has potentially XSS problem.
```
```sequence
<iframe src="javascript:alert('xss');">
\```
```

## Issue fixed

#3007 

## Type of changes

## Checklist:


browser/components/MarkdownPreview.js
```diff
@@ -845,7 +845,7 @@ export default class MarkdownPreview extends React.Component {
     _.forEach(
       this.refs.root.contentWindow.document.querySelectorAll('.mermaid'),
       el => {
-        mermaidRender(el, htmlTextHelper.decodeEntities(el.innerHTML), theme)
+        mermaidRender(el, htmlTextHelper.encodeEntities(el.innerHTML), theme)
```
ZeroX-DG: Your code fixed the XSS but the mermaid is unusable after your fix.
Example:
```
```mermaid
sequenceDiagram
loop every day
    Alice->>John: Hello John, how are you?
    John-->>Alice: Great!
end
```

The encodeEntities encoded the content of the mermaid code so it becomes invalid. It seems encodeEntities is not a good way to fix this bug. @Rokt33r what do you suggest for this? Maybe a write a plugin to parse mermaid directly from markdown?

AWolf81: Why is this change needed? I think the XSS is also working with-out this change - tested locally with just the `_fence` change in `markdown-it-sanitize-html.js`. 
Tested with the following [gist](https://gist.github.com/rebirthwyw/9836b2445abd4c1358e3d68ff2415247).

amedora: > Your code fixed the XSS but the mermaid is unusable after your fix.
oh, nice catch. I confirmed that too.

@AWolf81 
Even if we sanitized a content in markdown-it-sanitize-html.js, it's going to be original in MarkdownPreview.js:715
```js
    const renderedHTML = this.markdown.render(value)
    attachmentManagement.migrateAttachments(value, storagePath, noteKey)
    this.refs.root.contentWindow.document.body.innerHTML = attachmentManagement.fixLocalURLS(
      renderedHTML,
      storagePath
    )
```
I think we have to do something for mermaid only (other plugins don't have this vulnerability).

amedora:
I think that most straightforward way to fix this is passing "flowchart: { htmlLabels: false }" option to mermaidAPI.initialize().
https://github.com/BoostIO/Boostnote/blob/76335f78aca94b39841b03af4c212bec65862e5b/browser/components/render/MermaidRender.js#L29-L33
However it would be drawback for users who use HTML in mermaid charts.

AWolf81:
@amedora maybe we can add this as option - disabled html labels by default. With a warning at the option that there is a risk of xss. So Mermaid html label users can decide and accept the risk.

That's not perfect but it will fix xss. Maybe we can open an issue at Mermaid so we can track if there will be a fix.

amedora:
It's now configurable in the option panel. works perfect on my end.
option

ZeroX-DG:
LGTM :tada: But I'll need review from @Rokt33r before I can approve this.

Issue #3025, opened 2019-05-23
MAke markdownlinter to switch off


takacsot:
Please make an option to switch of markdown linter.

AWolf81:
Once PR #2983 is merged you can disable linting with the markdown config field:
```json
{
  "default": false
}
````

Maybe a checkbox to enable/disable markdown linting would be better instead of the config. Or is the config input enough?

d3vnick:
I am wondering how this feature has been accepted in the first place. One of the most appreciated "feature" of the Markdown markup language is its flexibility and syntax error tolerance. Also it nice to put some extra line when you read it (not parsed).

benbucklin:
@AWolf81 I think the config input is a good workaround for now, but having a checkbox would be much more intuitive and save a lot of users from having to dig for a solution. When I saw the linter warnings the first thing I did was look for a checkbox under "Interface".

Flexo013:
0.11.17 has been released and allows for configuration of the Markdownlinter. By default linting will be off.

PR #3034, opened 2019-05-24
Enable Markdown lint option

AWolf81:
After merging #2983 we can add this to toggle markdownLinting.

I've based this on branch `roottool:add-markdownlint-rules-form` from PR #2983

## Description

- Added `enableMarkdownLint` setting
- Add/remove gutter based on `enableMarkdownLint` 

## Issue fixed

#3025

## Type of changes

## Checklist:

## Todos:
- [x] Fix gutter toggling

## Discussion
- Is the position of the option OK or should it be closer to the config area for markdown lint? I've moved it to the textarea because I think that location is better.
- Is setting `enableMarkdownLint` to false by default OK?
- Any ideas how I can fix the gutter toggling? It seems to not display the linting errors after re-enabling the linting. I've found a work-around by toggling the gutter with `style.display` and keep it in the config. Not sure if there is a better way to handle it but it's working now.

## Screenshot
**Current setting**
grafik

**Below is the location from my first draft (changed it to the above location)**
grafik


browser/components/CodeEditor.js
```diff
@@ -644,6 +626,55 @@ export default class CodeEditor extends React.Component {
...
+  }
+
+  validatorOfMarkdown (text, updateLinting) {
+    const { customMarkdownLintConfig } = this.props
+    let lintConfigJson
+    try {
+      Jsonlint.parse(customMarkdownLintConfig)
+      lintConfigJson = JSON.parse(customMarkdownLintConfig)
+    } catch (err) {
+      throw err
```
ZeroX-DG: I don't think throwing an error here is a good thing to do because it will make the app unusable. I think you should put something more "user-friendly" for example not applying the lint config and display error message above the custom lint config box in the setting.

ZeroX-DG:
And please remove all `console.log` unless you think it's useful for debugging and bug reporting in the future.

browser/components/CodeEditor.js
AWolf81: Yes, you're right. It's bad to throw an error with-out handling it.
How can I add the error message display? I've tried to change the `throw err` to 
```js
eventEmitter.emit('APP_SETTING_ERROR')
return
```

But I don't like how it's working. It first shows success message and then the error occured message. It happens pretty fast - short flicker of the green message. Also on a second click it won't show the error again.

The `catch` can be triggered by entering invalid json for the custom markdown lint rule and then press the save button.

AWolf81: A custom error message wasn't working - not sure why. I wanted to add something like: `Failed to apply custom linting rule.`

Tried it with `eventEmitter.emit('APP_SETTING_ERROR', {message: 'Failed to apply...'})`

Rokt33r:
Submit this pr to https://issuehunt.io/r/BoostIO/Boostnote/issues/3038 :)

judsoniv:
Thank you! I've enabled linting and configured it the way I want. Thank you for being open to the feedback.

Issue #2925, opened 2019-03-14
Can't type chinese when searching


lonelyma1021:
I use two different chinese input method. Both of them can not type chinese in searching bar.  But I can type English letter there. I think it's a bug.

# Environment

- Version :0.11.15
- OS Version and name : Deepin linux 15.9 & windows 7

liupengfei66:
i had the same problem.

env: mac os, version 0.11.15

johan456789:
Doesn't work with Japanese too. Probably doesn't work for CJK languages.

forestail:
This is the same as #2885.

JunweiZheng93:
same problem +1

sdvdxl:
me too mac os 10.14.4 (18E226) ，Boostnote 0.11.15，0.11.13

Ronnie-6606:
same problem +1

mac os 10.13.3 boostnote 0.11.15

ANTON072:
same problem +1

mac os 10.14.5
boostnote 0.11.17

copetit:
also Korean. downgradinng to 0.11.10 , can type.

Flexo013:
This now possible in the current version of Boostnote.

Issue #3030, opened 2019-05-24
Finding a note Chinese input can't be entered


yuamera:
# Current behavior
![image](https://user-images.githubusercontent.com/50633505/58310344-02663880-7e39-11e9-9369-a34fa58c1144.png)

# Expected behavior
Finding a note Chinese input can't be entered
I am using mac sogou
https://github.com/yuamera/image/blob/master/issue-gif/boostnote-issue.gif

# Environment

- Version :0.11.16
- OS Version and name : macOS Mojave 10.14.4

jhdcruz:
It's probably a `Font` or `locales` problem...

yuamera:
When can I fix it?
Can't search Chinese is equal to not searching for a user who uses Chinese

ZeroX-DG:
@yuamera There's a pull request to fix this issue and it has been approved. This bug will be fixed in the next version of Boostnote.

ThaddeusJiang:
@ZeroX-DG when we get the new release?

ZeroX-DG:
That's depend on @Rokt33r

Rokt33r:
@ThaddeusJiang In the next week! Sorry for being late.

Flexo013:
Closing this issue as the next release 0.12.0 fixes this issue.

PR #2942, opened 2019-03-22
Fixes search input dropping key input

seanders:
## Description
This PR fixes https://github.com/BoostIo/Boostnote/issues/2843.

The source of the issue seems to have been the asynchronous setting of the component's state. Prior to this change, the search input's state was being updated by a debounced `updateKeyword` function. Pushing these updates to the event queue was causing keystrokes to be lost. The image below should make it more clear:

**Before** this code change:
image
---
**After** this code change:
image

## GIFs
Here is a gif of me trying to type `bacon` as fast as possible

#### Before the fix.
Characters dropped bug

#### After the fix.
Characters dropped bug FIXED

## Issue fixed
https://github.com/BoostIo/Boostnote/issues/2843

## Type of changes

- :radio_button: Bug fix (Change that fixed an issue)
- :white_circle: Breaking change (Change that can cause existing functionality to change)
- :white_circle: Improvement (Change that improves the code. Maybe performance or development improvement)
- :white_circle: Feature (Change that adds new functionality)
- :white_circle: Documentation change (Change that modifies documentation. Maybe typo fixes)

## Checklist:

- :radio_button: My code follows [the project code style](docs/code_style.md)
- :white_circle: I have written test for my code and it has been tested
- :radio_button: All existing tests have been passed
- :radio_button: I have attached a screenshot/video to visualize my change if possible

PR #2919, opened 2019-03-09
Chinese, Japanese, Korean are now available on search box. (fix for #2885)

forestail:
## Description

fix_for_2885

## Issue fixed
fix for #2885

## Type of changes

- :black_circle: Bug fix (Change that fixed an issue)
- :white_circle: Breaking change (Change that can cause existing functionality to change)
- :white_circle: Improvement (Change that improves the code. Maybe performance or development improvement)
- :white_circle: Feature (Change that adds new functionality)
- :white_circle: Documentation change (Change that modifies documentation. Maybe typo fixes)

## Checklist:

- :white_circle: My code follows [the project code style](docs/code_style.md)
- :white_circle: I have written test for my code and it has been tested
- :white_circle: All existing tests have been passed
- :white_circle: I have attached a screenshot/video to visualize my change if possible

PR #3037, opened 2019-05-25
Simplified search input & fixed chinese character input

AWolf81:
## Description

- Reduced complexity in seach input string handling & removed some states
- Tested chinese character search entry
- Fixed focus issue that occured if your doing fast typing and backspace delete. That caused the editor area to be focused instead of the search input.

There are mutliple open PRs that are related to search input:
- #2942 - dropping key - I'll check if that is still happening here. I couldn't reproduce it here.
- #2919 - also adds support for chinese characters by modifing `handleSearchChange`. I'm doing the same but with less code if it's OK to remove the `isIME` checking
- #2433 - support of chinese chars by adding a `compositionend` eventl istener with a new state `isIMEdone` that will be set with `compositionend` - I think that's not really needed as directly handling in change handler should work.
- #2676 - Input only applied on backspace - I haven't noticed that here. It would be good to check the mentioned keyboards here too.

We should check if all of them are fixed with this PR.

## Issue fixed

#2925, #3030, #2843

## Type of changes

## Checklist:

## Screenshot
**Note just some chinese letters in there to show that you can now enter these characters**
grafik

## Todos:
- [x] Remove state variables `isAlphabet, isIME, isConfirmTranslation` (if it's OK to simplify this, see question below)
- [x] Add enter key handling to re-apply search term - handy if you navigate away and you want to use the same search term

## Discussion
- Is it OK to change the search input to not check the character range? I think that check is not really needed and just adds complexity. But please let me know if there is a reason I'm missing or a case that's not working with my change.
- Please double check the chinese character entry. It should work but it's a bit difficult to test on a German laptop.
- How can I test if my change to `switchPreview === 'BLUR' || switchPreview === 'DBL_CLICK'` is working? That change was needed to fix the focus loss but I'm not sure where it is used.
- Fast typing issue: I'm not sure if it's fixed here because it could be related to notebook size. I have just a few notes and so searching is pretty fast. How can I debug this?


amedora:
Unfortunately this changes break MS-IME behavior.
I can no longer input Japanese correctly... (tested on Win7/64)
ime

AWolf81:
Oh, OK, can you please test if the following [demo](https://leoeatle.github.io/react-composition-input/) is working for you?
That seems to work & maybe we can use it or implement it directly. What do you think? Is it OK to use that component from npm?

It is using composition events to trigger the change event once the IME is done. I would use it as it's easier to read.

amedora:
> Oh, OK, can you please test if the following [demo](https://leoeatle.github.io/react-composition-input/) is working for you?

works nice :+1: 
inputdemo

I'm not sure it's ok or not to use that component, but looks good.

AWolf81:
@amedora can you please give it another try? I've added the composition input.
I have one point in the code where I'm not sure. I'll add an inline comment.

browser/main/TopBar/index.js
```diff
@@ -165,7 +129,7 @@ class TopBar extends React.Component {
   }
 
   handleCodeInit () {
-    ee.emit('top:search', this.refs.searchInput.value)
+    ee.emit('top:search', this.refs.searchInput.value || '')
```
AWolf81: Why is this `emit` there? With out adding the `|| ''` I've got an error from `CodeEditor.js` because it seems that the value is undefined here.
How can I test this?

browser/main/TopBar/index.js
```diff
@@ -84,58 +87,19 @@ class TopBar extends React.Component {
...
-    }
-  }
-
-  updateKeyword (keyword) {
-    this.context.router.push(`/searched/${encodeURIComponent(keyword)}`)
-    this.setState({
-      search: keyword
-    })
-    ee.emit('top:search', keyword)
+    const keyword = e.target.value
```
AWolf81: Here I wanted to use `this.refs.searchInput.value` but it wasn't working - maybe there is an issue with `CInput` and the ref but with `event.target.value` it is woking.
So that's OK and I just wanted to mention it.

amedora:
LGTM. works like a charm :+1:

browser/main/TopBar/index.js
Rokt33r: LGTM. I didn't notice that it can be undefined.

Rokt33r:
@ZeroX-DG Could you test this too? I want to merge this soon. :)

ZeroX-DG:
Works great for me!

Issue #923, opened 2017-09-30
plugin to scrape website & convert HTML to markdown


100ideas:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%24100%20Rewarded-%237E24E3.svg" />](https://issuehunt.io/r/BoostIo/Boostnote/issues/923)

I'm enjoying boostnote after switching from evernote & quiver.app - thank you to everyone who has contributed to this promising open source tool.

I keep a "code" notebook for technical notes-to-self and today I wanted to add a "clipping" of a [blog post](https://shapeshed.com/command-line-utilities-with-nodejs/) to it. I wasn't sure what the best way was (sometimes I try copying-and-pasting directly from the browser, which worked OK in quiver's rich-text note mode... but rtf, gross), so I tried out a few tools for automatically converting from HTML to markdown.

[`pandoc`](http://pandoc.org/MANUAL.html) has a command-line option to fetch content from URL and can convert to/from HTML, markdown, and many other formats.  Install on osx with `brew install pandoc`, then:

```bash
pandoc -f html --normalize --wrap=none -t markdown_github+backtick_code_blocks+autolink_bare_uris -o output.md <URL>
```
as a handy [fish](https://fishshell.com/) shell function:
```fish
❯ function panscrape --description='usage: panscrape [URL] > blog_clipping.md'
      pandoc -f html --normalize --wrap=none -t markdown_github+backtick_code_blocks+autolink_bare_uris $argv
  end
❯ funcsave panscrape
❯ panscrape "https://shapeshed.com/command-line-utilities-with-nodejs/" > clipping.md

# or to copy directly to system clipboard
❯ panscrape "https://shapeshed.com/command-line-utilities-with-nodejs/" | pbcopy
```

Pandoc does an OK job but isn't definitely not perfect, so some manual editing of the output may be necessary, for instance deleting header & footer content. 

![html-md-conversion-pandoc-boostnote](https://user-images.githubusercontent.com/57006/31045328-2ebcc570-a596-11e7-8b04-b5f5cbcb42dc.png)

If you don't want to install anything, [fuckyeahmarkdown.com](http://fuckyeahmarkdown.com) seems to have an alright hosted converter.

### feature request
Add a command (plugin?) to Boostnote that takes a URL as input, scrapes the page, converts the html to markdown, and creates a new note filled with the result.

Starting points:
[...]

Issue #1980, opened 2018-05-26
v0.11.5 image not shown


eugenia1815:
# Current behavior
<img width="1192" alt="image" src="https://user-images.githubusercontent.com/33351735/40571818-fb3437de-60e2-11e8-9996-9c224999a69d.png">

# Expected behavior
<img width="1201" alt="image" src="https://user-images.githubusercontent.com/33351735/40571791-9843a52e-60e2-11e8-9f88-90594a2110c0.png">

# Environment

- Version : v0.11.5
- OS Version and name : Sierra 10.12.6

Seanybalabala:
Yes, I have encounted the issue too. Now, images can only be displayed by using absolute path which is quite annoying.

i5uhail:
Able to reproduce the same. Hopefully a fix will be in place soon.

hminle:
I have this issue too, I don't want to upgrade. How to stop upgrade, because v0.11.4 is better, it does not have image issue

hminle:
This is a huge bug, can anyone help?

ehhc:
hi together, i fear it is my fault. I explained the problem here: https://github.com/BoostIO/Boostnote/issues/1941#issuecomment-390644169
I am very sorry for the problems i caused.. i've already submitted a PR that hopefully fixes the problem, but it can't be released before the next version (in 2-3 weeks,,,).

@OnlineVagrant  what is the problem with drag&drop? If there is an error in the development console, could you please post it?

as a current workaround: if you rename the folder "images" at your boostnote storage location to "attachments" it might work again (see the comment linked above for the explanation)

OnlineVagrant:
@ehhc When in edit mode drag and drop works well but in the split view mode it does not display when dragging and dropping. I just realised if you move the cursor by adding another photo or pressing enter the photo gets displayed.

There is an error but I think it is unrelated.

![one](https://user-images.githubusercontent.com/39413259/40581200-a522cd06-6128-11e8-9753-29f3721bfc04.JPG)

![two](https://user-images.githubusercontent.com/39413259/40581203-b09cf9a4-6128-11e8-94cd-85942b2c1bcb.JPG)

![error](https://user-images.githubusercontent.com/39413259/40581204-b7c22ede-6128-11e8-9548-0f7c967098f9.JPG)

ehhc:
@OnlineVagrant thank you for your answer. Yes it is unrelated but it will be (hopefully) fixed with the next release as well :)
[...]

PR #3099, opened 2019-06-29
Html to md feature

AWolf81:
## Description

I'm addressing the requested review from PR #1980. If I've missed anything please let me know. @StormBurpee I've based this on your branch and addressed the review comments - hope that's OK.

I think the feature is working as expected. The only thing that I have noticed is that the image is not copied to the created note.

What do you think can we address that after this PR or should I put some work into it and check it? But I think we would have to discuss this a bit e.g. copy the image as an asset or create a link to the original location. I think copying would be better.

**To the test:**
Is the test OK? It just checks that the created note is written to disc and that the keys are correct. I think that's OK as we don't have to check that `Turndown` is converting the html to markdown correctly.
Or is there anything else to test?

## Issue fixed

#923 

## Type of changes

## Checklist:

## Screenshots
**Link below the other new note options**
grafik

**Import dialog with error text**
grafik

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

This pull request has been submitted to:
- [#923: plugin to scrape website & convert HTML to markdown](https://issuehunt.io/repos/53266139/issues/923)
---

IssueHunt has been backed by the following sponsors. [Become a sponsor](https://issuehunt.io/membership/members)
</details>


browser/main/lib/dataApi/createNoteFromUrl.js
```diff
@@ -0,0 +1,90 @@
...
+
+function createNoteFromUrl (url, storage, folder, dispatch = null, location = null) {
+  return new Promise((resolve, reject) => {
+    const td = new TurndownService()
+
+    if (!validateUrl(url)) {
+      reject({result: false, error: 'Please check your URL is in correct format. (Example, https://www.google.com)'})
+    }
+
+    const request = url.includes('https') ? https : http
```
ZeroX-DG: Can you replace this with `url.startsWith("https")` instead?

ZeroX-DG:
@AWolf81 Can you fix the wrong color for dracula theme?
image

Also,  it seem that the code doesn't work on tables, can you check again? For example, try this URL and take a look at the table section.
https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet

arcturus140:
@ZeroX-DG darcula not dracula 🤣

AWolf81:
@ZeroX-DG fixed Dracula theme and added the gfm turndown plugin.

Table generation is not working perfectly but I'm not sure how to fix it. Text between two tables is not correctly rendered (see below screenshot). Adding a new line after the first table fixes the rendering. With-out the space it is rendering the three parts in tbodys. With the space the text is a p-tag and the rest are two tables.
I'll create a minimal example to reproduce the issue and maybe I ask at turndown repo. why it is occuring.

grafik

browser/main/lib/dataApi/createNoteFromUrl.js
```diff
@@ -0,0 +1,90 @@
...
+
+      res.on('data', (chunk) => {
+        data += chunk
+      })
+
+      res.on('end', () => {
+        const html = document.createElement('html')
+        html.innerHTML = data
+
+        const scripts = html.getElementsByTagName('script')
```
ZeroX-DG: @AWolf81 Instead of doing this, can we use the turndown remove rule?
https://github.com/domchristie/turndown#removefilter

AWolf81: @zerox-DG good idea. I'll change this in one week as I'm without a computer at the moment.

`remove` filter should simplify this.

AWolf81: @ZeroX-DG OK, I've changed the filtering. I've created a  basic Codesandbox example to check that the `script` tag is not available in Bootstnote.

https://4bckd.csb.app/

If you open the URL in the browser it will execute an `alert` and in BoostNote there is no `script` tag available.
Is there more I can test?

The `html` constant can be removed as Turndown is handling the conversion string to md - it was just there for the script filtering.

ZeroX-DG: I think it looks good now. Thank you for your contribution!

Flexo013:
Found an edgecase where this can crash the new note menu:
https://github.com/BoostIO/Boostnote/issues/3308

Issue #2876, opened 2019-02-09
Adding a Storage that you have unlinked once throws an error when unlinking it again


indrive:
# Current behavior
When adding a storage a second time that you have unlinked before, will throw the following error:

![2019-02-08_19-21-35](https://user-images.githubusercontent.com/47465744/52514405-04ccbb80-2bd7-11e9-98bc-16e1128ab173.png)

![2019-02-08_19-22-55](https://user-images.githubusercontent.com/47465744/52514410-144c0480-2bd7-11e9-95ba-2cd5479043b7.png)

This only happens when the storage has folders with notes in them but does not happen when the storage is empty or when the storage has empty folders.

This is error is also thrown when loading Boostnote for the first time:
```
C:\Users\{Username}\Boostnote\notes doesn't exist
resolveStorageNotes
```
When a storage is added with a folder and a note for the first time, it throws the following error:
```
{Storage Location}\notes doesn't exist
resolveStorageNotes..

Attachment folder ("{Storage Location}\attachments\49b4041a-fe96-4cdf-bb6b-621bd6adcf84") did not exist...
```

# Expected behavior

It shouldn't throw an error after adding a storage that you have unlinked before.

# Steps to reproduce

1. Add a new storage
2. Add a Folder to the new storage
3. Add a note to that folder you just created in the new storage
4. Unlink the new storage
5. Add the new storage again
6. Unlink it again

# Environment

- Version : v0.11.15
- OS Version and name : Windows 10

PR #3168, opened 2019-07-26
When storage or folder is removed, Detail components should render without error

hikerpig:
## Description

Edit a note and unlink its folder or storage, current implementation of `browser/main/Detail/MarkdownNoteDetail.js` will throw exception due to invalid object property access, this PR add a safe guard to prevent it.

## Issue fixed

- #2876

## Type of changes

## Checklist:


ZeroX-DG:
@hikerpig can you resolve the conflict please?

hikerpig:
@ZeroX-DG Done 🎉

ZeroX-DG:
@hikerpig  thank you for resolving the conflict. It seem the bug is still there, I haven't have the time to look into it yet, if you can investigate this bug before me please do:)
image

hikerpig:
@ZeroX-DG  I see many features are deployed after my initial commit. All this errors occur when no storage matches the storageKey, some of them are purposely thrown by this [`findStorage`](https://github.com/BoostIO/Boostnote/blob/master/browser/lib/findStorage.js#L3:L3) method.

I just added some guards to avoid some scenarios, currently my editor won't break after storage is unlinked. 

But I see there are still some places called `findStorage` without a try/catch guard, maybe a thorough mine-sweeping is needed?

ZeroX-DG:
@hikerpig yes, I check out some places that call `findStorage` and I think it's not very safe to not catch the error at all. But catching the error is a difficult work, you have to choose to ignore an action or show a dialog to the user so if you want we can separate it into another PR.

hikerpig:
That will be great. This PR should be done by now.

And I think instead of throwing an error in `findStorage`, it should just return null if no matches found, and all callers should keep in mind and handle the null result, a simple `if (storage)` test would be better than try/catch. How do you feel about this @ZeroX-DG ?

If that's ok, I will take a look in this and open a draft PR in one or two days.

ZeroX-DG:
@hikerpig that's actually a good suggestion since `findStorage` only throw 1 error, we can reduce that to return null instead. BTW, I think the bug is not quite fixed yet, after you perform all the steps to reproduce the bug, everything works ok untill you click on the tag list button on side bar or when you add and unlink the storage another time. The new bug is caused by this function:

https://github.com/BoostIO/Boostnote/blob/5f56d3e0de48e3c49b454558caac5c81ea4430cf/browser/main/NoteList/index.js#L1108

I think it's caused by inconsistent between note list and storage list since the storage has been removed but the note list still try to access it. Can you investigate and fix it also? Thank you :+1:

hikerpig:
@ZeroX-DG I've fixed the errors in `NodeList/index.js`, have a look ~

PR #3175, opened 2019-07-28
optimize: should highlight any non-empty search query, fix #3164

### LINKED ISSUE #3164: Chinese characters are not highlighted by search queries
# Current behavior

can't highlight search word(Chinese)
[Screenshot from Gyazo](https://gyazo.com/e155c4e4810aeb98703598093f7a8108)

# Expected behavior
The same as the English word.

[Screenshot from Gyazo](https://gyazo.com/3d901ce156d9650b58d01d13a13dee9d)

# Steps to reproduce

1. focus search input
2. type or copy "测试" in that

# Environment

- Version : 0.12
- OS Version and name : MacOS 10.14

hikerpig:
## Description

Digg a little into the situation that #3164 describes, turns out it has nothing to do with unicode characters.

The original `msg.length < 3` threshold may be out of performance concern, but since I don't have that much notes kept, it doesn't seem an issue to me.

## Type of changes

### DIFF
--- a/browser/components/CodeEditor.js
+++ b/browser/components/CodeEditor.js
@@ -106,7 +106,7 @@ export default class CodeEditor extends React.Component {
     const component = this
 
     if (component.searchState) cm.removeOverlay(component.searchState)
-    if (msg.length < 3) return
+    if (msg.length < 1) return
 
     cm.operation(function () {
       component.searchState = makeOverlay(msg, 'searching')


ZeroX-DG:
@hikerpig I think you should allow user to be able to customize this. Because some people would prefer performance more so it would be great if you can add a setting for this 👍

hikerpig:
@ZeroX-DG  How about a `editor.searchHighlightQueryLengthThreshold` number setting? By default it should be 1.

Though I don't think it's necessary, since this highlight will not change any React vdom. It only operates current editor's CodeMirror instance, add some little highlight overlays ontop of matched contents.

ZeroX-DG:
hmm....you're right, we can just leave it as `1` since it doesn't affect anything :+1:

Issue #3060, opened 2019-06-06
TOC Links broken in HTML export


Zorlin:
# Current behavior

When exporting a note to HTML with a Table of Contents, the links are broken. The exact same note worked perfectly ~2-3 versions ago (haven't had a chance to bisect yet).

This is how the URL is generated now:

/path/to/file/#Getting-help

# Expected behavior

This is what the URL is supposed to be like:

/path/to/file/Lustre_Testbed_Guide.html#Getting-help

Basically the new Boostnote is stripping off the filename and adding a forward slash...

# Steps to reproduce
* Create a document with at least two headers
* Generate a TOC
* Export as HTML and try to click one

# Environment

- Version : 0.11.17
- OS Version and name : macOS Mojave 10.14.5

Zorlin:
Sorry, let me put it another way - it's somehow creating links as "/#blah" instead of "#blah"

Zorlin:
Weirdly I don't seem to see the issue in the HTML! Very bizarre...

Issue occurs in both Firefox and Chrome.

Flexo013:
This problem is fixed in the 0.12.1 release.

sirrah23:
# This is still an issue

Hi @Flexo013 . I can still reproduce this issue with the new v0.12.1 release.

## Steps to reproduce

Here's what I did in order to reproduce this issue independently (note that there's overlap here with what @Zorlin has already done):

1. Download and install Boostnote v0.12.1
2. Open Boostnote and create a new markdown note containing a TOC and eight headers
3. Export the note to HTML
4. Open the HTML file in Firefox v68.0.1 and click on a link in the TOC

## Expected result
The page will to jump to the anchor for the link that I clicked.

## Actual result
Firefox tried to open the link that I clicked as if it were its own HTML file.

## Recordings of the actual and expected results

Here's a recording of this issue in action:

![Boostnote-Issue-3060_Repro](https://user-images.githubusercontent.com/3108640/62000706-e20e6080-b0ab-11e9-9b9e-9099c8bfb455.gif)

and here's a recording of the expected behavior which I got working by manually fiddling with the exported HTML:

![Boostnote-Issue-3060_HTML-Only-Fix](https://user-images.githubusercontent.com/3108640/62000710-ea669b80-b0ab-11e9-8346-45e7965b3e40.gif)

# Root cause analysis

[...]

PR #3181, opened 2019-07-29
Fix 3060

sirrah23:
## Description

Right now there are only two export types that are using a special
output formatter, pdf and html. Both of these formatters currently populate the
`/html/head/base` portion of the associated html document with the name
of the target directory for the file that the user is exporting.

In order for internal links within the exported document to work
correctly, the value of base must also include the filename. This fix
removes the call to `path.dirname`, which gets rid of the necessary
filename.

## Issue fixed

#3060 

## Type of changes

- :radio_button: : Bug fix (Change that fixed an issue)

## Checklist:

- :radio_button: : My code follows [the project code style](docs/code_style.md)


Flexo013:
Could be related to the latest issue with TOC's

PR #3271, opened 2019-10-11
Fixed #3251 - fixed to use the saved zoom settings value when Boostnote started

### LINKED ISSUE #3251: the zoom setting is not effective when restart Boostnote
# Current behavior
the zoom setting is "80%",but it display as "100%" when restart Boostnote.

# Expected behavior

# Steps to reproduce

1. Open any note and change the zoom setting to any ratio except “100%”.
2. Restart Boostnote.
3. Display right.

# Environment
Download from "https://open.holytech.tech/clientarea.php?action=productdetails&id=28479"
Version: 0.12.1
Electron: 3.0.8
Chrome: 66.0.3359.181
Node.js: 10.2.0
V8: 6.6.346.32
OS: Windows_NT x64 10.0.17763

roottool:
## Description

I fixed to use the saved zoom settings value when Boostnote started.

### Screenshot

useSavedZoomSettingsValue

## Issue fixed

#3251

## Type of changes

## Checklist:

### DIFF
--- a/lib/main-window.js
+++ b/lib/main-window.js
@@ -39,6 +39,7 @@ const windowSize = config.get('windowsize') || {
   width: 1080,
   height: 720
 }
+const zoomFactor = config.get('zoomFactor') || 1.0
 
 const mainWindow = new BrowserWindow({
   x: windowSize.x,
@@ -49,7 +50,7 @@ const mainWindow = new BrowserWindow({
   minWidth: 500,
   minHeight: 320,
   webPreferences: {
-    zoomFactor: 1.0,
+    zoomFactor,
     enableBlinkFeatures: 'OverlayScrollbars'
   },
   icon: path.resolve(__dirname, '../resources/app.png')


lib/main-window.js
```diff
@@ -39,6 +39,7 @@ const windowSize = config.get('windowsize') || {
   width: 1080,
   height: 720
 }
+const zoomFactor = config.get('zoomFactor') || 1.0
```
ZeroX-DG: I don't remember the value `zoomFactor` being saved to config, can you explain where this `zoomFactor` come from? I checked `ZoomManager` but it saving the value directly to the user config file with the key `zoom` not `zoomFactor`

roottool: This `zoomFactor` come from [the electron docs about zoom settings](https://electronjs.org/docs/api/web-frame).
This docs uses `zoomFactor` or `zoomLevel` which are the zoom settings value.
`ZoomManager` uses `zoomFactor`, so I used `zoomfactor`.

But I checked `ZoomManager` again.
The zoom settings value saves to `zoom` in `ConfigManager`.
Should I change from `config.get('zoomFactor')` to `config.get('zoom')`?

ZeroX-DG: @roottool I'm really sorry for the late response. The thing is the zoom factor is saved using the ConfigManager which save to the local storage instead of the config file using electron config so your code here won't be able to retrieve the zoom factor. You will have to communicate with the rendering process to get the factor from the localStorage.
https://github.com/BoostIO/Boostnote/blob/1dd71fc923a5b972bf6685ca4b7aab41faf0f3d5/browser/main/lib/ConfigManager.js#L126-L128

ZeroX-DG:
@roottool ping

ZeroX-DG:
@roottool ping 2

ZeroX-DG:
Close due to inactivity

PR #3398, opened 2019-12-12
Fix #3397

### LINKED ISSUE #3397: Cannot open attachment on storage with path containing non-English letter
# Current behavior

When a note with attachment saved on a storage with path containing non-English letter (e.g. `test_chinese` with Path `D:\測試` as follows), clicking on the link of this note does not open the attachment (like what happened in https://github.com/BoostIO/Boostnote/issues/2003).

**EDIT:** [Here](https://github.com/BoostIO/Boostnote/issues/3397#issuecomment-567133922) is an animation illustrating it.

image

Currently the best solution is *right click > Show in explorer*, and open that manually in explorer.

# Expected behavior

Attachment should be able to open on storage with path containing non-English letter.

# Steps to reproduce

1. Create a storage with path containing non-English letter 
(I only confirmed Chinese letter without trying other language)
2. Create a folder in that storage and add a note
3. Add an attachment into the note
4. Click on the link, the attachment fails to open, *without error message*.

# Environment

- Version : 0.13.0
- OS Version and name : Windows 10

# Diagnosis

When doing *right click -> copy url*, it will copy something like `file:///D:/%E6%B8%AC%E8%A9%A6/attachments/f1acb39f-c175-474c-9a68-fb5523d0d3dd/61b30404.pdf`, this cannot be opened by windows explorer. The url should be `file:///D:/測試/attachments/f1acb39f-c175-474c-9a68-fb5523d0d3dd/61b30404.pdf` instead.

This should be not difficult to solve. I can try to make a PR to fix this.

hiiwave:
## Description

This is trying to fix #3397.

(Update: Done)
~~This is working in progress. Since I have little experience building electron app, I haven't tried if this solution really works.~~ 

~~I can try to build it and confirm later.~~

## Issue fixed

#3397

## Type of changes

- :black_circle: Bug fix (Change that fixed an issue)

## Checklist:

- :black_circle: My code follows [the project code style](docs/code_style.md)
- :white_circle: I have written test for my code and it has been tested -> I think it's untestable
- :black_circle: All existing tests have been passed
- :black_circle: I have attached a screenshot/video to visualize my change if possible -> Please see https://github.com/BoostIO/Boostnote/issues/3397#issuecomment-567133922

### DIFF
--- a/browser/components/MarkdownPreview.js
+++ b/browser/components/MarkdownPreview.js
@@ -1109,7 +1109,7 @@ export default class MarkdownPreview extends React.Component {
     }
 
     // other case
-    const success = shell.openExternal(href) || shell.openExternal(unescape(href))
+    const success = shell.openExternal(href) || shell.openExternal(decodeURI(href))
     if (!success) console.error('failed to open url ' + href)
   }
 
--- a/browser/components/MarkdownPreview.js
+++ b/browser/components/MarkdownPreview.js
@@ -1109,8 +1109,17 @@ export default class MarkdownPreview extends React.Component {
     }
 
     // other case
-    const success = shell.openExternal(href) || shell.openExternal(decodeURI(href))
-    if (!success) console.error('failed to open url ' + href)
+    this.openExternal(href)
+  }
+
+  openExternal (href) {
+    try {
+      const success = shell.openExternal(href) || shell.openExternal(decodeURI(href))
+      if (!success) console.error('failed to open url ' + href)
+    } catch (e) {
+      // URI Error threw from decodeURI
+      console.error(e)
+    }
   }
 
   render () {
--- a/browser/components/MarkdownPreview.js
+++ b/browser/components/MarkdownPreview.js
@@ -1109,7 +1109,8 @@ export default class MarkdownPreview extends React.Component {
     }
 
     // other case
-    shell.openExternal(href)
+    const success = shell.openExternal(href) || shell.openExternal(unescape(href))
+    if (!success) console.error('failed to open url ' + href)
   }
 
   render () {


hiiwave:
(Asking for helps)

Hi,

I tried to run the tests for this project on Windows but failed. As I am not familiar to npm, can anyone advises what happened and how I could fix this here?
The steps to reproduce the problem are as follows:
```
git clone https://github.com/BoostIO/Boostnote.git
cd Boostnote
npm i
npm run test
```

It then shows the following log / error:
```
> boost@0.13.0 test C:\Users\hiiwa_000\resources\Boostnote
> npm run ava && npm run jest


> boost@0.13.0 ava C:\Users\hiiwa_000\resources\Boostnote
> cross-env NODE_ENV=test ava --serial

C:\Users\hiiwa_000\resources\Boostnote\tests\helpers\setup-browser-env.js:29
window.localStorage = {
                    ^

TypeError: Cannot set property localStorage of #<Window> which has only a getter
    at Object.<anonymous> (C:/Users/hiiwa_000/resources/Boostnote/tests/helpers/setup-browser-env.js:22:1)
    at Module._compile (module.js:652:30)
    at loader (C:\Users\hiiwa_000\resources\Boostnote\node_modules\babel-register\lib\node.js:144:5)
    at Object.require.extensions.(anonymous function) [as .js] (C:\Users\hiiwa_000\resources\Boostnote\node_modules\babel-register\lib\node.js:154:7)
    at Module.load (module.js:565:32)
    at tryModuleLoad (module.js:505:12)
    at Function.Module._load (module.js:497:3)
    at Module.require (module.js:596:17)
    at require (internal/module.js:11:18)
    at forEach.x (C:\Users\hiiwa_000\resources\Boostnote\node_modules\ava\lib\test-worker.js:34:3)

...

  32 exceptions

  × tests\dataApi\copyFile-test.js exited with a non-zero exit code: 1

...

```

All the tests failed with the exactly same reason. 

I also tried `npm run compile` before `npm run test`, but it still does not work. If I run `npm run start` or `npm run dev`, I can get electron open, but the app freezes with the spinner spinning forever.

I guessed it's a really basic question but I have no idea how to solve this. Appreciate for any help.

Flexo013:
I think you might need to use `yarn`, but I'm not sure about that. @Rokt33r do you know what can help?

ZeroX-DG:
You can always check out the job log in the travis build to find out what command was executed for running test:
https://travis-ci.org/BoostIO/Boostnote/builds/624286760

hiiwave:
Thanks for helping. The key here is that this project is using node v8. I have finished this PR and checked that it fixed #3397 under Windows. Awaiting review.

browser/components/MarkdownPreview.js
```diff
@@ -1109,7 +1109,17 @@ export default class MarkdownPreview extends React.Component {
     }
 
     // other case
-    shell.openExternal(href)
+    this.openExternal(href)
+  }
+
+  openExternal (href) {
+    try {
+      const success = shell.openExternal(href) || shell.openExternal(decodeURI(href))
```
arcturus140: the created Link should always be encoded and onclick decoded. Otherwise it could cause problems with edge cases when a non-encoded file looks like encoded. For example, if you have two files in the same directory:

- %E6%B8%AC%E8%A9%A6.txt
- %25E6%25B8%25AC%25E8%25A9%25A6

hiiwave: As it seems the original behavior was fine under mac os, and the issue seems to be Windows-only, I made it conservative that it prioritized original behavior. If you can confirm that "always decoding" works fine under each os, I'm happy to change to that case.

arcturus140: @hiiwave would you consider testing it under Windows with always decoding? I am currently testing it on Linux.

hiiwave: Always decoding behaves exactly as this implementation, because it simply does not work without decoding.

arcturus140: I can confirm that "always decoding" works fine under Linux.

arcturus140: @hiiwave ok, I was assuming edge cases on Windows that make `shell.openExternal(href)` succeed. 

If you create a storage location in the folder `%20` both raw and decoded will fail because it must be encoded:

`shell.openExternal(encodeURI(href)`

works in this case. I don't know where `href` is stored but it must be encoded on creation.

hiiwave: As https://github.com/BoostIO/Boostnote/issues/3397#issuecomment-569370829:
> IMHO, users should not create a folder named %20 in any case. I don't think anyone will ever do it as well.

ZeroX-DG:
@Rokt33r can you confirm this work on Windows as well?

Rokt33r:
Will do.

Rokt33r:
Confirmed it works. @hiiwave Thanks for the contribution!

Issue #3127, opened 2019-07-18
[bug] admonitions incorrectly displayed in HTML export


arcturus140:
# Current behavior

Admonitions are not displayed correctly in the exported HTML output. 

![image](https://user-images.githubusercontent.com/44707562/61465379-dd410280-a96f-11e9-9479-d57cc95afb5f.png)

Related: #2083

# Expected behavior

Admonition support for HTML output. 

# Steps to reproduce

1. create markdown note
2. File -> Export As / HTML
3. choose export location
4. open generated index.html in web browser 

## test case

<details>
<summary>MWE used to illustrates the issue: </summary>

    !!! note

    sample text

    !!!

</details>

# Environment

- Version : 0.11.12
- OS Version and name : 16.04 Ubuntu
- Browser name and version: Mozilla Firefox 68.0

Flexo013:
I can't reproduce this in 0.11.15 on Windows 10 using Chrome 75.

![image](https://user-images.githubusercontent.com/4507015/61656104-43a38900-acc0-11e9-912a-7aac97f64645.png)

arcturus140:
In the HTML source I can clearly see `font-family: 'Material Icons';` is being used and directs to the right sources. The font is correctly being applied to the pseudo element but the font has not been found. 

I can easily fix the issue by adding the googlefont to the HTML source file. 

maybe you only need to update the font: https://stackoverflow.com/a/34881022/10636865

arcturus140:
as a side note, we should discuss  wether the Boosnote font or Google font should be used for HTML. The former cannot be used without Boostnote and the latter cannot be used without internet. Or simply add the font to the output.

Issue #3396, opened 2019-12-12
Export Fonts when exporting HTML 


shmolf:
# Current behavior
When exporting a note to HTML, the CSS is exported, but not any font files.

As a result, Katex fonts are excluded, and disrupt the presentation of Latex.

Below is an example of Latex, missing the Katex fonts:
![image](https://user-images.githubusercontent.com/6433103/70685472-87029200-1c77-11ea-9273-ee5dcd54acdb.png)

# Expected behavior

Below is an example of Latex, with the proper font representation:
![image](https://user-images.githubusercontent.com/6433103/70685505-a00b4300-1c77-11ea-8f40-6267cb34ec44.png)

# Steps to reproduce

1.  Create a note with the following as part of the content:
```
$$[x^n + y^n = z^n]$$
$$\frac{d}{dx}\left( \int_{0}^{x} f(u)\,du\right)=f(x).$$
```
2. Export the note to an HTML file
3. Load the newly exported HTML file, and view the Latex

# Environment

- Version : 4.2.12
- OS Version and name : *, Windows 10/7

# Tangentially Related Issues
- https://github.com/BoostIO/Boostnote/issues/1772
- https://github.com/BoostIO/Boostnote/issues/1637
- https://github.com/BoostIO/Boostnote/issues/3349
- https://github.com/BoostIO/Boostnote/issues/3202
- [Discussion in a PR](https://github.com/BoostIO/Boostnote/pull/3317#issuecomment-561512876)

arcturus140:
#3127   is also tangentially related as it is missing a font on export.

shmolf:
@Flexo013 I'd like to be assigned this.
I've started working on it, and I'm checking on a few things.
- One of which is checking whether it's safe to [remove a previously added line](https://github.com/BoostIO/Boostnote/commit/bc24acd057d06a691af93c701e0da839cb1fbd20#r36410147)
- Next is how the PDF exports will play out

I'll create a PR after [the prior one](https://github.com/BoostIO/Boostnote/pull/3317) has been merged.

shmolf:
@arcturus140 As an update, I have the Font Export working, but only for fonts originating from the Application. System Fonts are not exported for the HTML.

I cannot get images to export yet, but that was only part of one of the tangent issues.

I got the admonitions working, as that was just another font.

[Here's a test.pdf](https://github.com/BoostIO/Boostnote/files/4006614/test.pdf) for an example.

arcturus140:
[...]

Issue #3202, opened 2019-08-04
Render equations in PDf and html exports


erex:
# Current behavior

Using Latex eqns in note, renders fine within Boostnote
![image](https://user-images.githubusercontent.com/330744/62428633-866f5300-b6fb-11e9-99c7-8e654c3f6625.png)

However PDF looks like this
![image](https://user-images.githubusercontent.com/330744/62428655-b28ad400-b6fb-11e9-8d7f-ea67f7c05fe8.png)

and html exported file looks like this (when viewed in Firefox browser)
![image](https://user-images.githubusercontent.com/330744/62428666-dc43fb00-b6fb-11e9-9ff9-f6425d2188a6.png)

# Expected behavior

Expect pdf and html rendering to match views within application

# Steps to reproduce

1.  Simply used menu option of File | Export as 
2.  html and pdf

# Environment
![image](https://user-images.githubusercontent.com/330744/62428679-04cbf500-b6fc-11e9-945a-dd1980a76033.png)

- Version :
- OS Version and name :  WIndows 10

askming:
same issue still exists in version 0.13.0

shmolf:
@askming Should be resolved in https://github.com/BoostIO/Boostnote/pull/3317

You can view the attachment in [the comment](https://github.com/BoostIO/Boostnote/pull/3317#issuecomment-549628624).

Issue #1637, opened 2018-03-05
[Feature Request] HTML export with custom font


shogochiai:
I could know we have custom font feature, but it may not apply to the planned HTML export feature.

https://github.com/BoostIO/Boostnote/issues/1625#issuecomment-370129455

Let's say we'd better to separate scope between this issue and HTML export issue, yet as the next scope, we'd better to discuss some design of integration.

For now let me do just initiation.

nlopin:
Hi, @shogochiai 

Export to HTML currently exports the same CSS styles Boostnote uses in the preview mode. If you set custom font, it will be in the exported HTML. 
![1](https://user-images.githubusercontent.com/1469636/37571134-b0984418-2b09-11e8-9ddf-92c3e0ecb3ba.png)
![2](https://user-images.githubusercontent.com/1469636/37571135-b0b52812-2b09-11e8-9108-37d3928ef1c5.png)


HTML is portable. If you copy HTML to another computer where there is no such font installed it will use fallback font. Currently, I don't know good way to export font file together with HTML. If you have any idea how to do that, please let me know.

shogochiai:
@kazup01 Do you have any plan/budget to support hosting popular fonts as webfont on BoostIO server?

Rokt33r:
We can exploit `unpkg.com`. I think it is quite reliable because all assets are served by official NPM server and cached permanently.

But, license of fonts should be a huge problem. I think we can serve fonts with open font license only.

shogochiai:
How about to set google fonts as blogging applicable fonts, and one's own custom font are only for local usage?

UI consolidation might be needed though.

nlopin:
Can you describe your issue? What problem are you trying to solve?

shogochiai:
My concern is:

```
1. We can enjoy custom font for local usage
2. We only able to enjoy free-licensed font such as google font on HTML Export mode
3. Hence we'd better to clarify how we tell that fact to users
```

This can be not a real issue, correct me if I wrong.

Rokt33r:
I assume providing some instructions should be enough because this is quite minor issue. I think navigating all fonts of google font in boostnote probably make users confused.

Rokt33r:
I think providing an option to include some scripts to exported html should be enough.

PR #3407, opened 2019-12-28
Export fonts

shmolf:
## Description
Defines a directory within `css` when exporting to HTML. The `fonts` directory includes files used by Katex, MaterialIcons, and Font Awesome.

This is related to the export of HTML and PDF.

[Good PDF Export](https://github.com/BoostIO/Boostnote/files/4006744/GoodExport.pdf) from this update compared to a [Bad PDF Export](https://github.com/BoostIO/Boostnote/files/4006745/BadExport.pdf) from current state.

[Good HTML Export](https://github.com/BoostIO/Boostnote/files/4006748/GoodExport.zip) from this update compared to a [Bad HTML Export](https://github.com/BoostIO/Boostnote/files/4006746/BadExport.zip) from the current state.

## Issue fixed
- [Anchor Issue](https://github.com/BoostIO/Boostnote/issues/3396)
- Handles the export of fonts that originate from within the application, but does not handle OS fonts. So only partially covers https://github.com/BoostIO/Boostnote/issues/1637
- Fixes https://github.com/BoostIO/Boostnote/issues/3127
- The rendering of Formulas were already fixed, but the Fonts and Admonitions mentioned in https://github.com/BoostIO/Boostnote/issues/3202 are fixed with this PR

## Type of changes

- :radio_button: Improvement (Change that improves the code. Maybe performance or development improvement)

## Checklist:


shmolf:
@ZeroX-DG I think it's safe to say this won't make the v14 release. Do you know about when the next minor/major release would be?

ZeroX-DG:
@shmolf sorry for the late response. I'm still investigating why your PR didn't work on my machine. Everything looks fine when they load to HTML but somehow my PDF export still have broken font

ZeroX-DG:
I discovered that the Katex css didn't even loaded therefore the font didn't get to loaded. But when I open the dev tool and format the Katex css file the view reloaded and everything is ok again. Could very be a Chrome bug.

shmolf:
@ZeroX-DG I've looked at the file using Adobe PDF reader to inspect the document properties. It shows the fonts as being embedded in the document.

image

It could be that the PDF export on different OS's is inconsistent. It does rely on the Print functionality.

There was a merge conflict, and so I'm pushing changes for the merge conflict.

ZeroX-DG:
@shmolf I tried to print multiple time and sometime the normal content of the note didn't get to load, sometime the katex is not display and sometime it produce a blank space. When the content of the note did get loaded I check the font property and there's only the font for that text appears, but when only the katex element is displayed, only katex fonts appears. I think this is a chrome bug, if you dont destroy the new window used for printing you will see that sometime the css is not getting loaded therefore it display the whole thing as plain text and no font are getting loaded in the network tab too.

Should we change to something  else, is there any other option for printing PDF?

shmolf:
I'm going to close this PR. We can re-open if it turns out the necessary change can be easily applied to this branch.

But for now, I'm going to start fresh, and look for a different approach. I'd prefer to use chromium's native PDF print capability, rather than a library. But I'll submit a new PR as a replacement for this when I'm finished.

PR #3051, opened 2019-05-31
Fix #2776

amedora:
## Description

Wrap longer line in the code when export to PDF.
Once #3048 has merged, this PR will be merged.

## Issue fixed

- #2776 

## Type of changes

- :radio_button: Bug fix (Change that fixed an issue)
- :white_circle: Breaking change (Change that can cause existing functionality to change)
- :white_circle: Improvement (Change that improves the code. Maybe performance or development improvement)
- :white_circle: Feature (Change that adds new functionality)
- :white_circle: Documentation change (Change that modifies documentation. Maybe typo fixes)

## Checklist:

- :radio_button: My code follows [the project code style](docs/code_style.md)
- :white_circle: I have written test for my code and it has been tested
- :white_circle: All existing tests have been passed
- :radio_button: I have attached a screenshot/video to visualize my change if possible
wrapline

PR #3447, opened 2020-01-24
Fix long code lines that do not wrap in exports

kotlarz:
## Description

Surprising easy fix. Break the line for code snippets so that the snippets do not get cut-off when exporting to HTML and PDF (via. HTML).

##### Before:
Selection_439

##### After:
Selection_440

##### Before:
Selection_441

##### After:
Selection_442

## Issue fixed
#3051 

## Type of changes

## Checklist:

### DIFF
--- a/browser/components/MarkdownPreview.js
+++ b/browser/components/MarkdownPreview.js
@@ -126,6 +126,8 @@ code {
   background-color: rgba(0,0,0,0.04);
   text-align: left;
   direction: ltr;
+  white-space: pre-wrap;
+  word-break: break-word;
 }
 .lineNumber {
   ${lineNumber && 'display: block !important;'}


Flexo013:
This is a great improvement. Could you also make the line number jump to the right line after the previous line wrapped one or more times?

kotlarz:
@Flexo013 Sorry about the late reply.
I'll take a look at it sometime today 👍

Flexo013:
@kotlarz I see you pushed new commits. Could you update the screenshots of the PR description so that they match with the latest version?

ZeroX-DG:
ping @kotlarz can you fix the conflict please?

kotlarz:
@ZeroX-DG done.

@Flexo013 I took a look and it's a bit more work than I initially thought. It comes down to how the preview component grid is setup. MarkdownPreview renders the gutter/content differently than CodeEditor (which does it correctly).

I'm not sure, but I think MarkdownPreview kinda uses CodeMirror incorrectly as the preview and gutters are instead created via. the markdown renderer (`/browser-lib/markdown.js`):
Selection_472

I read through the CodeMirror documentation and there's a `lineWrapping` option for the editor, and it works 100% for the CodeEditor. But the component grid layout is different, as seen in the picture below.
Selection_473

I think the MarkdownPreview grid need some rework, unless there's a simple fix I'm not seeing.

ZeroX-DG:
@kotlarz yes I think there's no easy way to fix this. Can you rework on the code rendering for markdown preview? Do you think it's better to restructure the code preview html structure or use a external lib like https://prismjs.com/ or https://highlightjs.org/

kotlarz:
@ZeroX-DG Yeah I can rework it once I have some free time. Might be better to use an external library instead of creating your own solution (why re-invent the wheel?). I'll take a look if it's worth restructuring to use a lib vs. fix the preview component.

ZeroX-DG:
@kotlarz have you found a way to solve this yet?

ZeroX-DG:
I'll take over this PR on Wednesday. If you interested in solving this yourself, feel free to re-open the PR

PR #3465, opened 2020-02-12
Fix prettier setup

AWolf81:
## Description

During PR #3002 I've noticed that Travis failed because of linting errors. With this PR I'd like to fix that problem and use prettier-standard.

Prettier by itself is not customizable. That's why I've used Standardx in the PR #2992 that was planned to add code formatting.

Not sure what's the best way to proceed. I'm also not sure when Prettier was merged to master but I think the current setup is a bit confusing and the Standardx setup is easier to understand. 

I think both will work - just the other PR needs to be updated. Please let me know how to proceed if we take this one we can close the other PR. 
If the other is preferred  I should update it and check linting.

There were also two prettier config files that I have combined into one. I removed trailing comma rule from .prettierrc as standard.js is with-out trailing comma.

Most changes are adding space before parens to fix the space-before-parens lint issue.

## Type of changes

## Checklist:


AWolf81:
OK, I think we can close this. It was a mistake from me with `space-before-parens` rule. Removed the spaces with `yarn fix` and everything is fine.

I thought the space is required but it is not on master, so the we don't have that space.

Just not sure why format on save is not working for me.

Issue #3364, opened 2019-11-17
Ctrl+click to follow link not working


yetAnotherUsernameTaken:
# Current behavior

Clicking on toc contents does not scroll to linked content in Xubuntu 18.04 .

# Expected behavior

After clicking on toc link, it should scroll to linked or related content.

# Steps to reproduce

1) Make a new note in boostnote.
2) Generate toc 
3) click on toc content

Then you can notice that it does not scroll to linked content.

# Environment

- Version : 0.13
- OS Version and name : xubuntu 18.04

Flexo013:
I can't reproduce this on Ubuntu 16.04.

There were some errors with links not working in version 0.13.0. Those should be fixed in the next release, so hopefully this issue will also be resolved by that.

arcturus140:
this might be caused by a special character. To reproduce we need to know the TOC entry that is failing.

yetAnotherUsernameTaken:
Please check with this markdown note. I am not able to, make the toc links click and make it go to content in xubuntu 18.04.
As you have said, it might issue with recent build, i hope this might be fixed in next version.

[interview questions.txt](https://github.com/BoostIO/Boostnote/files/3858965/interview.questions.txt)
 
Thank you.

Flexo013:
When I create a note with that exact content the TOC links all work fine for me on Ubuntu 16.04. 

Could somebody else with xubuntu 18.04 confirm that this issue exists?

ignaciogarcia:
This happens on MacOS too, it's driving me nuts. Version 0.13.0.

Has been happening for a while now, I must add. I even rolled back to an older version some time ago to recover the functionality, but now I've upgraded to see if the issue had dissapeared and no, it's there.

Flexo013:
@ignaciogarcia Was all TOC links that did not work for you or just scrolling issues?

ignaciogarcia:
I didn't thoroughly click in all TOC links, but the ones I tried didn't work. The "Press Cmd + click to follow link" popup does appear in the editor window, but following the instructions doesn't have any effect. Nor does it if I click on a link in the preview window.

As mentioned, falling back to 0.12.1 fixes the problem IN THE PREVIEW WINDOW. Doesn't work either in the editor window.

I have tried with an old installation (0.11.13) I have at another Mac at home and the behavior is the same; links in the preview window do work, links in the editor window preceded by the popup don't.

Flexo013:
[...]

Issue #1007, opened 2017-10-23
Random words underlined


Parasrah:
## Problem
Words such as `feedbac` are underlined for no apparent reason.

## Example
![image](https://user-images.githubusercontent.com/14935550/31906409-45f9c7e6-b7ee-11e7-833d-8402812c0455.png)

asmsuechan:
Hi, @Right2Drive. Thank you for reporting. I'll investigate it.

Right2Drive:
Np, I'm on version 0.8.15 if that helps you out at all 😄

asmsuechan:
This is the condition that this unexpected behavior happens: if the word is any of `{w| consisted of the combination of more than 7 letters from {a, b, c, d, e, f}}`

These words accept the condition: aaaaaaa, aabbccddee, aefffff
These words reject the condition: a, abcdefg

xpac1985:
If that wasn't obvious - it seems to be some automatic underlining HEX values...

Flexo013:
This issue no longer seems to exist on 0.15.3.

PR #3478, opened 2020-02-19
Fix 3364 editor ctrl click

AWolf81:
## Description

I've added the missing link handling for `:tag`, `:line` & `:note` links in editor (ctrl+click handling). This branch is based on branch `feature-tag-links` so this PR is waiting for PR #3002 to be approved/merged.

I first wanted to move the script from MarkdownPreview link handling to a file that we can use at both location but I think it's OK to have some code duplication at two locations.
It's also not identical code but pretty similar. If there is a location where I can put the `specialLinkHandler` method, I can add it but I think it's also OK like I have added it.

You can test the link handling with the following Markdown (requires some updates so tags & notes are available):
```md
# Tag link test

- Item 1 [#Typescript](:tag:#Typescript)
- Item 2 [#1st tag](:tag:#1st)

# Note link
[XSS plantuml](:note:9a4c36c2-84e9-448a-a01c-eb1e555b7a6b)

# Some dummy text (for line link test)
Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet.

# Line link
[Goto line 3](:line:3)

# External link
[Google](https://www.google.com)
```

## Issue fixed

#3364

## Type of changes

## Checklist:

## Screenshot:
grafik

## Code decisions:
- Created `eventEmitter.emit` method in `hyperlink.js` as importing from `eventEmitter.js` was not working. I think that's OK as only the emit method is needed here.
- New event `dispatch:push` added to `browser/main/main.js` it's used to dispatch a push action to `connected-router` so it's possible to change the route for tag link handling. Not sure about the naming but I think it is OK. Location of the listener is also OK in main.js or is there a better location?
- Reading the link href from the editor is a bit different compared to the MarkdownPreview. It is using the `innerText` of the clicked node, trims whitespaces (in case there are some - probably optional) and remove the braces.

### DIFF
--- a/extra_scripts/codemirror/addon/hyperlink/hyperlink.js
+++ b/extra_scripts/codemirror/addon/hyperlink/hyperlink.js
@@ -1,22 +1,31 @@
-(function (mod) {
-  if (typeof exports === 'object' && typeof module === 'object') { // Common JS
+;(function(mod) {
+  if (typeof exports === 'object' && typeof module === 'object') {
+    // Common JS
     mod(require('../codemirror/lib/codemirror'))
-  } else if (typeof define === 'function' && define.amd) { // AMD
+  } else if (typeof define === 'function' && define.amd) {
+    // AMD
     define(['../codemirror/lib/codemirror'], mod)
-  } else { // Plain browser env
+  } else {
+    // Plain browser env
     mod(CodeMirror)
   }
-})(function (CodeMirror) {
+})(function(CodeMirror) {
   'use strict'
 
   const shell = require('electron').shell
+  const remote = require('electron').remote
+  const eventEmitter = {
+    emit: function() {
+      remote.getCurrentWindow().webContents.send.apply(null, arguments)
+    }
+  }
   const yOffset = 2
 
   const macOS = global.process.platform === 'darwin'
   const modifier = macOS ? 'metaKey' : 'ctrlKey'
 
   class HyperLink {
-    constructor (cm) {
+    constructor(cm) {
       this.cm = cm
       this.lineDiv = cm.display.lineDiv
 
@@ -28,11 +37,16 @@
       this.tooltip = document.createElement('div')
       this.tooltipContent = document.createElement('div')
       this.tooltipIndicator = document.createElement('div')
-      this.tooltip.setAttribute('class', 'CodeMirror-hover CodeMirror-matchingbracket CodeMirror-selected')
+      this.tooltip.setAttribute(
+        'class',
+        'CodeMirror-hover CodeMirror-matchingbracket CodeMirror-selected'
+      )
       this.tooltip.setAttribute('cm-ignore-events', 'true')
       this.tooltip.appendChild(this.tooltipContent)
       this.tooltip.appendChild(this.tooltipIndicator)
-      this.tooltipContent.textContent = `${macOS ? 'Cmd(⌘)' : 'Ctrl(^)'} + click to follow link`
+      this.tooltipContent.textContent = `${
+        macOS ? 'Cmd(⌘)' : 'Ctrl(^)'
+      } + click to follow link`
 
       this.lineDiv.addEventListener('mousedown', this.onMouseDown)
       this.lineDiv.addEventListener('mouseenter', this.onMouseEnter, {
@@ -47,7 +61,7 @@
         passive: true
       })
     }
-    getUrl (el) {
+    getUrl(el) {
       const className = el.className.split(' ')
 
       if (className.indexOf('cm-url') !== -1) {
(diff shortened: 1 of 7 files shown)


extra_scripts/codemirror/addon/hyperlink/hyperlink.js
```diff
@@ -60,13 +74,83 @@
...
+      // this will match the old link format storage.key-note.key
+      // e.g.
+      // 877f99c3268608328037-1c211eb7dcb463de6490
+      const regexIsLegacyNoteLink = /^(.{20})-(.{20})$/
+      if (regexIsLegacyNoteLink.test(linkHash)) {
+        eventEmitter.emit('list:jump', linkHash.split('-')[1])
+        return
+      }
+
+      const regexIsTagLink = /^:tag:#([\w]+)$/
```
ZeroX-DG: Can you remove the `#` since we decided to remove `#` from tag link in the previous PR?

ZeroX-DG:
@AWolf81 ping

AWolf81:
@ZeroX-DG Sorry, for not responding. I've thought your requested changes are easy to fix but there is a new issue once I've removed the `#` from the link and I'm not sure how to fix it.

Codemirror is changing the link handling with-out the `#`. In the following screenshot you can see the difference.
grafik
The first link is as we'd like to have it `:tag:1st` and the second is the old one with the hash `:tag:#1st`. You can see at the editor highlighting that Codemirror does something different - blue text vs red. 
In the html markup there are several spans for the url instead of one.

I need to do more debugging for this. Do you have an idea what could cause this? An easy fix would be to use the `#` in tag links but I'll try to find a solution with the consistent tag-links as discussed before.

extra_scripts/codemirror/addon/hyperlink/hyperlink.js
```diff
@@ -60,13 +74,83 @@
...
 
+      const rawHref = e.target.innerText.trim().slice(1, -1) // get link text from markdown text
+
+      if (!rawHref) return // not checked href because parser will create file://... string for [empty link]()
+
+      const parser = document.createElement('a')
+      parser.href = rawHref
+      const { href, hash } = parser
+
+      if (!rawHref) return // not checked href because parser will create file://... string for [empty link]()
```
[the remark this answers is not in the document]
AWolf81: OK, no problem. I've removed it.

extra_scripts/codemirror/addon/hyperlink/hyperlink.js
AWolf81: I've removed it but there is a new issue introduced by this change - see comment below.

ZeroX-DG:
CodeMirror used to do weird things in the markdown syntax. For example:
- https://github.com/BoostIO/Boostnote/issues/1007

So I think it's CodeMirror Markdown parsing issue. Currently Boostnote used its' own CodeMirror mode, so we can fix this by overriding the rule in mode I think. But I'm not quite sure how to do it yet. I'll see if I can find out a way to fix this but in the mean time u are welcome to help me investigate on it.

AWolf81:
@ZeroX-DG OK, I've found the location of the problem. It's inside `codemirror/mode/gfm.js` in the regex 
```js
var urlRE = /^((?:(?:aaas?|about|acap|adiumxtra|af[ps]|aim|apt|attachment|aw|beshare|bitcoin|bolo|callto|cap|chrome(?:-extension)?|cid|coap|com-eventbrite-attendee|content|crid|cvs|data|dav|dict|dlna-(?:playcontainer|playsingle)|dns|doi|dtn|dvb|ed2k|facetime|feed|file|finger|fish|ftp|geo|gg|git|gizmoproject|go|gopher|gtalk|h323|hcp|https?|iax|icap|icon|im|imap|info|ipn|ipp|irc[6s]?|iris(?:\.beep|\.lwz|\.xpc|\.xpcs)?|itms|jar|javascript|jms|keyparc|lastfm|ldaps?|magnet|mailto|maps|market|message|mid|mms|ms-help|msnim|msrps?|mtqp|mumble|mupdate|mvn|news|nfs|nih?|nntp|notes|oid|opaquelocktoken|palm|paparazzi|platform|pop|pres|proxy|psyc|query|res(?:ource)?|rmi|rsync|rtmp|rtsp|secondlife|service|session|sftp|sgn|shttp|sieve|sips?|skype|sm[bs]|snmp|soap\.beeps?|soldat|spotify|ssh|steam|svn|tag|teamspeak|tel(?:net)?|tftp|things|thismessage|tip|tn3270|tv|udp|unreal|urn|ut2004|vemmi|ventrilo|view-source|webcal|wss?|wtai|wyciwyg|xcon(?:-userid)?|xfire|xmlrpc\.beeps?|xmpp|xri|ymsgr|z39\.50[rs]?):(?:\/{1,3}|[a-z0-9%])|www\d{0,3}[.]|[a-z0-9.\-]+[.][a-z]{2,4}\/)(?:[^\s()<>]|\([^\s()<>]*\))+(?:\([^\s()<>]*\)|[^\s`*!()\[\]{};:'".,<>?«»“”‘’]))/i
```

There is a `tag` matcher that's causing the problem. I have to check where I can override the regex rule in bfm.js or should we change the tag link syntax back to `:tag:#tagname`? Because changing the regex would limit the user from using `tag` inside urls.

The other issue with `feedback` is not happening.

During testing, if found another issue: If your search term is matching `tag` then the link handling is not working as expected.
grafik

ZeroX-DG:
It's tempting to switch back to the `#` syntax but that will conflict with @Rokt33r idea. We probably need to fix this instead of using a workaround.

AWolf81:
Yes, you're right. We should try to fix it.

I've changed the regex URL_re in gfm by removing the `tag` then it was working. I'll check later what link it was planned to match because I think it shouldn't match:
`tag:something`

Do you know how I can overwrite or replace the gfm inside the bfm mode?

ZeroX-DG:
That's a hard problem, I'm still investigating how we can do that. I wonder if we create our custom `gfm` mode file and include it like `bfm` mode, would it work?

AWolf81:
@ZeroX-DG Yes, that was a hard one.
But I think it's OK now. Please have a look.

I've added gfm with a modified regex to Boostnote. I removed the `tag` from `urlRe`. Do you know what link the tag should match? Is it tag://something? I haven't found anything about it in my Google search - so I can't improve the Regex.

I've also fixed the other issue with the search term hitting a link. Just highlighting of the link looks a bit off but the click handler is working as expected. (I also added this to normal link handler - code duplication is OK here as it's only used in two locations)
Should I check the highlighting as well or is it OK as it is?

ZeroX-DG:
The highlighting works flawless on my side. I think this looks cool. Can you show me the highlighting bug on your side @AWolf81 ?

AWolf81:
@ZeroX-DG I wouldn't say it's a bug. I just think this could be improved.
Hovering over the search term `tag` just shows `ctrl+click` tooltip but no link highlighting displayed (see first screenshot below). Or hovering over the other parts : or :1st will just highlight that part of the link instead of the complete link.

Hovered search term screenshot:
grafik

And here with-out a search term and correct light blue highlighting:
grafik

I'm not sure if I'm a bit too critical here but fixing this could be a bit tricky as the yellow highlighting is conflicting the hover effect and that there a multiple spans involved. I think that could be also fixed later as the click handling is working as expected.

ZeroX-DG:
@AWolf81 I see. I think it's gonna be very tricky to fix the highlighting for the search term. Since it's not a major issue, I don't think with should hold this PR and fix it. I think for now this PR looks great, I'll approve it and get it merge first. Then we can find a way to deal with that search term highlighting stuff.

ZeroX-DG:
A possible solution for that issue is by using markText to create our own link element
https://codemirror.net/doc/manual.html#markText

Flexo013:
Nice fix! This works properly for both internal links (to different views) and external links.

However, there does still seem to be an issue with this not working for references within a single note. For example in the TOC: 
image

PR #3577, opened 2020-05-19
Fix/update

### LINKED ISSUE #2827: Improve update procedure
[Issuehunt badges](https://issuehunt.io/r/BoostIo/Boostnote/issues/2827)

We need to fix https://github.com/BoostIO/Boostnote/blob/master/lib/main-app.js#L53.

## New spec

Boostnote detects any updates. It shows dialog box with `Download now` and `Ignore updates` on main window rather than sending notification. In the same time. If user click `Download`, do same.
If user choose `Ignore updates`, don't proceed and set `ignoreUpdates=true` to configuration.

If it is sets, the app won't check updates anymore until the user unset `ignoreUpdates` from the preferences. And add `Update` menu item so user can download the update when they are ready.

To suppress the further update checking, it sets `update-found` flag right before showing the dialog box with `Download now` and `Ignore updates`. So it won't ask again. And it discards the flag after downloading.

After the downloading is finished, do same as the current build. If the user chooses "not now" and click `Update` menu item, the app should show the update dialog(with `Quiti & update` and  `Not now` buttons) again.

## This improvement should resolve the below issues:
- https://github.com/BoostIO/Boostnote/issues/2306
- https://github.com/BoostIO/Boostnote/issues/1792
- https://github.com/BoostIO/Boostnote/issues/2718

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $102.00)

- [boostio boostio](https://issuehunt.io/u/boostio) ($100.00)
- [siditsrivastava siditsrivastava](https://issuehunt.io/u/siditsrivastava) ($2.00)

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/BoostIo/Boostnote/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

gregueiras:
## Description

boostnote

Fixed the About panel in preferences, as it was not storing the correct values in the preferences manager.

When a new update request is triggered, either by the auto-update or from clicking the update option in the menu, a dialog is shown where the user can disable auto-updates or download the new version.
After the download is finished, the user can apply the update and restart or postpone it.

## Issue fixed

Fixes #2827 

## Type of changes

## Checklist:

- :white_circle: This PR will modify the UI or affects the UX
- :white_circle: This PR will add/update/delete a keybinding

### DIFF
--- a/browser/main/modals/PreferencesModal/InfoTab.js
+++ b/browser/main/modals/PreferencesModal/InfoTab.js
@@ -20,21 +20,28 @@ class InfoTab extends React.Component {
     }
   }
 
+  componentDidMount() {
+    const { autoUpdateEnabled, amaEnabled } = ConfigManager.get()
+
+    this.setState({ config: { autoUpdateEnabled, amaEnabled } })
+  }
+
   handleLinkClick(e) {
     shell.openExternal(e.currentTarget.href)
     e.preventDefault()
   }
 
   handleConfigChange(e) {
-    const newConfig = { amaEnabled: this.refs.amaEnabled.checked }
+    const newConfig = {
+      amaEnabled: this.refs.amaEnabled.checked,
+      autoUpdateEnabled: this.refs.autoUpdateEnabled.checked
+    }
 
     this.setState({ config: newConfig })
   }
 
   handleSaveButtonClick(e) {
-    const newConfig = {
-      amaEnabled: this.state.config.amaEnabled
-    }
+    const newConfig = this.state.config
 
     if (!newConfig.amaEnabled) {
       AwsMobileAnalyticsConfig.recordDynamicCustomEvent('DISABLE_AMA')
@@ -59,15 +66,6 @@ class InfoTab extends React.Component {
       type: 'SET_CONFIG',
       config: newConfig
     })
-  }
-
-  toggleAutoUpdate() {
-    const newConfig = {
-      autoUpdateEnabled: !this.state.config.autoUpdateEnabled
-    }
-
-    this.setState({ config: newConfig })
-    ConfigManager.set(newConfig)
   }
 
   infoMessage() {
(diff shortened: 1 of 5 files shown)


ZeroX-DG:
Looks good but....I don't think forcing people click save button to save config for auto update check too is a good idea. 2 reasons:
1. UI: The auto check update is not in the same section with the analytics config so this will confuse user and they will end up not clicking save
2. UX: The analytics is an important setting since user will have to agree to share their usage. But auto update is not really that important so I think user should be able to enable this simply  by check or uncheck the checkbox.

@Rokt33r what do you think?

gregueiras:
I think both of those reasons make sense, I'll update it

browser/main/modals/PreferencesModal/InfoTab.js
```diff
@@ -20,21 +20,29 @@ class InfoTab extends React.Component {
...
   }
 
   handleConfigChange(e) {
-    const newConfig = { amaEnabled: this.refs.amaEnabled.checked }
+    const newConfig = {
+      amaEnabled: this.refs.amaEnabled.checked,
+      autoUpdateEnabled: this.refs.autoUpdateEnabled.checked
+    }
 
     this.setState({ config: newConfig })
```
ZeroX-DG, line 40: This function will change the `amaEnabled` value when you update `autoUpdateEnabled` even when users haven't click save yet. Demonstration:
Peek 2020-05-26 16-14

gregueiras: Thanks for catching that!

I've fixed it

boostnote

gregueiras:
Hey @Rokt33r,

can you please take a look at this?

Thanks!

gregueiras:
Hi @ZeroX-DG hope you are doing well! Can you please ping @Rokt33r so he can review this PR? Thanks! 😄

